PR #1, opened 2023-03-23
Use conan manage dependency & Fix compile error

jiaoew1991:


loloxwg:
May I ask what is the right version of Conan and Python?
This is my config
- Conan version 1.60.1
- Python 3.11
- MacOS 13
- Arm
- Path: ~/dev/milvus-storage/cpp
- Command: conan install .
- Output:
CleanShot 2023-06-14 at 11 32 05@2x

PR #23, opened 2023-07-20
feat(logger): add spdlog log library

loloxwg:


jiaoew1991:
Most of the projects in milvus-io are using glog, so it would be best if milvus-storage also uses it.

loloxwg:
@jiaoew1991 I will replace it by using glog

PR #52, opened 2023-09-08
feat(read): basic reader

loloxwg:


sunby:
Not implemented. I will continue to add details.

PR #68, opened 2023-09-29
[Go] support concurrent writing, fix concurrent bug, add some unit test

loloxwg:


go/storage/lock/lock_manager.go
```diff
@@ -9,10 +13,18 @@ type LockManager interface {
 	Release(version int64, success bool) error
 }
 
-type EmptyLockManager struct{}
+type EmptyLockManager struct {
+	lock sync.Mutex
```
sunby, line 17: Should not add a lock here. If your target is to test concrrent write, maybe it's better to write a memory lock manager.

go/io/fs/minio_fs.go
```diff
@@ -21,6 +22,19 @@ type MinioFs struct {
...
+}
+
+func ExtractFileName(path string) (string, error) {
+	p := strings.Index(path, "/")
+	if p == -1 {
+		return "", errors.ErrInvalidPath
+	}
+	return path[p+1:], nil
+}
+
```
sunby, line 37: Dont need this func anymore

PR #125, opened 2024-08-02
use util and not internal

mslacken:
fix complie failure as arrow::internal::Uri is a forward declaration


shaoting-huang:
I checked the namespace is arrow::internal::Uri. There is no arrow::util namespace

mslacken:
Right, it needs to be `arrow::internal::Uri`, sorry for the typo. Should I update the PR?

shaoting-huang:
Please update to the latest version, we have updated to arrow 17.0.0.

PR #126, opened 2024-08-02
fix linking for cpp and add right boost link

mslacken:
cmake fixes


shaoting-huang:
What is the point? You can follow README to compile

PR #131, opened 2024-08-13
integrate packed reader and writer

shaoting-huang:
issue: #127 

feat:
- separate chunk logic from the packed reader into a chunk manager
- add unit test for the packed reader and chunk manager
- fix writer bug
- add reader and writer integration test


cpp/include/milvus-storage/packed/reader.h
```diff
@@ -95,19 +72,19 @@ class PackedRecordBatchReader : public arrow::RecordBatchReader {
   private:
   // Advance buffer to fill the expected buffer size
   arrow::Status advanceBuffer();
+  std::vector<const arrow::Array*> collectChunks(int64_t chunksize) const;
```
tedxu, line 75: Do we have an implementation for this function?

cpp/src/common/fs_util.cpp
```diff
@@ -20,10 +20,13 @@
 #include <cstdlib>
 #include "common/log.h"
 #include "common/macro.h"
+#include <iostream>
 #ifdef MILVUS_OPENDAL
 #include "common/opendal_fs.h"
 #endif
 
+using namespace std;
```
tedxu, line 28: nit: the namespace is not actually used.

cpp/src/common/fs_util.cpp
```diff
@@ -20,10 +20,13 @@
 #include <cstdlib>
 #include "common/log.h"
 #include "common/macro.h"
+#include <iostream>
```
tedxu, line 23: nit: not needed.

cpp/src/packed/chunk_manager.cpp
```diff
@@ -0,0 +1,92 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <arrow/array/array_base.h>
+#include <arrow/status.h>
+#include <arrow/table.h>
+#include <parquet/properties.h>
+#include "packed/chunk_manager.h"
+
+#include <iostream>
```
tedxu, line 21: nit: not needed.

cpp/src/packed/chunk_manager.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+using namespace std;
```
tedxu, line 23: nit: not needed.

cpp/src/packed/chunk_manager.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+  // Slice chunks and advance chunk index as appropriate
+  std::vector<std::shared_ptr<arrow::ArrayData>> batch_data(column_offsets_.size());
+  for (int i = 0; i < column_offsets_.size(); ++i) {
+    batch_data[i] = SliceChunk(chunks[i], i, chunksize);
+  }
+  return batch_data;
+}
+
+std::vector<const arrow::Array*> ChunkManager::CollectChunks(const std::vector<std::shared_ptr<arrow::Table>>& tables,
```
tedxu, line 46: Returning a vector of `std::shared_ptr<arrow::Array>` is probably better, or else we must handle the lifecycle of the pointers.

cpp/src/packed/column_group.cpp
```diff
@@ -12,20 +12,23 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "writer/column_group.h"
+#include "packed/column_group.h"
 #include "common/arrow_util.h"
 #include "arrow/table.h"
```
tedxu, line 17: nit:

```suggestion
#include <arrow/table.h>
```

cpp/src/packed/column_group.cpp
```diff
@@ -12,20 +12,23 @@
...
 #include "common/status.h"
+#include <iostream>
+
+using namespace std;
```
tedxu, line 21: nit: no need.

cpp/test/packed/stream_writer_test.cpp
```diff
@@ -0,0 +1,139 @@
...
+#include <gtest/gtest.h>
+#include <arrow/api.h>
+#include <parquet/api/writer.h>
+#include <packed/stream_writer.h>
+#include <parquet/properties.h>
+#include <common/status.h>
+#include <packed/reader.h>
+#include <filesystem>
+#include <memory>
+#include <filesystem>
```
tedxu, line 25: Duplicated include.

cpp/test/packed/stream_writer_test.cpp
```diff
@@ -0,0 +1,139 @@
...
+// limitations under the License.
+
+#include <arrow/filesystem/localfs.h>
+#include <gtest/gtest.h>
+#include <arrow/api.h>
+#include <parquet/api/writer.h>
+#include <packed/stream_writer.h>
+#include <parquet/properties.h>
+#include <common/status.h>
+#include <packed/reader.h>
```
tedxu, line 22: nit: 

```suggestion
#include "packed/reader.h"
```

cpp/test/packed/stream_writer_test.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+namespace milvus_storage {
+
+class StreamWriterTest : public ::testing::Test {
+  protected:
+  StreamWriterTest() : props_(*parquet::default_writer_properties()) {}
+
+  void SetUp() override {
+    file_path_ = std::filesystem::temp_directory_path() / "stream_writer_test";
+    std::filesystem::create_directory(file_path_);
```
tedxu, line 46: There is a `BuildFileSystem` call in fs_util.h, which returns a `arrow::fs::FileSystem`. We'd better use that call. I suppose we won't use `std::filesystem` anywhere in this project.

cpp/test/packed/stream_writer_test.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+  std::vector<std::string> strs;
+  strs.reserve(str_res->length());
+  for (int i = 0; i < str_res->length(); ++i) {
+    strs.push_back(str_res->GetString(i));
+  }
+
+  EXPECT_THAT(strs, testing::ElementsAre(std::string(10000, 'a'), std::string(10000, 'b'), std::string(10000, 'c'),
+                                         std::string(10000, 'a'), std::string(10000, 'b'), std::string(10000, 'c')));
+  for (int i = 1; i < 11; ++i) {
```
tedxu, line 126: It may be necessary to add a test that read and write files that larger than the buffer size, to verify the state transitions on advancing new buffers.

cpp/include/milvus-storage/packed/reader.h
shaoting-huang: it is implemented in chunk_manager.cpp

cpp/test/packed/stream_writer_test.cpp
shaoting-huang: updated the test

PR #133, opened 2024-08-20
enhance: fix packed reader and writer

shaoting-huang:
issue: #131 

feat:
- Fix packed writer row group size
- Fix packed reader row group memory size by reading from file metadata
- Add logging


tedxu (commented):
/lgtm

PR #136, opened 2024-08-22
enhance: support multiple reading row groups to avoid missing data

shaoting-huang:
issue: #127

### DIFF
--- a/cpp/src/packed/chunk_manager.cpp
+++ b/cpp/src/packed/chunk_manager.cpp
@@ -16,6 +16,8 @@
 #include <arrow/status.h>
 #include <arrow/table.h>
 #include <parquet/properties.h>
+#include "common/log.h"
+#include "packed/reader.h"
 #include "packed/chunk_manager.h"
 
 namespace milvus_storage {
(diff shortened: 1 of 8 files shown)


cpp/src/packed/chunk_manager.cpp
```diff
@@ -60,12 +57,19 @@ std::vector<std::shared_ptr<arrow::ArrayData>> ChunkManager::SliceChunks(
       chunk_state.addChunk(1);
       chunk_state.resetOffset();
       slice_data = (offset > 0) ? chunk->Slice(offset, chunksize_)->data() : chunk->data();
+      if (chunk_state.chunk == chunk_sizes[i]) {
+        table_to_pop.insert(column_offsets_[i].path_index);
```
tedxu, line 61: Why can't we just pop the tables in place?

cpp/src/packed/reader.cpp
```diff
@@ -108,23 +108,21 @@ arrow::Status PackedRecordBatchReader::advanceBuffer() {
...
+    int64_t size_in_plan = std::stoll(reader->parquet_reader()->metadata()->key_value_metadata()->value(next_rg));
+    if (plan_buffer_size + size_in_plan >= buffer_available_) {
+      break;
+    }
+    int64_t next_row_group_size = advance_row_group(i);
+    sorted_offsets.pop();
+    sorted_offsets.emplace(i, column_group_states_[i].row_offset);
   }
 
+  std::shared_ptr<arrow::Table> concatenated_table;
```
tedxu, line 125: The variable is not used.

cpp/src/packed/reader.cpp
```diff
@@ -108,23 +108,21 @@ arrow::Status PackedRecordBatchReader::advanceBuffer() {
...
-        }
-        sorted_offsets.pop();
-        sorted_offsets.emplace(i, column_group_states_[i].row_offset);
-        continue;
-      }
+    if (next_rg >= reader->parquet_reader()->metadata()->num_row_groups()) {
+      break;
     }
-    break;
+    int64_t size_in_plan = std::stoll(reader->parquet_reader()->metadata()->key_value_metadata()->value(next_rg));
```
tedxu, line 116: nit: the `stoll` conversion can be cached.

cpp/src/packed/reader.cpp
shaoting-huang: removed

cpp/src/packed/chunk_manager.cpp
shaoting-huang: the tables are from row groups that needed to be read. If the chunk achieves the chunk size, there is no more chunk in the table to be read. Therefore we can pop out the table from the queue. In the next run, it will process the new table.

PR #139, opened 2024-08-28
fix packed writer memory limit

shaoting-huang:
issue: #127

### DIFF
--- a/cpp/include/milvus-storage/common/fs_util.h
+++ b/cpp/include/milvus-storage/common/fs_util.h
@@ -23,4 +23,9 @@ Result<std::shared_ptr<arrow::fs::FileSystem>> BuildFileSystem(const std::string
 
 std::string UriToPath(const std::string& uri);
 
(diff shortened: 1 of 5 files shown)


cpp/include/milvus-storage/common/fs_util.h
```diff
@@ -23,4 +23,9 @@ Result<std::shared_ptr<arrow::fs::FileSystem>> BuildFileSystem(const std::string
 
 std::string UriToPath(const std::string& uri);
 
+static constexpr int64_t DEFAULT_MAX_ROW_GROUP_SIZE = 1024 * 1024;  // 1 MB
```
tedxu, line 26: Given `ARROW_PART_UPLOAD_SIZE=10MB`, the row group size may be too small. 

If the row group size is not configurable, we'd better find out the optimal value from some benchmarks.

tedxu (commented):
/lgtm

PR #141, opened 2024-09-01
enhance: Init buffer for packed writer

shaoting-huang:
issue: #127

### DIFF
--- a/cpp/src/packed/writer.cpp
+++ b/cpp/src/packed/writer.cpp
@@ -15,12 +15,14 @@
 #include "packed/writer.h"
 #include <cstddef>
 #include "common/log.h"
+#include "common/macro.h"
 #include "common/status.h"
 #include "packed/column_group.h"
 #include "packed/column_group_writer.h"
 #include "packed/splitter/indices_based_splitter.h"
 #include "packed/splitter/size_based_splitter.h"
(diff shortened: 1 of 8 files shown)


cpp/src/packed/writer.cpp
```diff
@@ -35,40 +37,43 @@ PackedRecordBatchWriter::PackedRecordBatchWriter(size_t memory_limit,
...
+      current_memory_usage_(0),
+      size_split_done_(false) {}
 
-Status PackedRecordBatchWriter::Init(const std::shared_ptr<arrow::RecordBatch>& record) {
-  // split first batch into column groups
-  std::vector<ColumnGroup> groups = SizeBasedSplitter(record->num_columns()).Split(record);
+Status PackedRecordBatchWriter::Write(const std::shared_ptr<arrow::RecordBatch>& record) {
+  if (!size_split_done_) {
+    size_t next_batch_size = GetRecordBatchMemorySize(record);
+    if (current_memory_usage_ + next_batch_size < memory_limit_) {
```
tedxu, line 46: I believe the upper bound of memory in the probing phase should be `memory_limit_ / 2`, or else the total use of memory could exceed `memory_limit_` when the records are actually written to the buffer.

PR #142, opened 2024-09-03
enhance: optimize kv metadata and fix first large row group

shaoting-huang:
issue: #127

### DIFF
--- a/cpp/src/format/parquet/file_writer.cpp
+++ b/cpp/src/format/parquet/file_writer.cpp
@@ -12,13 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "common/log.h"
 #include "common/macro.h"
 #include "format/parquet/file_writer.h"
 #include <parquet/properties.h>
 #include <memory>
 #include <string>
 #include "common/fs_util.h"
+#include <boost/variant.hpp>
 
 namespace milvus_storage {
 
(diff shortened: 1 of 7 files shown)


cpp/src/format/parquet/file_writer.cpp
```diff
@@ -83,13 +83,17 @@ Status ParquetFileWriter::WriteRecordBatches(const std::vector<std::shared_ptr<a
...
     RETURN_ARROW_NOT_OK(WriteRowGroup(current_batches, current_size));
   }
-  RETURN_ARROW_NOT_OK(writer_->AddKeyValueMetadata(kv_metadata_));
   return Status::OK();
 }
 
 int64_t ParquetFileWriter::count() { return count_; }
 
 Status ParquetFileWriter::Close() {
+  std::stringstream ss;
```
tedxu, line 92: Can we just write bytes rather than comma-separated-strings?

cpp/src/packed/splitter/size_based_splitter.cpp
```diff
@@ -27,65 +27,60 @@ SizeBasedSplitter::SizeBasedSplitter(size_t max_group_size) : max_group_size_(ma
...
-    if (!column) {
-      throw std::runtime_error("Column is null");
+  // calculate column sizes and rows
+  std::vector<size_t> column_sizes(batches[0]->num_columns(), 0);
+  std::vector<int64_t> column_rows(batches[0]->num_columns(), 0);
+  for (const auto& record : batches) {
+    for (int i = 0; i < record->num_columns(); ++i) {
+      std::shared_ptr<arrow::Array> column = record->column(i);
+      if (!column) {
+        throw std::runtime_error("Column is null");
```
tedxu, line 42: The error handling is supposed to be either exception or return value of type `Result`, not both. BTW, I don't think it's necessary to check null here.

tedxu (commented):
/lgtm

PR #147, opened 2024-09-10
Fix azure compile

sunby:


cpp/test/common/fs_test.cpp
```diff
@@ -0,0 +1,69 @@
...
+#include "common/log.h"
+#include "gtest/gtest.h"
+#include "arrow/result.h"
+#include "arrow/buffer.h"
+
+TEST(fs, aliyun) {
+  // this test case is just for aliyun access testing, please don't run it
+  GTEST_SKIP();
+
+  const std::string access_key = "";
```
tedxu, line 16: nit: get secrete keys from environment variables.

tedxu (commented):
/lgtm

PR #151, opened 2024-09-29
fix: enable benchmark

tedxu:


tedxu:
Benchmark result:

```
------------------------------------------------------------------------
Benchmark                              Time             CPU   Iterations
------------------------------------------------------------------------
S3Fixture/Write32MB/real_time   43236907 ns     38985846 ns           13
S3Fixture/Read32MB/real_time     9751422 ns      8241057 ns           70
```

Write is much slower than read, not sure if it is expected.

cpp/benchmark/benchmark_packed.cpp
```diff
@@ -75,22 +78,23 @@ static void PackedRead(benchmark::State& st, arrow::fs::FileSystem* fs, const st
...
-    auto r = pr.ToTable();
-    if (!r.ok()) {
-      st.SkipWithError(r.status().ToString());
-    }
-    auto table = r.ValueOrDie();
-    r = pr.Close();
-    if (!r.ok()) {
-      st.SkipWithError(r.status().ToString());
+    auto r = arrow::RecordBatch::MakeEmpty(schema);
+    SKIP_IF_NOT_OK(r.status(), st)
```
shaoting-huang, line 90: Does it affect the benchmark result if we skip?

shaoting-huang (approved):

cpp/benchmark/benchmark_packed.cpp
tedxu: Skip means abort the benchmark if there is error.

PR #168, opened 2025-01-19
feat: Storagev2 cpp only

shaoting-huang:
1. move go part to Milvus
2. support azure compile flag


.gitignore
```diff
@@ -11,3 +11,4 @@ go/.idea/*
 compile_commands.json
 CMakeUserPresets.json
 .vscode/*
+go/internal/core/output/*
```
tedxu, line 14: Not needed.

cpp/CMakeLists.txt
```diff
@@ -1,15 +1,15 @@
 cmake_minimum_required(VERSION 3.20.0)
 
 project(milvus-storage VERSION 0.1.0)
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
-option(WITH_UT "Build the testing tree." ON)
+option(WITH_UT "Build the testing tree." OFF)
```
tedxu, line 7: Can we still have UT if this option is turned off?

cpp/CMakeLists.txt
```diff
@@ -19,57 +19,65 @@ if (WITH_OPENDAL)
 endif()
 
 find_package(Boost REQUIRED)
+include_directories(${Boost_INCLUDE_DIRS})
```
tedxu, line 22: I believe the `find_package` contains the `include_directories` primitive already, so it is not necessary to call it explicitly.

cpp/include/milvus-storage/common/arrow_util.h
```diff
@@ -16,8 +16,8 @@
 #include <memory>
 #include "parquet/arrow/reader.h"
 #include "arrow/filesystem/filesystem.h"
-#include "common/result.h"
-#include "storage/options.h"
+#include "milvus-storage/common/result.h"
+#include "milvus-storage/storage/options.h"
```
tedxu, line 20: It is strange to include the root path in those `.h` code, we could either set `cpp/include/milvus-storage/` as include path, or just remove the `milvus-storage` subdir.

cpp/CMakeLists.txt
shaoting-huang: test bin is moved into `build/Release/test`

cpp/CMakeLists.txt
shaoting-huang: removed

Issue #171, opened 2025-02-10
[feat]: packed reader and writer new api


shaoting-huang:
For the new packed writer API: Column groups and file paths are explicitly defined by users. The packed writer will not split column groups by memory. It follows the user-defined column groups to split arrow records and write into the corresponding file path.
For the new packed reader API: read paths are explicitly defined by users.

PR #172, opened 2025-02-10
feat: packed reader and writer new api

shaoting-huang:
main feat for the patch:
* Column groups and file paths are explicitly defined by users. The packed writer will not split column groups by memory. It follows the user-defined column groups to split arrow records and write into the corresponding file path.
* Change factory arrow filesystem to singleton arrow file system to support Milvus init core.
* Separate arrow fs config from storage config.
related: #171


cpp/src/c/packed_reader_c.cpp
```diff
@@ -24,11 +24,13 @@
 #include <memory>
 
 int NewPackedReader(const char* path,
+                    CPaths paths,
                     struct ArrowSchema* schema,
                     const int64_t buffer_size,
                     CPackedReader* c_packed_reader) {
   try {
     auto truePath = std::string(path);
```
tedxu: Is this `path` only used to initialize fs?

cpp/include/milvus-storage/packed/writer.h
```diff
@@ -32,16 +32,34 @@ using MemoryMaxHeap =
...
+   */
+  PackedRecordBatchWriter(std::shared_ptr<arrow::fs::FileSystem> fs,
+                          std::vector<std::string>& paths,
                           std::shared_ptr<arrow::Schema> schema,
-                          std::shared_ptr<arrow::fs::FileSystem> fs,
-                          const std::string& file_path,
-                          const StorageConfig& storage_config);
+                          StorageConfig& storage_config,
+                          std::vector<std::vector<int>>& column_groups,
+                          size_t buffer_size = DEFAULT_WRITE_BUFFER_SIZE);
```
tedxu, line 50: Why bother passing `fs` instance if we can create it from `storage_config`?

Or, if `fs` contains all needed information from `storage_config`, we can pass `fs` only.

cpp/include/milvus-storage/filesystem/fs.h
```diff
@@ -41,10 +62,59 @@ class FileSystemProducer {
...
+
+  void Init(const ArrowFileSystemConfig& config) {
+    if (afs_ == nullptr) {
+      afs_ = createArrowFileSystem(config).value();
+    }
+  }
+
+  void Release() {}
+
+  ArrowFileSystemPtr GetArrowFileSystem() { return afs_; }
```
tedxu, line 86: Why bother creating a `ArrowFileSystemSingleton` instance, if we have the `afs_` instance already?

Additionally, we should assert the `afs_` exist in such method.

cpp/src/c/packed_reader_c.cpp
shaoting-huang: moved cgo part to milvus

cpp/include/milvus-storage/packed/writer.h
shaoting-huang: Move arrow fs related config out from storage_config to arrow_fs_config

PR #175, opened 2025-02-24
feat: support gcs iam

shaoting-huang:
related: #173


shaoting-huang:
result of testing on gcp instance
Screenshot 2025-02-24 at 22 46 53

Issue #180, opened 2025-03-25
[feat]: support schema evolution for packed reader and file reader


shaoting-huang:
Schema Evolution Support for PackedReader and FileReader

* Partial Field Reading: Allow reading only a subset of fields if the reader schema contains fewer fields than the file schema.
* Field Order Alignment: Ensure that the reading order follows the sequence defined in the reader schema, regardless of the field order in the file schema.
* Null Value Filling: If the reader schema includes additional fields that are not present in the file schema, fill these fields with null values.

PR #181, opened 2025-03-25
feat: schema evolution for packed reader and file reader

shaoting-huang:
Schema Evolution Support for PackedReader and FileReader

1. Partial Field Reading: Allow reading only a subset of fields if the reader schema contains fewer fields than the file schema.
2. Field Order Alignment: Ensure that the reading order follows the sequence defined in the reader schema, regardless of the field order in the file schema.
3. Null Value Filling: If the reader schema includes additional fields that are not present in the file schema, fill these fields with null values.

related: #180

### DIFF
--- a/cpp/src/format/parquet/file_reader.cpp
+++ b/cpp/src/format/parquet/file_reader.cpp
@@ -12,11 +12,16 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <memory>
 #include <numeric>
+#include <string>
 
+#include <arrow/array/util.h>
+#include <arrow/chunked_array.h>
 #include <arrow/record_batch.h>
 #include <arrow/table.h>
 #include <arrow/table_builder.h>
+#include <arrow/type.h>
 #include <arrow/type_fwd.h>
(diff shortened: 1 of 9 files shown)


cpp/src/format/parquet/file_reader.cpp
```diff
@@ -56,20 +63,32 @@ Status FileRecordBatchReader::init(std::shared_ptr<arrow::fs::FileSystem> fs,
...
+  field_id_list_ = status.value();
+  std::vector<std::shared_ptr<arrow::Field>> fields;
+  for (int i = 0; i < field_id_list_.size(); ++i) {
+    FieldID field_id = field_id_list_.Get(i);
+    if (field_id_mapping.find(field_id) != field_id_mapping.end()) {
+      needed_columns_.push_back(field_id_mapping[field_id].col_index);
+      fields.push_back(schema->field(i));
+    } else {
+      // mark nullable if the field can not be found in the file, in case the reader schema is not marked
+      fields.push_back(schema->field(i)->WithNullable(true));
```
tedxu, line 82: If the schema of missing column is not nullable, we need to raise error here.

PR #191, opened 2025-05-08
fix: s3fs client finalizer should be a singleton

tedxu:
This PR fixes error `S3 subsystem is finalized` reported by `PackedWriter` initialization.

s3fs client finalizer is designed to be a singleton that collects all clients. The current implementation missed the `static` keyword, making `GetClientFinalizer` returns an empty instance every time.


congqixia (approved):
/lgtm

Issue #204, opened 2025-06-18
[Bug]: file row group reader reads despite insufficient memory


shaoting-huang:
when file row group reader reading memory limit is less than row group size, it should throw warning to user and read next row group

PR #211, opened 2025-06-19
fix: file row group reader reads despite insufficient memory

shaoting-huang:
related: #204

### DIFF
--- a/cpp/src/format/parquet/file_reader.cpp
+++ b/cpp/src/format/parquet/file_reader.cpp
@@ -204,12 +204,7 @@ arrow::Status FileRowGroupReader::ReadNextRowGroup(std::shared_ptr<arrow::Table>
 
   // If no row groups can fit in memory, still try to read at least one row group
   if (rgs_to_read.empty() && rg <= rg_end_) {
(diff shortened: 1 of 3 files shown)


cpp/src/format/parquet/file_reader.cpp
```diff
@@ -202,9 +202,25 @@ arrow::Status FileRowGroupReader::ReadNextRowGroup(std::shared_ptr<arrow::Table>
     rg++;
   }
 
+  // If no row groups can fit in memory, still try to read at least one row group
+  if (rgs_to_read.empty() && rg <= rg_end_) {
+    auto current_row_group_size = file_metadata_->GetRowGroupMetadataVector().Get(rg_start_).memory_size();
+    std::string warning_msg =
+        "Insufficient memory to read row group. Required: " + std::to_string(current_row_group_size) +
+        ", Available: " + std::to_string(buffer_size_limit_ - buffer_size_) + ". Attempting to read anyway.";
+    LOG_STORAGE_WARNING_ << warning_msg;
```
tedxu, line 211: There will be too many warning logs if every row group is larger than buffer size.

PR #222, opened 2025-07-23
enhance: handle reader and writer corner case

shaoting-huang:
* Fix segmentation faults when reading schemas with non-existent columns
* Add robust schema evolution support for missing columns (returns null columns with correct row counts)
* Use proper indexing between file readers and original paths
* Enhanc test coverage for multiple row groups, nullable fields, and complex schema evolution scenarios


congqixia (approved):
/lgtm

PR #234, opened 2025-08-13
enhance: adding new API set

tedxu:


cpp/include/milvus-storage/writer.h
```diff
@@ -0,0 +1,552 @@
...
+    return *this;
+  }
+
+  /**
+   * @brief Sets the target file size
+   *
+   * @param size Target size of each column group file in bytes
+   * @return Reference to this builder for method chaining
+   */
+  WritePropertiesBuilder& with_target_file_size(int64_t size) {
```
shaoting-huang, line 166: what is the use case of this method

shaoting-huang:
I think we can remove deprecated manifest.cpp and manifest.h files.

cpp/include/milvus-storage/writer.h
```diff
@@ -0,0 +1,552 @@
...
+  int64_t max_row_group_size = 64 * 1024;
+
+  /// Target size of each column group file in bytes
+  int64_t target_file_size = 128 * 1024 * 1024;  // 128MB
+
+  /// Write buffer size for each column group writer
+  int64_t buffer_size = 64 * 1024 * 1024;  // 64MB
+
+  /// Compression algorithm to use for data storage
+  CompressionType compression = CompressionType::SNAPPY;
```
shaoting-huang, line 61: why do we change from zstd to snappy?

cpp/src/writer.cpp
```diff
@@ -0,0 +1,322 @@
...
+  if (!base_path_.empty() && base_path_.back() != '/') {
+    path_stream << "/";
+  }
+  path_stream << "column_group_" << column_group_id << ".parquet";
+  return path_stream.str();
+}
+
+arrow::Status Writer::finalize_manifest() {
+  // Update manifest version
+  manifest_->set_version(1);
```
shaoting-huang, line 316: current version + 1

cpp/src/manifest.cpp
```diff
@@ -0,0 +1,327 @@
...
+  column_to_group_map_.clear();
+
+  for (const auto& cg : column_groups_) {
+    for (const auto& column_name : cg->columns) {
+      column_to_group_map_[column_name] = cg->id;
+    }
+  }
+}
+
+Status Manifest::validate_column_group(const std::shared_ptr<ColumnGroup>& column_group) const {
```
shaoting-huang, line 246: I think we can put validation into column group builder during build(). So that all of ColumnGroups are valid and will not need to be validated on top of the function.

PR #239, opened 2025-08-19
feat: integrate packed parquet format into unified api

shaoting-huang:


cpp/include/milvus-storage/reader.h
```diff
@@ -202,21 +213,46 @@ class ChunkReader {
...
+   *
+   * @param chunk_index Zero-based index of the chunk
+   * @return Result containing the number of rows in the chunk, or error status
+   */
+  [[nodiscard]] virtual arrow::Result<int64_t> get_chunk_row_num(int64_t chunk_index) const = 0;
+
+  protected:
   std::shared_ptr<arrow::fs::FileSystem> fs_;  ///< Filesystem interface for data access
-  std::shared_ptr<ColumnGroup> column_group_;  ///< Column group metadata and configuration
+  std::string file_path_;                      ///< Path to the data file
```
tedxu, line 243: Changing this will lose Column Group schema, thus lose the ability of schema evolution.

cpp/include/milvus-storage/writer.h
```diff
@@ -322,6 +356,27 @@ class SizeBasedColumnGroupPolicy : public ColumnGroupPolicy {
   private:
   int64_t max_avg_column_size_;
   int64_t max_columns_in_group_;
+  ColumnGroupConfig config_;
+  mutable std::vector<int64_t> column_sizes_;  // Cached column sizes from sampling
+};
+
+// ==================== Internal ColumnGroupWriter Implementation ====================
+
+class ColumnGroupWriter {
```
tedxu, line 365: Can we use a `FormatWriter` to replace this?

cpp/src/format/binary/file_reader.cpp
```diff
@@ -0,0 +1,349 @@
...
+    return result.status();
+  }
+  input_stream_ = result.ValueOrDie();
+
+  ARROW_RETURN_NOT_OK(LoadMetadata());
+
+  return arrow::Status::OK();
+}
+
+arrow::Status BinaryFileReader::LoadMetadata() {
```
tedxu, line 44: Is this format compatible with milvus stats log?

PR #242, opened 2025-09-04
Enhance: Cleanup c/cpp code

jiaqizho:
Currently, the code style of miluvs-storage is not entirely consistent. The current changes aim to unify the code style as much as possible and modify some code that does not follow common C++ practices, including:

- Using get() on shared_ptr/unique_ptr to obtain a raw pointer for method calls. Which is unnecessary and dangerous.
- Removed class-private unused variables and local unused variables.
- Implicit type conversions, ex `for (int i = 0; i < size_t)`, which can be dangerous.
- Unnecessary `std::move`, such as using `std::move` in cases where Return Value Optimization (RVO) already applies.
- Missing `std::move` (resulting in extra memory copies), such as when a local variableis assigned and then passed to `push_back` or `emplace_back`.
- Replacing most `push_back` calls with `emplace_back`, as `push_back` always incurs an additional memory copy.
- Renaming some methods that used underscore naming to camelCase.

This commit does not yet cover all code comprehensively—only parts that were encountered during code review have been modified.

### DIFF
--- a/cpp/include/milvus-storage/common/macro.h
+++ b/cpp/include/milvus-storage/common/macro.h
@@ -25,26 +25,16 @@ namespace milvus_storage {
 #undef RETURN_NOT_OK
 #define RETURN_NOT_OK(status) \
   do {                        \
-    auto _s = (status);       \
-    if (!_s.ok()) {           \
-      return _s;              \
+    if (!(status).ok()) {     \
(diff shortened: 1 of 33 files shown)


tedxu (commented):
/lgtm

PR #247, opened 2025-09-09
Introducing reader FFI

tedxu:


cpp/include/milvus-storage/reader_c.h
```diff
@@ -0,0 +1,206 @@
...
+ * @brief Retrieves multiple chunks by their indices with optional parallel processing
+ *
+ * @param reader ChunkReader handle
+ * @param chunk_indices Array of chunk indices to retrieve
+ * @param num_indices Number of indices in the array
+ * @param parallelism Number of threads to use for parallel reading
+ * @param arrays Output array of RecordBatch handles (caller must free)
+ * @param num_arrays Output number of record batches
+ */
+void get_chunks(ChunkReaderHandle reader,
```
jiaqizho, line 119: if we got a exception in the C++ layer, `try...catch` in the extern "C" function. How does the caller know what specific error occurred?

In C, used to return a `int`, 0 eq success, neg value is the error code. we can provider a C function `char *get_errmsg(int errcode);` to tell caller the details of error.

cpp/src/reader_c.cpp
```diff
@@ -0,0 +1,379 @@
...
+
+    *chunk_indices = static_cast<int64_t*>(malloc(sizeof(int64_t) * output_indices.size()));
+    if (*chunk_indices) {
+      std::copy(output_indices.begin(), output_indices.end(), *chunk_indices);
+      *num_chunk_indices = output_indices.size();
+    } else {
+      *num_chunk_indices = 0;
+    }
+  } catch (...) {
+    *chunk_indices = nullptr;
```
jiaqizho, line 187: the exception(if throw) will be dropped. we need log the exception or pass the details message of error to the caller ...

cpp/src/reader_c.cpp
```diff
@@ -0,0 +1,379 @@
...
+ReaderHandle reader_new(FileSystemHandle fs,
+                        char* manifest,
+                        ArrowSchema* schema,
+                        const char* const* needed_columns,
+                        size_t num_columns,
+                        const ::ReadProperties* properties) {
+  // TODO: Implement reader creation
+  // This function has a void return type but needs to somehow return the created reader
+  // The API design seems incomplete - missing a ReaderHandle* parameter
+  if (!fs || !manifest || !schema || !properties) {
```
jiaqizho, line 274: `assert` replace the `return nullptr`? i don't think caller will pass a nullptr in those field

cpp/src/format/parquet/file_writer.cpp
```diff
@@ -162,14 +164,17 @@ arrow::Status ParquetFileWriter::Flush() {
...
 }
 
-arrow::Status ParquetFileWriter::WriteRowGroup(const std::vector<std::shared_ptr<arrow::RecordBatch>>& batch,
-                                               size_t row_group_size) {
-  ARROW_ASSIGN_OR_RAISE(auto table, arrow::Table::FromRecordBatches(batch));
-  ARROW_RETURN_NOT_OK(writer_->WriteTable(*table));
-
+arrow::Status ParquetFileWriter::write_row_group(const std::vector<std::shared_ptr<arrow::RecordBatch>>& batch,
+                                                 size_t row_group_size) {
+  ARROW_RETURN_NOT_OK(writer_->NewBufferedRowGroup());
```
jiaqizho, line 169: pls add a TODO here, we need add a gtest to make sure this case is expect.

PR #253, opened 2025-09-17
enhance: add generic properties support

tedxu:


cpp/include/milvus-storage/properties.h
```diff
@@ -0,0 +1,65 @@
...
+
+#include <unordered_map>
+#include <string>
+
+namespace milvus_storage::api {
+using Properties = std::unordered_map<std::string, std::string>;
+
+// Template class to represent config keys with their expected value type
+template <typename T>
+class Key {
```
jiaqizho, line 25: defined but unused.

cpp/include/milvus-storage/properties.h
```diff
@@ -0,0 +1,65 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include <unordered_map>
+#include <string>
+
+namespace milvus_storage::api {
+using Properties = std::unordered_map<std::string, std::string>;
```
jiaqizho, line 21: should be `std::unordered_map<std::string, T>;` in the c++ side.

The public api can be `string key, string value`. or `const char *key, uint32_t keysz,const char *value, uint32_t valsz`

The `value` can cast to the `T` type, if it can't then return a `error type of arrow::status`.

jiaqizho (approved):

PR #256, opened 2025-09-22
enhance: support vortex reader

jiaqizho:
Based on the vortex writer commit, this commit implements the
vortex reader. Similar to the vortex writer, the core logic is
still implemented in Rust, with the majority of the functionality
originating from vortex-cxx itself.

However, since vortex-cxx cannot be "directly referenced" (unless
the entire Vortex project is cloned), this commit has adapted a
significant portion of the code while retaining its original
Apache 2.0 license headers.

Unlike the implementation of the vortex writer, the reader
implementation is achieved through rust::cxxbridge, which allows
Rust types to be directly mapped (logically mapped) into C++
memory and called. Furthermore, the currently implemented vortex
reader is capable of reading from object storage (though this
functionality is also not yet fully refined).


jiaqizho:
base on: !254

PR #257, opened 2025-09-22
enhance: Introduce error code in storage reader FFI

jiaqizho:
This commit changes the return values of all FFI functions
(which may failed) to int. The caller can use the `errno_c.h`
as header to obtain specific error types. Before returning
an error in FFI, detailed error information will be printed
to stderr.

### DIFF
--- a/cpp/include/milvus-storage/reader_c.h
+++ b/cpp/include/milvus-storage/reader_c.h
@@ -12,21 +12,25 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#pragma once
-
 #ifdef __cplusplus
 extern "C" {
 #endif
 
+#ifndef READER_C
+#define READER_C
+
 #include <stdbool.h>
 #include <stddef.h>
 #include <stdint.h>
-#include "arrow/c/abi.h"
+#include "milvus-storage/result_c.h"
+
+// TODO: move out this header
+#include <arrow/c/abi.h>
 
 // ==================== Forward Declarations ====================
 
 /// Opaque handle for filesystem interface
-typedef void* FileSystemHandle;
+typedef uintptr_t FileSystemHandle;
 
 // ==================== ReadProperties C Interface ====================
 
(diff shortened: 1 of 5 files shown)


cpp/include/milvus-storage/errno_c.h
```diff
@@ -0,0 +1,69 @@
...
+typedef enum {
+  FFI_SUCCESS = 0,
+  FFI_INVALID_ARGS,
+  FFI_MEMORY_ERROR,
+  FFI_ARROW_ERROR,
+  FFI_LOGICAL_ERROR,
+  FFI_GOT_EXCEPTION,
+  FFI_UNREACHABLE_ERROR,
+  FFI_RC_MAX,
+} ffi_error_code_t;
```
tedxu, line 33: The "ffi_" prefixes are redundant. They can be removed or replaced with "LOON_" which is our project code. 

And, is it easier if we just use `int` instead of an enum?

cpp/include/milvus-storage/errno_c.h
```diff
@@ -0,0 +1,69 @@
...
+// limitations under the License.
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#ifndef ERRRNO_C
+#define ERRRNO_C
+
+#include <time.h>
```
tedxu, line 22: nit: unused include.

cpp/src/errno_c.cpp
```diff
@@ -0,0 +1,103 @@
...
+  ffi_error_code_t code;
+  const char* message;
+  const char* file;
+  int line;
+  time_t timestamp;
+};
+
+struct ffi_error_context {
+  ffi_error_info_t base;
+  char context[ERROR_MSG_MAX_LEN];
```
tedxu, line 39: It is a bit over-design. If we want to record the actual format, it is OK to add a prefix (e.g., `[parquet]`) in `ffi_error_info.message`

PR #261, opened 2025-09-24
enhance: add FFI unit tests

jiaqizho:


cpp/Makefile
```diff
@@ -30,7 +30,7 @@ clean:
 	rm -rf build
 
 test: build
-	cd build/Release/test && ./milvus_test
+	cd build/Release/test && ./milvus_test && ./Test_FFI
```
shaoting-huang, line 33: should we put all tests into milvus_test instead of two bins?

cpp/test/ffi/arrow_c_wrapper.c
```diff
@@ -0,0 +1,380 @@
...
+  assert(schema->dictionary == NULL);
+  assert(schema->private_data == NULL);
+  schema->release = NULL;
+}
+
+void struct_schema_release(struct ArrowSchema* schema) {
+  if (schema->children != NULL) {
+    // release each child schema
+    for (int64_t i = 0; i < schema->n_children; i++) {
+      if (schema->children[i] != NULL) {
```
shaoting-huang, line 105: Does the level of nested chirldren only 1?

jiaqizho: three childs in the test

PR #265, opened 2025-09-29
fix: invalid check-format and fix-format in Makefile

jiaqizho:
The target `check-format` and `fix-format` in Makefile is invalid in the UNIX os system, current commit add the `()` to make sure different group in `find` command.


shaoting-huang:
verified with linux env.
/lgtm
/approve

PR #270, opened 2025-10-10
enhance: Storage top-level API access to the vortex file format

jiaqizho:
In the current commit, vortex has been integrated into the top-level writer/reader API of the storage layer. Additionally, support for reading and writing files using vortex has been added in gtest, where the desired format can be specified via `PROPERTY_FORMAT`. Within the storage, the format can be directly selected using `LOON_FORMAT_PARQUET` or `LOON_FORMAT_VORTEX`.

Meanwhile, this commit modifies part of the implementation in `vx-bridge` by removing portions of the C bridge and fully transitioning to the `cxx::bridge` approach.

Furthermore, unlike Parquet, the vortex format stores `arrow::utf8` as STRING_VIEW after writing, which is an implementation specific to vortex itself.


tedxu:
/lgtm
/approve

Per offline discussion, will support streaming write in the coming PRs.

PR #282, opened 2025-10-21
fix: get_chunks is invalid in parquet format

jiaqizho:
The invalid logical is
```
for (int64_t i = 0; i < chunk_indices.size(); ++i) {
   auto row_group_index = row_group_indices_[i];
   ...
}

auto status = read_file(file_index, chunk_index_from, chunk_indices.size());
```

I really don't understand what's logical doing here. If we want get the right `row group`, we can use the `row_group_indices_[chunk_indices[i]]` which is still invalid. So i decided direct rewrite this function. It's a simple logical here:

1. Reorder the chunk indices(row group indices) into different files
2. Call `ReadRowGroups` get the `arrow::Table`
3. Use the `arrow::TableBatchReader` get the record batchs

Also current commit remove the `ConvertTableToRecordBatch` which require the memory copy. Direct use the `arrow::TableBatchReader` can easy to get the minimal recordbatch(contiguous memory) in the `arrow::Table`.

### DIFF
--- a/cpp/src/format/parquet/chunk_reader.cpp
+++ b/cpp/src/format/parquet/chunk_reader.cpp
@@ -144,56 +144,35 @@ arrow::Result<std::shared_ptr<arrow::RecordBatch>> ParquetChunkReader::get_chunk
 arrow::Result<std::vector<std::shared_ptr<arrow::RecordBatch>>> ParquetChunkReader::get_chunks(
     const std::vector<int64_t>& chunk_indices) {
   assert(!file_readers_.empty());
-
+  std::vector<std::vector<int>> row_groups_in_files(file_readers_.size());
   std::vector<std::shared_ptr<arrow::RecordBatch>> result;
+  std::shared_ptr<arrow::Table> table;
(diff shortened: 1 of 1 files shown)


cpp/src/format/parquet/chunk_reader.cpp
```diff
@@ -144,56 +144,35 @@ arrow::Result<std::shared_ptr<arrow::RecordBatch>> ParquetChunkReader::get_chunk
...
-  for (int64_t i = 0; i < chunk_indices.size(); ++i) {
-    auto row_group_index = row_group_indices_[i];
-
-    if (i == 0) {
-      file_index = row_group_index.file_index;
-      chunk_index_from = 0;
+  for (int64_t chunk_index : chunk_indices) {
+    auto row_group = row_group_indices_[chunk_index];
+    row_groups_in_files[row_group.file_index].emplace_back(row_group.row_group_index_in_file);
+  }
```
shaoting-huang: should we sort the row group index?

jiaqizho: no need, the `chunk_indices` must be sorted before it is passed in.

PR #284, opened 2025-10-23
enhance: Introducing LRUCache to cache the filesystem

jiaqizho:
In the milvus-storage, an LRUCache (template class) has been introduced, and the filesystem is cached using the filesystem config as the key.

Since the time required to create and destroy a filesystem is too long, caching the filesystem is necessary. Since the filesystems used by Parquet and Vortex are different (and there is very hard to create an adapter between two of format), the cache objects used for different entities in `format.cpp` are also different.

### DIFF
--- a/cpp/include/milvus-storage/common/lrucache.h
+++ b/cpp/include/milvus-storage/common/lrucache.h
@@ -1,0 +1,117 @@
+// Copyright 2023 Zilliz
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#pragma once
+
+#include <memory>
+#include <string>
+#include <mutex>
+#include <map>
+#include <shared_mutex>
+#include <functional>
+#include <list>
+
+#include <arrow/status.h>
+#include <arrow/result.h>
+
+namespace milvus_storage {
+
+template <typename K, typename V>
+class LRUCache {
+  public:
+  static LRUCache& getInstance() {
+    static LRUCache instance;
+    return instance;
+  }
+
+  // Set the capacity of cache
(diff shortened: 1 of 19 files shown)


cpp/include/milvus-storage/common/lrucache.h
```diff
@@ -0,0 +1,127 @@
...
+      lru_list_.pop_back();
+    }
+  }
+
+  // Get the value associated with the key
+  [[nodiscard]] arrow::Result<V> get(const K& key, std::function<arrow::Result<V>(const K&)> CreateFunc) {
+    // fast path: check under shared lock
+    {
+      std::unique_lock lock(mutex_);
+      auto it = cache_.find(key);
```
congqixia: unique_lock or shared_lock?

cpp/include/milvus-storage/common/lrucache.h
```diff
@@ -0,0 +1,127 @@
...
+    while (cache_.size() > capacity_) {
+      auto last_key = lru_list_.back();
+      cache_.erase(last_key);
+      lru_list_.pop_back();
+    }
+
+    return instance;
+  }
+
+  // Get the size of cached entrys
```
congqixia: NIT: typo "entrys" => "entries"

cpp/include/milvus-storage/common/lrucache.h
jiaqizho: it's actully the write lock here, because we may update the LRU list if current key hit.

```
lru_list_.splice(lru_list_.begin(), lru_list_, it2->second.second);`
```

cpp/src/reader.cpp
```diff
@@ -161,14 +156,14 @@ PackedRecordBatchReader::PackedRecordBatchReader(std::shared_ptr<arrow::fs::File
...
     auto& column_group = column_groups_[i];
-    auto chunk_reader = internal::api::GroupReaderFactory::create(schema, column_group, fs_, needed_columns_,
-                                                                  properties_, key_retriever_callback_);
+    auto chunk_reader_result = internal::api::GroupReaderFactory::create(schema, column_group, needed_columns_,
+                                                                         properties_, key_retriever_callback_);
 
-    if (!chunk_reader) {
+    if (!chunk_reader_result.ok()) {
       throw std::runtime_error("Failed to create chunk reader for column group " + std::to_string(i));
     }
```
tedxu: nit: change to lazy initialize to better handle exceptions.

tedxu (commented):
/lgtm
/approve

PR #286, opened 2025-10-23
enhance: improve JNI exception handling with detailed error messages

shaoting-huang:
- Replace catch(...) with catch(const std::exception& e) and include exception details (e.what()) acroos JNI functions.

This change provides better error visibility and debugging capabilities when exceptions occur in the JNI layer.

### DIFF
--- a/cpp/src/ffi/reader_writer_jni.cpp
+++ b/cpp/src/ffi/reader_writer_jni.cpp
@@ -24,6 +24,12 @@
 
 // ==================== JNI Utility Functions ====================
 
+/**
+  Should use try-catch block to catch exceptions in C++ code.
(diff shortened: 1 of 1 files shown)


cpp/src/ffi/reader_writer_jni.cpp
```diff
@@ -50,6 +51,7 @@ void ThrowJavaExceptionFromFFIResult(JNIEnv* env, const struct ffi_result* resul
 
   jclass exc_class = env->FindClass(exception_class);
   assert(exc_class != nullptr);
+  env->ThrowNew(exc_class, message);
```
jiaqizho: still need add `return` after `ThrowNew`

PR #289, opened 2025-10-29
enhance: move the implementation of S3Client out of cpp

jiaqizho:
This commit moves the implementation of S3Client out of `multi_part_upload_s3_fs` (because it has become overly bloated). After the storage team decided not to remove the `multi_part_upload_s3_fs` and to maintain a private S3 filesystem(port from arrow), then we should reoriginal arrow implementation should be split out into different files.

However, this does not mean that internal APIs can directly call S3Client. Except for special cases, we should still uniformly use the filesystem's API.


cpp/include/milvus-storage/filesystem/fs.h
```diff
@@ -43,43 +44,41 @@ struct ArrowFileSystemConfig {
...
   std::string region = "";
   bool use_ssl = false;
   std::string ssl_ca_cert = "";
   bool use_iam = false;
   bool use_virtual_host = false;
   int64_t request_timeout_ms = 3000;
-  bool gcp_native_without_auth = false;
-  std::string gcp_credential_json = "";
-  bool use_custom_part_upload = true;
+  [[maybe_unused]] bool gcp_native_without_auth = false;
```
shaoting-huang: the gcp native config is supported in chunk manager in Milvus, but storage only support s3 api for gcs for now

Issue #296, opened 2025-11-05
feat: support huaweicloud credential provider


MrPresent-Han:
support huaweicloud credential provider

jiaqizho:
merged, closed

PR #297, opened 2025-11-05
feat: support huaweicloud iam credentials(#296)

MrPresent-Han:
related: #296


shaoting-huang:
Do we have a end to end validation?

PR #299, opened 2025-11-10
enhance: Clean up the unused code.

jiaqizho:
Removed most of the currently unused APIs in storage, including `space/fragment/filter/reader`. This part of code was used in earlier versions of Milvus and is no longer used by any components.

In the current version of storage, two of interfaces are provided:
- FFI: Provides a C interface, which can be directly called by other languages.
- C++ API: APIs starting with the namespace milvus_storage::api.

In Milvus, the interfaces under `packed/` are directly used, but they will be deprecated after the FFI integration.


jiaqizho:
> should also delete OPENDAL related implementation

@shaoting-huang done.

PR #301, opened 2025-11-10
feat: add manifest JNI bindings and reorganize JNI structure

shaoting-huang:
Add JNI bindings for manifest transaction APIs (begin, commit, abort, getColumnGroups) to enable Java/Scala integration.

Refactor JNI code organization:
  - Move all JNI code from src/ffi/ to dedicated src/jni/ directory
  - Split reader_writer_jni.cpp into reader and writer modules
  - Extract common utilities to jni_utils.cpp
  - Update CMakeLists.txt to use GLOB for automatic JNI file inclusion

This improves maintainability by separating C FFI and JNI interfaces and breaking down monolithic files into manageable modules.

### DIFF
--- a/cpp/include/milvus-storage/ffi_jni.h
+++ b/cpp/include/milvus-storage/ffi_jni.h
@@ -316,7 +316,7 @@ JNIEXPORT void JNICALL Java_io_milvus_storage_ArrowUtils_00024_releaseArrowStrea
  * @param properties_ptr Pointer to properties
  * @return Column groups as JSON string
  */
-JNIEXPORT jstring JNICALL Java_io_milvus_storage_MilvusStorageManifest_getLatestColumnGroups(
+JNIEXPORT jstring JNICALL Java_io_milvus_storage_MilvusStorageManifest_00024_getLatestColumnGroupsNative(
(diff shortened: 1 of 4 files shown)


cpp/src/jni/arrow_utils_jni.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+#include "milvus-storage/ffi_jni.h"
+#include <arrow/c/abi.h>
+#include <string>
+
+// ==================== JNI Arrow Resource Management ====================
+
+extern "C" {
+
+JNIEXPORT void JNICALL Java_io_milvus_storage_ArrowUtils_00024_releaseArrowArray(JNIEnv* env,
```
jiaqizho: what is `00024` mean?

shaoting-huang: 00024 is the JNI encoding for the dollar sign $. You need to add 00024 when the method belongs to a Scala object because the compiler generates a class with a $ suffix. You don't need 00024 for
  regular Scala class methods because they compile to normal class names without the $ suffix.

PR #305, opened 2025-11-12
fix: std::vector out of range in ChunkReaderImpl

jiaqizho:
There are two issues in ChunkReader.

    1. It assumes that the number of `RecordBatches` returned by
    `chunk_reader->get_chunks` equals the number of input chunk ids,
    but in fact, the number of RecordBatches returned by get_chunks
    can be less than or equal to the number of input chunk ids. This
    is because after merging IOs, it might return a single RecordBatch
    (both parquet and vortex).

    2. For a single `chunk_reader`, concurrency cannot be performed, at
    least for parquet(But vortex can). For the same file reader,
    multi-threading will cause many issues.

    The current commit disables the concurrency logic and re-write the
    logical of `chunk_reader->get_chunks`. After get the result of get_chunks
    it will call `Slice()` to make the number correspond to the number
    of input chunk ids.

    And there might be a complete concurrency solution that will default to
    letting the caller manage concurrency resources in the future.

### DIFF
--- a/cpp/test/api_writer_reader_test.cpp
+++ b/cpp/test/api_writer_reader_test.cpp
@@ -18,6 +18,8 @@
 #include <arrow/io/api.h>
 #include <arrow/testing/gtest_util.h>
 #include <unistd.h>
+#include <algorithm>
+#include <random>
 
 #include "milvus-storage/filesystem/fs.h"
 #include "milvus-storage/common/lrucache.h"
@@ -53,7 +55,7 @@ class APIWriterReaderTest : public ::testing::TestWithParam<std::string> {
     ASSERT_OK(fs_->CreateDir(base_path_));
 
     // Create test data
-    CreateTestData();
+    ASSERT_AND_ASSIGN(test_batch_, CreateTestData());
 
     milvus_storage::InitTestProperties(properties_, "/", base_path_);
   }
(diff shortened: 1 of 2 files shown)


cpp/test/api_writer_reader_test.cpp
```diff
@@ -1260,6 +1289,81 @@ TEST_P(APIWriterReaderTest, TestLargeBatch) {
...
+    size_t random_times = 5;
+    for (size_t i = 1; i <= random_times; ++i) {
+      std::vector<int64_t> chunkidx_samples;
+
+      float fraction = static_cast<float>(i) / random_times;
+      size_t samples_size = static_cast<size_t>(chunk_indices.size() * fraction);
+
+      std::sample(chunk_indices.begin(), chunk_indices.end(), std::back_inserter(chunkidx_samples), samples_size, gen);
+
+      std::cout << "sample indices: ";
```
shaoting-huang: remove the std::cout

cpp/test/api_writer_reader_test.cpp
```diff
@@ -63,32 +65,59 @@ class APIWriterReaderTest : public ::testing::TestWithParam<std::string> {
     ASSERT_OK(fs_->DeleteDirContents(GetTempDir()));
   }
 
-  void CreateTestData(uint64_t num_rows = 100, uint64_t vector_dim = 4) {
+  std::string generateRandomString(int max_len) {
```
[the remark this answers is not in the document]
jiaqizho: under TODO, same as use S3 running gtest

cpp/test/api_writer_reader_test.cpp
jiaqizho: the `indices` is random. if current test failed we can't reproduce the case without std::cout here.

PR #308, opened 2025-11-14
enhance: Introduce kv metadata in column groups

jiaqizho:
The current changes enable putting custom key-value pairs in column groups. However, both the key and value must be C strings (i.e., null-terminated char*). The metadata can be added when `writer_close` is called and read out in the function `get_column_group_infos`.

Additionally, the transaction interface now returns with the `read_version`.


congqixia (approved):
/lgtm

PR #310, opened 2025-11-17
enhance: use the arrow::fs::filesystem replace the objectstore in rust

jiaqizho:
In vortex, `apache-rs-object-store` is used as the implementation of object store, and `writer/reader` based on objectstore (`ObjectStoreWriter` and `ObjectStoreIoSource`, both in different paths but named `object_store.rs`) are provided. In storage, access the object storage has already been implemented through objects under `cpp/include/milvus-storage/filesystem`, and the C++ version of object storage already supports multiple `CredentialsProvider` (including aliyun, tencent, huawei, etc.).

This commit removes the logic in storage where vortex `writer/reader` used `apache-rs-object-store`, and instead adds FFI for the filesystem C Interface, enabling the `writer/reader` implementation in rust to directly use the interface `arrow::fs::FileSystem`.

Regarding the call stack:
```
Storage CPP API
  -> Vortex C++ part
    -> Vortex Rust part
      -> Filesystem C Interface
        -> Filesystem C++ implementation.
```

Regarding compilation/linking:

- The Rust part of storage only has declarations for the FFI of the filesystem C Interface. The Rust part compiles into libpvxbridge.a, and directly using this static library will result in undefined symbols.
- The C++ part of storage implements the FFI for the filesystem C Interface through extern "C". Since it is extern "C", after linking libpvxbridge.a, the undefined symbols are resolved. The final compiled libmilvus-storage is complete.

Regarding exception handling:

- The FFI for the filesystem C Interface will convert exceptions into FFIResult (consistent with ffi_c.h).
- Rust generates Error/Result from FFIResult.
- vx-bridge will generate exceptions based on Rust Error/Result.
- vortex writer/reader does not catch any exceptions.
- The top-level FFI in storage will catch all exceptions.

Note that: There are two layers of glue code, i.e., two layers of `try...catch`, which will incur some performance loss.

### DIFF
--- a/cpp/include/milvus-storage/filesystem/ffi/filesystem_internal.h
+++ b/cpp/include/milvus-storage/filesystem/ffi/filesystem_internal.h
@@ -1,0 +1,31 @@
+// Copyright 2024 Zilliz
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include <memory>
+#include <arrow/filesystem/filesystem.h>
+
+template <template <typename> class P, typename T>
+class PolymorphicWrapper {
+  public:
+  PolymorphicWrapper(P<T> obj) : obj_(obj) {}
+  inline P<T> get() const { return obj_; }
+
+  private:
+  P<T> obj_;
+};
+
+using FileSystemWrapper = PolymorphicWrapper<std::shared_ptr, arrow::fs::FileSystem>;
+using OutputStreamWrapper = PolymorphicWrapper<std::shared_ptr, arrow::io::OutputStream>;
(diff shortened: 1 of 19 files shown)


cpp/include/milvus-storage/filesystem/ffi/filesystem_internal.h
```diff
@@ -0,0 +1,31 @@
...
+
+#pragma once
+
+#include <memory>
+#include <arrow/filesystem/filesystem.h>
+
+template <template <typename> class P, typename T>
+class PolymorphicWrapper {
+  public:
+  PolymorphicWrapper(P<T> obj) : obj_(obj) {}
```
shaoting-huang: why use PolymorphicWrapper?

cpp/src/format/vortex/vx-bridge/src/filesystem_c.rs
```diff
@@ -0,0 +1,317 @@
...
+
+struct ThreadSafePtr<T> {
+    ptr: *mut c_void,
+    _marker: PhantomData<T>,
+}
+
+unsafe impl<T> Send for ThreadSafePtr<T> {}
+unsafe impl<T> Sync for ThreadSafePtr<T> {}
+
+impl<T> ThreadSafePtr<T> {
```
shaoting-huang: can we validate the wrapped C++ pointers are actually thread-safe?

shaoting-huang:
can we add some ffi error handling test cases?

cpp/src/format/vortex/vx-bridge/src/filesystem_c.rs
jiaqizho: It must be thread-safe rust struct..

PR #311, opened 2025-11-17
fix: add nullptr check for credentials_provider to prevent crash

shaoting-huang:
Defensive hotfix for intermittent GCP IAM nullptr crashes. Fallback to AnonymousCredentialsProvider when credentials_provider is null.

### DIFF
--- a/cpp/src/filesystem/s3/s3_client.cpp
+++ b/cpp/src/filesystem/s3/s3_client.cpp
@@ -481,7 +481,7 @@ arrow::Result<std::shared_ptr<S3ClientHolder>> ClientBuilder::BuildClient(
                      << "This indicates a race condition or missing initialization. "
                      << "Using AnonymousCredentialsProvider as fallback. "
                      << "Please report this error with stack trace.";
-    credentials_provider_ = std::make_shared<Aws::Auth::AnonymousAWSCredentialsProvider>();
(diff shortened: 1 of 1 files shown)


cpp/src/filesystem/s3/s3_client.cpp
```diff
@@ -473,6 +473,18 @@ const S3Options& ClientBuilder::options() const { return options_; }
...
+
+  // HOTFIX: Prevent nullptr crash in GCP IAM and other race condition scenarios
+  // Root cause: Race condition between S3Options construction and ConfigureAccessKey call
+  // TODO: Investigate and fix the root cause of the race condition
+  if (!credentials_provider_) {
+    ARROW_LOG(ERROR) << "[HOTFIX] credentials_provider is nullptr! "
+                     << "This indicates a race condition or missing initialization. "
+                     << "Using AnonymousCredentialsProvider as fallback. "
+                     << "Please report this error with stack trace.";
+    credentials_provider_ = std::make_shared<Aws::Auth::AnonymousAWSCredentialsProvider>();
```
jiaqizho: can you test in follow case:

1. Temporarily remove the line `s3_fs.cpp:L164`
2. Test GCP with IAM? 

Then the GCP will got invalid credentials or work normal?

PR #313, opened 2025-11-19
enhance: Add start_index/end_index fields to column groups

jiaqizho:
The current changes are made to support external tables in stroage. Since individual files of external tables can be relatively large, it is necessary to perform "logical range" partitioning within the external table files. To support it, the column groups have been augmented with start_index and end_index fields.


shaoting-huang:
can we add tests for start_index/end_index?

cpp/src/format/parquet/file_reader.cpp
```diff
@@ -264,60 +264,4 @@ arrow::Status FileRowGroupReader::Close() {
...
-    skipped += batch->num_rows();
-  } while (skipped < offset);
-
-  auto offset_batch = offset - skipped + batch->num_rows();
-  // zero-copy slice
-  return batch->Slice(offset_batch, 1);
-}
-
-// TODO: support projection
-arrow::Result<std::shared_ptr<arrow::Table>> ParquetFileReader::ReadByOffsets(std::vector<int64_t>& offsets) {
```
shaoting-huang: don't we use the api anymore?

cpp/src/format/parquet/parquet_chunk_reader.cpp
```diff
@@ -50,16 +65,16 @@ arrow::Status ParquetChunkReader::open() {
...
     file_readers_.emplace_back(std::move(result.ValueOrDie()));
 
+    // get the RowGroupMetadataVector
     auto metadata = file_readers_[i]->parquet_reader()->metadata();
-    auto metadata_result = PackedFileMetadata::Make(metadata);
-    if (!metadata_result.ok()) {
-      return arrow::Status::Invalid("Error making file metadata:" + metadata_result.status().ToString());
-    }
-    file_metadatas_.emplace_back(metadata_result.ValueOrDie());
+    assert(metadata);
```
shaoting-huang: can we add error log if assert fails?

PR #317, opened 2025-11-21
enhance: FFI implement ConditionalTransHandler add fill transaction API

jiaqizho:
This commit implements a transhandler based on conditional write, which prevents the manifest from being overwritten at the root level. Additionally, it adjusts some interfaces of the transaction.


congqixia (approved):
/lgtm

PR #328, opened 2025-11-26
enhance: Update FFI tests to use custom test runner and remove libcheck dependency

tedxu:
This PR removes the dependency on libcheck for C FFI tests, replacing it with a custom lightweight test runner. It also includes critical fixes for memory leaks and AddressSanitizer (ASAN) failures identified during the migration and testing process.


cpp/include/milvus-storage/format/parquet/file_reader.h
```diff
@@ -22,6 +22,20 @@ namespace milvus_storage {
 
 class FileRowGroupReader {
```
jiaqizho: question here: `FileRowGroupReader` is called by Milvus. Changing this API would require the changes to Milvus,  Is that expected?

PR #330, opened 2025-11-27
enhance: Introduce exttable_generate_column_groups FFI function

jiaqizho:
Added FFI interface `eexttable_generate_column_groups`, enabling external-table to construct the column groups. Meanwhile, the external-table interfaces in `ffi_c.h` have been moved to the `ffi_exttable_c.h`.

The corresponding test for `exttable_generate_column_groups` can be found in FFI C tests. However, since the storage reader does not yet support reading with start/end indices, only the full file read process is currently tested.


cpp/include/milvus-storage/ffi_exttable_c.h
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * This function creates a ColumnGroups structure containing a single column group
+ * with the specified columns and file paths. It's used for external table support.
+ *
+ * @param columns Array of column names to include in the column group
+ * @param col_lens Number of columns in the columns array
+ * @param format File format (e.g., "parquet", "vortex")
+ * @param paths Array of file paths
+ * @param start_indexs Optional array of start row indices for each file (can be NULL)
+ * @param end_indexs Optional array of end row indices for each file (can be NULL)
```
sunby: explain when start_indexs and end_indexs can be null.

PR #332, opened 2025-11-28
enhance: Chunk reader support start/end indices.

jiaqizho:
In a previous commit (bfbe28c), the column group was modified to support reading with start/end indices. The current commit enables the chunk reader support reading with start/end indices.

`ColumnGroupReaderImpl` is specifically designed to handle the relationship between `ChunkInfo()` and `RowGroupInfo()`. This allows most of the logic to be reused, making the implementation of the format reader simpler. However, vortex may experience read amplification similar to parquet, which can be avoided.

### DIFF
--- a/cpp/src/format/column_group_reader.cpp
+++ b/cpp/src/format/column_group_reader.cpp
@@ -1,0 +1,306 @@
+// Copyright 2023 Zilliz
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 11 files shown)


cpp/src/format/column_group_reader.cpp
```diff
@@ -0,0 +1,305 @@
...
+  ARROW_RETURN_NOT_OK(milvus_storage::ArrowFileSystemConfig::create_file_system_config(properties_, fs_config));
+  ARROW_ASSIGN_OR_RAISE(auto file_system, create_arrow_file_system(fs_config));
+  if (column_group_->format == LOON_FORMAT_PARQUET) {
+    ::parquet::ReaderProperties reader_properties = ::parquet::default_reader_properties();
+    if (key_retriever_) {
+      reader_properties.file_decryption_properties(::parquet::FileDecryptionProperties::Builder()
+                                                       .key_retriever(std::make_shared<KeyRetriever>(key_retriever_))
+                                                       ->plaintext_files_allowed()
+                                                       ->build());
+    }
```
tedxu: Consider move this piece of code to parquet reader `open`.

PR #336, opened 2025-12-04
enhance: add format reader interface to decouple chunk reader logic

jiaqizho:
In previous versions, a format-related reader was required to implement the chunk reader interface. This interface needed to produce a set of format readers and manage the reading in the chunks.

In practice, the two existing readers in storage (parquet and vortex) both had significant redundant logic in chunk management. Therefore, current commit extracts the common chunkreader logic and provides a new format reader interface, allowing each format to focus solely on reading individual files.


cpp/include/milvus-storage/format/parquet/parquet_format_reader.h
```diff
@@ -0,0 +1,70 @@
...
+class ParquetFormatReader final : public FormatReader {
+  public:
+  ParquetFormatReader(const std::shared_ptr<arrow::fs::FileSystem>& fs,
+                      const std::string& path,
+                      const milvus_storage::api::Properties& properties,
+                      const std::vector<std::string>& needed_columns,
+                      const std::function<std::string(const std::string&)>& key_retriever);
+
+  // open the file
+  [[nodiscard]] arrow::Status open() override;
```
tedxu: `open` can be an implementation rather than interface.

PR #337, opened 2025-12-10
fix: credentials provider defined conflict between the storage and milvus

jiaqizho:
Both Milvus and Storage have defined the same C++ credentials provider class (with the same namespace) for different cloud providers. While I'm not entirely sure why a redefinition error doesn’t occur during compilation, it is indeed happening that the stack trace from Storage calls into Milvus’s stack.

problem stack:

```
_ZN3Aws4Auth49AliyunSTSAssumeRoleWebIdentityCredentialsProviderC2Ev /root/milvus/internal/core/src/storage/AliyunCredentialsProvider.cpp:102 pc=0x7f849c5ebf5
_ZNSt15__new_a1locatorIN3Aw$4Auth49AliyunSTSAssumeRoleWebIdentityCredentialsProviderEE9constructIS2_JEEEVPT_DpOTO_ /usr/include/c++/12/bits/new_allocator.h:175 pc=0x7f84967d06df
...
_ZN14milvus_storage20S3FileSystemProducer31CreateAliyunCredentialsProviderEv /root/milvus/cmake_build/thirdparty/milvus-storage/milvus-storage-src/cpp/src/filesystem/$3/$3_fs.cpp:198 pc=0x7f84967d06d
```

Even if the two implementations are identical, they could still be affected by different configurations and cause issues. Therefore, this commit changes the cloud provider namespace defined in Storage to `milvus_storage` to distinguish between the different definitions.

### DIFF
--- a/cpp/include/milvus-storage/filesystem/s3/provider/AliyunCredentialsProvider.h
+++ b/cpp/include/milvus-storage/filesystem/s3/provider/AliyunCredentialsProvider.h
@@ -24,37 +24,37 @@
 #include <memory>
 #include "AliyunSTSClient.h"
 
-namespace Aws {
-namespace Auth {
+namespace milvus_storage {
+
 /**
  * To support retrieving credentials of STS AssumeRole with web identity.
  * Note that STS accepts request with protocol of queryxml. Calling GetAWSCredentials() will trigger (if expired)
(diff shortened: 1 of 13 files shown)


congqixia (approved):
/lgtm

PR #338, opened 2025-12-11
enhance: Support take function in storage

jiaqizho:
This commit adds support for `take` in the FFI and introduces a lazy-loading-compatible which named `column_group_lazy_reader`. Since `take` operations typically only involve a subset of files, the `column_group_lazy_reader` only loads the necessary files for reading. Additionally, to support lazy loading, the field `start_index/end_index` in `ColumnGroupFile` have been changed from optional to required, as we now need to know the row range of each file in order to determine which files should be loaded.


cpp/include/milvus-storage/reader.h
tedxu: Ordered required here.

PR #339, opened 2025-12-11
feat: adding external filesystem support

tedxu:
This PR introduces support for external filesystems in milvus-storage, allowing users to configure and use multiple filesystem backends simultaneously. The implementation provides a unified `FilesystemCache` API that automatically resolves filesystems based on URI schemes and property configurations.

### DIFF
--- a/cpp/include/milvus-storage/properties.h
+++ b/cpp/include/milvus-storage/properties.h
@@ -69,6 +69,8 @@ struct PropertyInfo {
 // --- Define FS property keys ---
 // Standard filesystem properties (fs.*)
 // These configure the default filesystem for the storage instance
+#define PROPERTY_FS_PREFIX "fs."
+#define PROPERTY_EXTFS_PREFIX "extfs."
 #define PROPERTY_FS_ADDRESS "fs.address"
 #define PROPERTY_FS_BUCKET_NAME "fs.bucket_name"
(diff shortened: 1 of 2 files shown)


cpp/src/format/column_group_writer.cpp
```diff
@@ -43,9 +42,9 @@ arrow::Result<std::unique_ptr<ColumnGroupWriter>> ColumnGroupWriter::create(
   auto column_group_schema = arrow::schema(fields);
 
   // create the file system by cache
-  milvus_storage::ArrowFileSystemConfig fs_config;
-  ARROW_RETURN_NOT_OK(milvus_storage::ArrowFileSystemConfig::create_file_system_config(properties, fs_config));
-  ARROW_ASSIGN_OR_RAISE(auto file_system, CreateCachedArrowFileSystem(fs_config));
+  // Use first file path if available, otherwise use empty string for default filesystem
+  std::string path = column_group->files.empty() ? "" : column_group->files[0].path;
+  ARROW_ASSIGN_OR_RAISE(auto file_system, milvus_storage::FilesystemCache::getInstance().get(properties, path));
```
jiaqizho: no need pass the `path`?

tedxu: Let's leave the code there to preserve the way of writing to an external fs.

PR #342, opened 2025-12-15
fix: update is_initialized_ flag after DoInitialize completes

congqixia:
Move the is_initialized_ flag update from before DoInitialize() to after it completes. This ensures the initialization flag is only set to true after the actual initialization is done, preventing potential issues if DoInitialize() fails or if other code checks the flag during initialization.


jiaqizho (approved):
/lgtm
/approve

PR #353, opened 2025-12-24
enhance: localfs automatically creates folders in writer

jiaqizho:
In the storage system, there are both `local` and `remote` filesystems. For remote filesystems, when performing operations such as PutObject or other Put actions, S3 automatically creates the parent directory. However, for the local filesystems, the parent path is not automatically created.

This commit checks whether the filesystem type is local or remote. If it is a local filesystem, the parent path will be automatically created during write operations.

## Summary by CodeRabbit

* **Bug Fixes**
  * Parent directories are now created for local filesystem writes; writing to non-existent local paths returns clear errors on failure.
  * Improved validation to prevent writes when file groups are empty.

* **Refactor**
  * More reliable detection and handling of local vs. remote storage to ensure correct pre-write behavior.

* **Tests**
  * Added test coverage for writing to non-existent paths across local and cloud environments.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/cpp/src/filesystem/fs.cpp
+++ b/cpp/src/filesystem/fs.cpp
@@ -34,6 +34,32 @@ namespace milvus_storage {
 
 static constexpr auto local_uri_scheme = "file://";
 
+enum class StorageType : int8_t {
+  None = 0,
+  Local,
+  Remote,
+};
+
+enum class CloudProviderType : int8_t {
+  UNKNOWN = 0,
+  AWS,
+  GCP,
+  ALIYUN,
+  AZURE,
+  TENCENTCLOUD,
+  HUAWEICLOUD,
+};
+
+static std::map<std::string, StorageType> StorageType_Map = {{"local", StorageType::Local},
+                                                             {"remote", StorageType::Remote}};
+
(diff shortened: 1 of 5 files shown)


jiaqizho:
base on !351

cpp/src/filesystem/fs.cpp
```diff
@@ -88,6 +114,27 @@ arrow::Result<ArrowFileSystemPtr> CreateArrowFileSystem(const ArrowFileSystemCon
...
+  // typename define in arrow::fs::LocalFileSystem
+  if (fs->type_name() == "local") {
+    return true;
+  }
+
+  // typename define in arrow::fs::SubTreeFileSystem
+  if (fs->type_name() == "subtree") {
+    auto subtree_fs = std::dynamic_pointer_cast<arrow::fs::SubTreeFileSystem>(fs);
+    if (subtree_fs) {
+      return IsLocalFileSystem(subtree_fs->base_fs());
```
shaoting-huang: is it possible that subtree fs might not be a local fs?

PR #354, opened 2025-12-24
enhance: support GCS condition write access via ak&sk

tinswzy:
support GCS condition write access via ak&sk

- Core invariant: when initialized with use_iam=false and valid AK/SK, conditional (generation-match) writes to GCS will use GOOG4-HMAC-SHA256 signing; when use_iam=true the existing OAuth2/IAM flow is preserved. This is enforced in GoogleHttpClientFactory::CreateHttpRequest (returns unsigned requests in HMAC mode) and GoogleHttpClientDelegator::MakeRequest (applies signer::goog4::SignRequest only when !use_iam && ak/sk present && request has x-goog-if-generation-match).

- What was simplified/removed: previous workarounds that implicitly ignored or dropped custom object metadata and relied on AWS-auth-only request creation have been replaced by a clear delegator + signer flow and explicit metadata propagation. MultiPartUploadS3FS now accepts optional metadata in OpenConditionalOutputStream and ConditionalOutputStream, stores custom metadata into request->SetHeader/AddMetadata (converted from x-amz-meta-* to x-goog-meta-* in the delegator), eliminating the old silent-drop behavior for unknown metadata keys.

- Why this does NOT cause data loss or regressions: IAM path and behavior are unchanged (CreateHttpRequest still attaches OAuth2 headers when use_iam==true). Metadata handling is optional (metadata_ defaults to nullptr), so callers that do not supply metadata behave identically. The delegator wraps (does not replace) CurlHttpClient and only rewrites headers + signs in the narrow case (non-IAM + AK/SK + conditional write), so all other HTTP flows hit the unchanged underlying client. Concrete code paths: MultiPartUploadS3FS::OpenConditionalOutputStream(s, metadata) -> ConditionalOutputStream(metadata_) sets metadata on PutObject/UploadPart when present; GoogleHttpClientDelegator performs header conversion and calls signer::goog4::SignRequest(request, access_key_, secret_key_) only when conditional write is detected.

- New capability and scope: adds static HMAC (AK/SK) GCS support for conditional writes across the S3-compatible filesystem adapter—implemented by signer::goog4::SignRequest (signer.h/cpp), GoogleHttpClientDelegator/GoogleHttpClientFactory (s3_fs.cpp) and metadata-aware ConditionalOutputStream/OpenConditionalOutputStream overloads (multi_part_upload_s3_fs.*). This enables writing to GCS buckets using HMAC credentials without requiring IAM service accounts.

### DIFF
--- a/cpp/include/milvus-storage/filesystem/s3/s3_auth_signer.h
+++ b/cpp/include/milvus-storage/filesystem/s3/s3_auth_signer.h
@@ -29,4 +29,4 @@ bool SignRequest(const std::shared_ptr<Aws::Http::HttpRequest>& request,
                  const std::string& access_key,
                  const std::string& secret_key);
 
-}  // namespace signer::goog4
+}  // namespace milvus_storage::auth_signer::googv4
--- a/cpp/src/filesystem/filesystem_extend.cpp
+++ b/cpp/src/filesystem/filesystem_extend.cpp
@@ -14,12 +14,12 @@
 
 #include "milvus-storage/filesystem/filesystem_extend.h"
 
+#include <unordered_set>
+#include <string>
+
 #include <arrow/status.h>
 #include <arrow/result.h>
 #include <arrow/util/key_value_metadata.h>
-#include <string>
-#include <unordered_map>
-#include <unordered_set>
 
 #include "milvus-storage/filesystem/fs.h"
 
--- a/cpp/src/filesystem/s3/s3_auth_signer.cpp
+++ b/cpp/src/filesystem/s3/s3_auth_signer.cpp
@@ -112,7 +112,7 @@ static std::string BuildCanonicalQueryString(const Aws::Http::URI& uri) {
 
 // Build canonical headers
 static std::string BuildCanonicalHeaders(const std::shared_ptr<Aws::Http::HttpRequest>& request) {
-  std::unordered_map<std::string, std::string> headers_map;
+  std::map<std::string, std::string> headers_map;
 
   for (const auto& [key, val] : request->GetHeaders()) {
     std::string lower_key = key;
@@ -292,4 +292,4 @@ bool SignRequest(const std::shared_ptr<Aws::Http::HttpRequest>& request,
   return true;
 }
 
-}  // namespace signer::goog4
+}  // namespace milvus_storage::auth_signer::googv4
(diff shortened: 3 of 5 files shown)


cpp/include/milvus-storage/filesystem/s3/signer.h
```diff
@@ -0,0 +1,31 @@
+// Licensed to the LF AI & Data foundation under one
```
jiaqizho: change this file named to `s3_auth_signer.h` ?

cpp/src/filesystem/s3/signer.cpp
```diff
@@ -0,0 +1,294 @@
...
+
+#include <openssl/hmac.h>
+#include <openssl/sha.h>
+#include <algorithm>
+#include <chrono>
+#include <iomanip>
+#include <iostream>
+#include <map>
+#include <sstream>
+#include <vector>
```
jiaqizho: The header order should follow:

img_v3_02rv_7d9900d2-f387-4b25-9b56-4f19be3b493g

change to 
```
#include "milvus-storage/filesystem/s3/signer.h"

#include <algorithm>
#include <chrono>
#include <iomanip>
#include <iostream>
#include <map>
#include <sstream>
#include <vector>

#include <openssl/hmac.h>
#include <openssl/sha.h>
```

cpp/src/filesystem/s3/signer.cpp
```diff
@@ -0,0 +1,294 @@
...
+    if (i > 0)
+      result += "&";
+    result += parts[i];
+  }
+  return result;
+}
+
+// Build canonical headers
+static std::string BuildCanonicalHeaders(const std::shared_ptr<Aws::Http::HttpRequest>& request) {
+  std::map<std::string, std::string> headers_map;
```
jiaqizho: unordered_map

cpp/include/milvus-storage/filesystem/s3/signer.h
```diff
@@ -0,0 +1,31 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include <aws/core/http/HttpRequest.h>
+#include <memory>
+#include <string>
```
jiaqizho:
```
#pragma once

#include <memory>
#include <string>

#include <aws/core/http/HttpRequest.h>
```

tinswzy:
> one more question, do we have tested this PR in `IAM GCP` env?

Yes, it doesn't change the original iam request logic because it only re-signs the request in the ak&sk way

jiaqizho (approved):
/lgtm
/approve

PR #355, opened 2025-12-24
enhance: move multipart-upload-size to filesystem config.

jiaqizho:
In the parquet writer, there has always been the practice of forcibly casting `arrow::fs::FileSystem` to the `class MultiPartUploadS3FS` and then calling methods not provided by `arrow::fs::FileSystem`. This breaks the encapsulation of the Arrow file system.

The current commit removes the forced cast of file system and allows users to decide the multipart upload size when creating the `arrow::fs::FileSystem`.

## Summary: Move Multipart Upload Size to Filesystem Configuration

**Core Invariant:** Multipart upload size must be known at filesystem creation time, not deferred to writer instantiation, to maintain proper Arrow filesystem abstraction boundaries.

**What Was Removed & Why:**
The PR eliminates the writer-time `StorageConfig` parameter and forced downcasting pattern that violated Arrow's filesystem encapsulation:
- Removed `struct StorageConfig` from public API—writers previously accepted this parameter, then performed an unsafe `static_cast<MultiPartUploadS3FS*>` on the generic `arrow::fs::FileSystem` to call `OpenOutputStreamWithUploadSize()` (a non-standard method not in Arrow's API).
- Removed public `OpenOutputStreamWithUploadSize()` overloads from the filesystem interface—these methods are now protected and used only internally.
- Simplified all writer signatures: `ParquetFileWriter::Make()` and `PackedRecordBatchWriter::Make()` no longer accept `storage_config` as a parameter.

**Why No Behavior Regression:**
The upload size is still fully configurable and defaults to the same value (10 MB):
1. **Configuration path preserved:** `ArrowFileSystemConfig::multi_part_upload_size` and `S3Options::multi_part_upload_size` are populated from `PROPERTY_WRITER_MULTI_PART_UPLOAD_SIZE` at filesystem creation time (verified in `fs.cpp` and `s3_fs.cpp`).
2. **Default unchanged:** `DEFAULT_MULTIPART_UPLOAD_PART_SIZE = 10 * 1024 * 1024` remains identical to the prior hardcoded default.
3. **Functionality preserved:** `MultiPartUploadS3FS::OpenOutputStream()` now retrieves upload size directly from `impl_->options().multi_part_upload_size` instead of expecting it as a runtime parameter.

**Concrete Fix:** Users now configure multipart behavior at S3 filesystem instantiation (via properties) rather than passing it through every writer, removing the need for unsafe downcasts and keeping writers filesystem-agnostic.

### DIFF
--- a/cpp/include/milvus-storage/common/config.h
+++ b/cpp/include/milvus-storage/common/config.h
@@ -27,9 +27,8 @@ static constexpr int64_t DEFAULT_READ_BATCH_SIZE = 1024;
 static constexpr int64_t DEFAULT_READ_BUFFER_SIZE = 16 * 1024 * 1024;   // 16 MB
 static constexpr int64_t DEFAULT_WRITE_BUFFER_SIZE = 16 * 1024 * 1024;  // 16 MB
 
-struct StorageConfig {
-  int64_t part_size = 0;
-};
+// Default part size for multi-part upload
+#define DEFAULT_MULTIPART_UPLOAD_PART_SIZE (10 * 1024 * 1024)  // 10 MB
 
 #define LOON_FORMAT_PARQUET "parquet"
 #define LOON_FORMAT_VORTEX "vortex"
--- a/cpp/src/properties.cpp
+++ b/cpp/src/properties.cpp
@@ -446,11 +446,15 @@ static std::unordered_map<std::string, PropertyInfo> property_infos = {
                       "The buffer size(Bytes) used in the writer.",
                       32 * 1024 * 1024,  // 32MB
                       ValidatePropertyType()),
-    REGISTER_PROPERTY(PROPERTY_WRITER_MULTI_PART_UPLOAD_SIZE,
-                      PropertyType::INT32,
-                      "The multi-part upload size(Bytes) used in the writer.",
-                      10 * 1024 * 1024,  // 10MB
-                      ValidatePropertyType()),
+    REGISTER_PROPERTY(
+        PROPERTY_WRITER_MULTI_PART_UPLOAD_SIZE,
+        PropertyType::UINT64,
+        "The multi-part upload size(Bytes) used in the writer.",
+        uint64_t(DEFAULT_MULTIPART_UPLOAD_PART_SIZE),  // 10 MB
+        // According to https://docs.aws.amazon.com/AmazonS3/latest/userguide/qfacts.html
+        // the part numbers can be 5MB ~ 5GB
+        // R2 limit is 10MB ~ 5GB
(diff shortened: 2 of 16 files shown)


cpp/src/properties.cpp
```diff
@@ -447,10 +447,11 @@ static std::unordered_map<std::string, PropertyInfo> property_infos = {
...
                       ValidatePropertyType()),
     REGISTER_PROPERTY(PROPERTY_WRITER_MULTI_PART_UPLOAD_SIZE,
-                      PropertyType::INT32,
+                      PropertyType::UINT32,
                       "The multi-part upload size(Bytes) used in the writer.",
-                      10 * 1024 * 1024,  // 10MB
-                      ValidatePropertyType()),
+                      uint32_t(DEFAULT_MULTIPART_UPLOAD_PART_SIZE),  // 10MB
+                      ValidatePropertyType() + ValidatePropertyRange<uint32_t>(1024,
+                                                                               1024 * 1024 * 1024)),  // 1kb ~ 1GB
```
shaoting-huang: https://docs.aws.amazon.com/AmazonS3/latest/userguide/qfacts.html, change to 5mb - 5GB?

cpp/include/milvus-storage/common/config.h
```diff
@@ -27,9 +27,8 @@ static constexpr int64_t DEFAULT_READ_BATCH_SIZE = 1024;
 static constexpr int64_t DEFAULT_READ_BUFFER_SIZE = 16 * 1024 * 1024;   // 16 MB
 static constexpr int64_t DEFAULT_WRITE_BUFFER_SIZE = 16 * 1024 * 1024;  // 16 MB
 
-struct StorageConfig {
-  int64_t part_size = 0;
-};
+// Default part size for multi-part upload
```
shaoting-huang: TODO: We should also change from Milvus side once we remove storage config

cpp/test/s3_client_metrics_test.cpp
```diff
@@ -80,11 +87,11 @@ TEST_F(S3ClientMetricsTest, TestMetricsAfterFileOperations) {
   std::string test_file_name = GenerateTestFileName();
 
   // Create some test data (large enough to trigger multipart upload)
-  std::string test_data(6 * 1024 * 1024, 'A');  // 6MB of data
+  std::string test_data(11 * 1024 * 1024, 'A');  // 11MB of data
```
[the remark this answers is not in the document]
jiaqizho: Old test use the dynamic cast and call the function `OpenOutputStreamWithUploadSize`, which set the multipart upload size to 5mb.

cpp/src/properties.cpp
jiaqizho: ack

PR #356, opened 2025-12-25
enhance: Introduce s3_extend to provide extensibility outside the filesystem

jiaqizho:
This commit completely removes `ExtendFileSystem` and modifies the S3 file system's handling of metadata.

- Metadata now allows callers pass the custom metadata, which will be automatically added to headers.
- Since AWS-SDK metadata is set via method calls rather than a map, The `MetadataSetters` now use SFINAE to determine which metadata can be set for different types of requests.

The `s3_extend` leverages metadata to extend the functionality provided by the S3 file system, such as conditional writes and other features.

- Core invariant: conditional-write behavior is moved from a runtime-only interface (ExtendFileSystem + dynamic_cast) to a metadata-driven model — callers provide KeyValueMetadata and the library detects request-level metadata setters (SFINAE-based MetadataSetters) and provider type via GetFileSystemTypeName(); open_condition_write_output_stream() injects provider-specific conditional headers and delegates to OpenOutputStream(path, metadata).

- Removed/simplified logic: eliminated ExtendFileSystem, IsExtendFileSystem(), OpenConditionalOutputStream(), ConditionalOutputStream class, and runtime dynamic_cast checks. These were redundant because all conditional-write semantics can be expressed as metadata sent with standard OpenOutputStream calls; open_condition_write_output_stream centralizes provider branching and reuses the existing filesystem API instead of extending class hierarchies.

- No data loss or regression: SetObjectMetadata now preserves known keys via detected setters and preserves unknown keys by calling SetAdditionalCustomHeaderValue so all caller metadata is forwarded. Multipart upload finalization uses SetMetadataInRequest so metadata persists through completion. Transactional conditional-write tests and error mappings remain (first conditional write succeeds; subsequent writes fail and surface AlreadyExists), preserving prior observable behavior.

- New capability: s3_extend (open_condition_write_output_stream + SFINAE MetadataSetters) enables caller-provided, provider-specific extensions outside filesystem implementations (e.g., AWS If-None-Match, GCP x-goog-if-generation-match, Tencent/Aliyun overwrite-forbid headers, Azure If-None-Match) by augmenting metadata passed to OpenOutputStream — adding extensibility without changing filesystem class hierarchies.

### DIFF
--- a/cpp/include/milvus-storage/common/config.h
+++ b/cpp/include/milvus-storage/common/config.h
@@ -19,13 +19,12 @@
 namespace milvus_storage {
 
 // current config will be used in packed writer/reader layer
-// TODO: will move it to properties.h in the future
-static constexpr int64_t DEFAULT_MAX_ROW_GROUP_SIZE = 1024 * 1024;  // 1 MB
+inline constexpr int64_t DEFAULT_MAX_ROW_GROUP_SIZE = 1024 * 1024;  // 1 MB
 
 // Default number of rows to read when using ::arrow::RecordBatchReader
-static constexpr int64_t DEFAULT_READ_BATCH_SIZE = 1024;
-static constexpr int64_t DEFAULT_READ_BUFFER_SIZE = 16 * 1024 * 1024;   // 16 MB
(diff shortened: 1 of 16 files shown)


tinswzy (approved):
/lgtm

PR #364, opened 2026-01-05
feat: reduce parquet footer size

tedxu:
This commit introduces a new benchmark file, `benchmark_footer_size.cpp`, which includes functionality to measure the footer size of Parquet files. It defines a schema and generates test data for benchmarking, while also implementing a fixture for setup and teardown processes. The benchmark reports footer size, file size, and footer size as a percentage of the file size, enhancing performance analysis capabilities for Parquet file handling.

Additionally, the `parquet_writer.cpp` file has been updated to disable statistics for vector columns, ensuring that the writing process remains efficient for fixed-size binary and binary types.


congqixia (approved):
/lgtm

PR #367, opened 2026-01-06
fix: fix use-after-free in FFIResult error handling

shaoting-huang:
Swap order of FreeFFIResult and ThrowJavaExceptionFromFFIResult to prevent accessing freed memory. Previously FreeFFIResult was called before ThrowJavaExceptionFromFFIResult, causing the error message pointer to be freed before being used, leading to SIGSEGV crashes.


jiaqizho (approved):
/lgtm
/approve

PR #370, opened 2026-01-08
enhance: Files created in storage are stored using relative paths.

jiaqizho:
In Storage, there are internal tables and external tables. Internal tables contain data generated via `write.h`, while external tables contain imported data.

- The `ColumnGroupFiles.path` written by internal tables will always be stored using relative paths.
- The `ColumnGroupFiles.path` generated by external tables will always be stored using absolute paths.

This process occurs during the serialization and deserialization of column groups. During serialization, the path in the `ColumnGroupFile` of an internal table has the `{base_path}/{data_dir_prefix}` removed before being written into Avro. During deserialization, the system checks whether the path in the current `ColumnGroupFile` is a relative path. If it is a relative path, the `{base_path}/{data_dir_prefix}` will be appended to the path. Therefore, the `ColumnGroupFile` in memory always contains an absolute path.

### DIFF
--- a/cpp/include/milvus-storage/column_groups.h
+++ b/cpp/include/milvus-storage/column_groups.h
@@ -49,4 +49,16 @@ struct ColumnGroup {
  */
 using ColumnGroups = std::vector<std::shared_ptr<ColumnGroup>>;
 
+static ColumnGroups copy_column_groups(const ColumnGroups& source) {
+  ColumnGroups dest(source.size());
+  for (size_t i = 0; i < source.size(); ++i) {
+    if (source[i]) {
+      dest[i] = std::make_shared<ColumnGroup>(*source[i]);
+    } else {
+      dest[i] = nullptr;
+    }
+  }
+  return dest;
+}
+
 }  // namespace milvus_storage::api
--- a/cpp/include/milvus-storage/common/layout.h
+++ b/cpp/include/milvus-storage/common/layout.h
@@ -16,6 +16,8 @@
 
 #include <string>
 
+#include "milvus-storage/common/path_util.h"
+
 namespace milvus_storage {
 
 /**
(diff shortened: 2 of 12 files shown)


cpp/src/column_groups.cpp
```diff
@@ -36,7 +38,42 @@ ColumnGroups::ColumnGroups(std::vector<std::shared_ptr<ColumnGroup>> column_grou
...
-arrow::Result<std::string> ColumnGroups::serialize() const {
+static inline bool IsPathEq(const std::string& left, const std::string& right) {
+  std::filesystem::path lp(left);
+  std::filesystem::path rp(right);
+  return (lp / "").lexically_normal() == (rp / "").lexically_normal();
+}
+
+static inline std::string AbsoluteDataPathToRelative(const std::string& path,
+                                                     const std::optional<std::string>& base_path) {
+  if (!base_path.has_value()) {
```
shaoting-huang: should handle case like trailing slashes

cpp/src/column_groups.cpp
```diff
@@ -36,7 +38,42 @@ ColumnGroups::ColumnGroups(std::vector<std::shared_ptr<ColumnGroup>> column_grou
...
+    return path;
+  }
+
+  std::string data_path = get_data_path(base_path.value());
+  if (path.size() >= data_path.size() && path.substr(0, data_path.size()) == data_path) {
+    return path.substr(data_path.size());
+  }
+
+  // external table keep the absolute path
+  return path;
```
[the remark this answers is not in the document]
jiaqizho: won't be, all of path will base on "base_path".

all of path in storage will generated in layout.h, won't create the "../" path in this interface

cpp/src/column_groups.cpp
jiaqizho: used `substr`, 

ex.
```
path="base_dir/_data/xxxx.parquet"
base_path="base_dir"
data_path="base_dir/_data" // or data_path="base_dir/_data/" is both ok.
```
because 
```
if (path.size() >= data_path.size() && path.substr(0, data_path.size()) == data_path) {
```

PR #374, opened 2026-01-13
enhance: Implement file rolling for Writer

jiaqizho:
In the storage, files generated from a single write operation without rolling. Even though a column-splitting logic already exists (different columns are stored in separate files). But for compactions triggered by Milvus, this can still lead to the creation of very large individual files (exceeding 2GB). This increases the likelihood of `import/export` failures when multi-part upload is not enabled.

This commit introduces a new property (PROPERTY_WRITER_FILE_ROLLING_SIZE, default 2GB). It allows a single write operation to split files based on the specified size limit of uncompressed data written.

### DIFF
--- a/cpp/src/format/column_group_writer.cpp
+++ b/cpp/src/format/column_group_writer.cpp
@@ -21,6 +21,7 @@
 #include <sstream>
 #include <filesystem>
 
+#include "milvus-storage/common/arrow_util.h"
 #include "milvus-storage/common/layout.h"
 #include "milvus-storage/common/path_util.h"  // for kSep
 #include "milvus-storage/filesystem/fs.h"
@@ -32,7 +33,6 @@
 
 namespace milvus_storage::api {
 
-// TODO(jiaqizho): implements file rolling in this class
 class ColumnGroupWriterImpl final : public ColumnGroupWriter {
   public:
   ColumnGroupWriterImpl(const std::string& base_path,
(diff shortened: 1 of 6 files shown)


cpp/src/format/column_group_writer.cpp
```diff
@@ -44,24 +44,49 @@ class ColumnGroupWriterImpl final : public ColumnGroupWriter {
...
+
+    return arrow::Status::OK();
+  }
 
   [[nodiscard]] arrow::Result<std::vector<ColumnGroupFile>> Close() override {
+    assert(format_writer_);
     ARROW_ASSIGN_OR_RAISE(auto column_group_file, format_writer_->Close());
-    return std::vector<ColumnGroupFile>{std::move(column_group_file)};
+    // If current column_group_file without any data, do not add it to written_files_
+    if (written_bytes_ != 0) {
```
shaoting-huang: if all written bytes is zero, does it create a empty file?

cpp/src/format/column_group_writer.cpp
```diff
@@ -44,24 +44,49 @@ class ColumnGroupWriterImpl final : public ColumnGroupWriter {
...
   [[nodiscard]] arrow::Status Write(const std::shared_ptr<arrow::RecordBatch> record) override {
+    written_bytes_ += GetRecordBatchMemorySize(record);
     return format_writer_->Write(record);
   }
 
-  [[nodiscard]] arrow::Status Flush() override { return format_writer_->Flush(); }
+  [[nodiscard]] arrow::Status Flush() override {
+    ARROW_RETURN_NOT_OK(format_writer_->Flush());
+
+    if (written_bytes_ == 0 || written_bytes_ < max_bytes_limit_) {
```
shaoting-huang: it seems to roll every time during flush if max_bytes_limit is zero. Does it the expected behavior?

PR #385, opened 2026-01-16
fix: use calloc to zero-initialize ArrowArrayStream

shaoting-huang:
malloc returns uninitialized memory. If get_record_batch_reader fails, the cleanup code checks stream->release and may call a garbage pointer, causing core dump. calloc zero-initializes memory, ensuring release is NULL and preventing the crash.


jiaqizho (approved):
/lgtm
/approve

PR #386, opened 2026-01-16
fix: add delta and stats directories in layout

tedxu:
Added new directories for delta logs and stats in layout documentation and updated related functions to support these changes

### DIFF
--- a/cpp/Makefile
+++ b/cpp/Makefile
@@ -79,7 +79,7 @@ clean:
 	rm -rf build
 
 test: build
-	cd build/Release/test && ASAN_OPTIONS=detect_container_overflow=0 ./milvus_test && ASAN_OPTIONS=detect_container_overflow=0 ./Test_FFI
+	cd build/Release/test && ./milvus_test && ./Test_FFI
 
 # Test cloud storage with all providers or a specific one
 # Usage: make test-cloud-storage     # Test all cloud providers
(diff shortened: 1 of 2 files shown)


cpp/src/transaction/transaction.cpp
```diff
@@ -442,18 +442,11 @@ arrow::Result<int64_t> Transaction::get_latest_version() {
...
-  std::vector<std::string> file_names;
-  {
-    ARROW_ASSIGN_OR_RAISE(auto file_infos_result, fs_->GetFileInfo(selector));
-
-    file_names.reserve(file_infos_result.size());
-    for (const auto& file_info : file_infos_result) {
-      file_names.push_back(file_info.base_name());
-    }
-  }
+  ARROW_ASSIGN_OR_RAISE(auto file_infos_result, fs_->GetFileInfo(selector));
```
jiaqizho: nit: `file_infos_result` -> `file_infos `

cpp/Makefile
```diff
@@ -79,7 +79,7 @@ clean:
 	rm -rf build
 
 test: build
-	cd build/Release/test && ./milvus_test && ./Test_FFI
+	cd build/Release/test && ASAN_OPTIONS=detect_container_overflow=0 ./milvus_test && ASAN_OPTIONS=detect_container_overflow=0 ./Test_FFI
```
jiaqizho: Only mac will got this false alarm? 

Then we can use 
1. `ASAN_OPTIONS=detect_container_overflow=0 make test`  in mac
2. Or add a `ifeq ($(uname_s),Darwin)` check?

PR #390, opened 2026-01-20
feat: add observable support to local filesystem

tedxu:
This commit introduces the `LocalFileSystemProducer` class, which encapsulates the logic for creating a local filesystem instance. It replaces the previous implementation in `fs.cpp` with a more modular approach, improving code clarity and maintainability. Additionally, the `LocalFileSystemWrapper` class is added to track filesystem metrics, enhancing observability of read and write operations.

### DIFF
--- a/cpp/include/milvus-storage/filesystem/observable.h
+++ b/cpp/include/milvus-storage/filesystem/observable.h
@@ -18,6 +18,11 @@
 #include <cstdint>
 #include <memory>
 
+#include <arrow/buffer.h>
+#include <arrow/io/interfaces.h>
+#include <arrow/result.h>
+#include <arrow/status.h>
+
 namespace milvus_storage {
 
 /// \brief Metrics collection for filesystem operations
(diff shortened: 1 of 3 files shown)


cpp/include/milvus-storage/filesystem/observable.h
```diff
@@ -141,4 +146,108 @@ class Observable {
   [[nodiscard]] virtual std::shared_ptr<FilesystemMetrics> GetMetrics() const = 0;
 };
 
+/// \brief Wrapper for InputStream to track read bytes
+class MetricsInputStream : public arrow::io::InputStream {
```
jiaqizho: Is it unnecessary to define it in the header?

cpp/include/milvus-storage/filesystem/observable.h
```diff
@@ -141,4 +146,108 @@ class Observable {
...
+  bool closed() const override { return stream_->closed(); }
+  arrow::Result<int64_t> Tell() const override { return stream_->Tell(); }
+
+  private:
+  std::shared_ptr<arrow::io::InputStream> stream_;
+  std::shared_ptr<FilesystemMetrics> metrics_;
+};
+
+/// \brief Wrapper for RandomAccessFile to track read bytes
+class MetricsRandomAccessFile : public arrow::io::RandomAccessFile {
```
jiaqizho: Is it unnecessary to define it in the header?

PR #404, opened 2026-01-27
feat: integrate libfiu fault injection for IT Phase 2

jiaqizho:
Brought in libfiu to simulate failures in tests — write errors, read errors, S3 going down, that sort of thing.

The key idea is that fiu_local.h is the single source of truth for all fault point names (FIUKEY_* macros). C++ code uses these macros directly, and Python reads the same strings via extern const char* through FFI, so no one can typo a fault point name.

On the C++ side, added FIU_RETURN_ON / FIU_DO_ON injection points in multi code(contains writer/reader and filesystem ...) . Also added a set of gtests (fiu_test.cpp) covering some fault scenarios. The FFI layer exposes four new functions: loon_fiu_enable/disable/disable_all/is_enabled. The enable function takes a one_time flag to choose between single-shot and always-on modes. Both Linux and macOS export maps are updated with all 4 functions and multi key symbols.

On the Python side, everything is wrapped in a FaultInjector class with context manager support. All keys are loaded dynamically from the C library at runtime — no hardcoded strings.

Build-wise, CMakeLists.txt pulls libfiu via FetchContent, gated behind WITH_FIU=ON. The Makefile got new build-fiu and test-fiu targets.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -12,6 +12,7 @@ option(WITH_BENCHMARK "Build with micro benchmark." ON)
 option(WITH_AZURE_FS "Build with azure file system." ON)
 option(WITH_JNI "Build with JNI library." OFF)
 option(WITH_PYTHON_BINDING "Build for Python bindings with hidden symbols." OFF)
+option(WITH_FIU "Build with fault injection support (libfiu)." OFF)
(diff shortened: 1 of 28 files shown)


jiaqizho:
!407 merged, closed

PR #406, opened 2026-01-29
enhance: add comprehensive benchmark framework for multi-format comparison

jiaqizho:
This adds a benchmark framework to compare Parquet, Vortex, and Lance formats.

The benchmarks cover the following dimensions:

1. Format layer write performance and compression analysis
2. Format layer read including full scan, column projection, and random access (take)
3. V2 packed API vs V3 high-level API for both read and write
4. Storage layer with transactions comparing MilvusStorage (Parquet/Vortex) against Lance native API
5. Multi-reader concurrency under different thread pool sizes

The benchmarks report throughput (MB/s and rows/s), compression ratio, file size, and peak memory usage. Also for the `Multi-reader concurrency`, the benchark report the number of thread to make sure resource is under control.

Added a "Typical" benchmark suite with representative parameters for quick validation. Use --benchmark_filter="Typical/" to run a fast sanity check instead of the full matrix.

### DIFF
--- a/cpp/benchmark/benchmark_format_common.h
+++ b/cpp/benchmark/benchmark_format_common.h
@@ -1,0 +1,575 @@
+// Copyright 2024 Zilliz
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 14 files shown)


cpp/benchmark/benchmark_format_common.h
```diff
@@ -0,0 +1,575 @@
...
+      return "Unknown";
+  }
+}
+
+//=============================================================================
+// Format Utilities
+//=============================================================================
+
+// Get list of available formats based on build configuration
+inline std::vector<std::string> GetAvailableFormats() { return {LOON_FORMAT_PARQUET, LOON_FORMAT_VORTEX}; }
```
tedxu: Missing of `lance` support?

PR #407, opened 2026-01-30
feat: add comprehensive Python integration and stress tests

jiaqizho:
This commit fills in all the integration tests we planned for IT Phase 3 && 4. We now have complete test coverage for write/read operations including boundary conditions, chunk reader, compression codecs, various data types, error handling, file rolling, and writer policies.

Transaction tests cover append workflows, schema evolution via add_column_group, concurrent operations, and mixed read/write scenarios. The concurrent tests verify that exactly one transaction succeeds when multiple threads race to commit.

Recovery tests use libfiu for fault injection to simulate crashes during manifest writes and verify the system recovers correctly.

Stress tests exercise the system under load with large row counts (up to 100M rows), wide tables, high concurrency (up to 100 threads), and long-running operations. All stress test parameters are scalable via --stress-scale flag for quick validation runs.

Also added tests/requirements.txt for test dependencies, updated CLAUDE.md and README.md with integration test instructions, and fixed some Python lint issues across all test files.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -43,6 +43,22 @@ pip install -e ".[dev]"
 pytest tests/ -v
 ```
 
+Integration Tests (`/tests`):
+```bash
+# Build C++ library for Python FFI
+cd cpp && make python-lib && cd ..
+
+# Install Python FFI package and test dependencies
+cd python && pip install -e ".[dev]" && cd ..
+pip install -r tests/requirements.txt
+
+# Run integration tests
+cd tests && pytest integration/ -v
+
+# Run stress tests (with scale factor for quick validation)
+cd tests && pytest stress/ --stress-scale=0.01 -v
+```
+
 Rust (`/rust`):
 ```bash
 cargo build --release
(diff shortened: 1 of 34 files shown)


cpp/src/ffi/ffi_fiu_c.cpp
```diff
@@ -0,0 +1,163 @@
...
+    if (ret != 0) {
+      // fiu_disable returns -1 if the point was not enabled, which is not an error
+      // Just ignore this case
+    }
+
+    RETURN_SUCCESS();
+  } catch (const std::exception& e) {
+    RETURN_ERROR(LOON_GOT_EXCEPTION, e.what());
+  }
+}
```
congqixia: NIT: add mutex protection in case of racing condition

jiaqizho: The `FIU_ENABLE_FAULT_*` and `FIU_DISABLE_FAULT` is thread-safe. This is guaranteed by the internal implementation of `libfiu`.

And `ensure_fiu_init` used the `std::call_once` to make sure thread-safe.

PR #409, opened 2026-02-05
 feature: add text column support with LOB storage

zhagnlu:
This adds support for storing large text columns separately from the main columnar data using a LOB (Large Object) storage layer.

The core idea is that text data can be either inlined (for small values under 64KB) or stored in dedicated LOB files, with encoded references kept in the main parquet files. This avoids bloating the columnar storage with large variable-length text.

The main pieces are: a LobColumnManager/Writer/Reader that handles the actual LOB file I/O, a LobReference encoding scheme that distinguishes inline vs file-based data, and a SegmentWriter/SegmentReader layer that transparently intercepts text columns during read/write and routes them through the LOB path. The manifest is bumped to v5 to track LOB file metadata (path, field_id, row counts, file size) for GC and query planning.

FFI (C) and JNI (Java/Scala) bindings are included so the Go and Spark layers can use the new segment reader/writer directly. The Scala side adds MilvusSegmentReader and MilvusSegmentWriter wrappers.

Tests cover the LOB layer (manager, writer, reader, reference encoding), segment-level text round-trips, edge cases like empty strings and large batches, and a cloud storage integration test.

### DIFF
--- a/cpp/include/milvus-storage/common/layout.h
+++ b/cpp/include/milvus-storage/common/layout.h
@@ -58,6 +58,9 @@ inline const std::string kDataPath = kDataDir + kSep;
 inline const std::string kDeltaPath = kDeltaDir + kSep;
 inline const std::string kStatsPath = kStatsDir + kSep;
 inline const std::string kIndexPath = kIndexDir + kSep;
+// LOB files live at partition level (sibling to {seg_id}/), so from the segment
+// base_path ({partition}/{seg_id}) we go up one level: {partition}/{seg_id}/../lobs/
+inline const std::string kLobPath = std::string("..") + kSep + "lobs" + kSep;
 
 std::string get_manifest_path(const std::string& base_path);
 std::string get_manifest_filename(const size_t& version);
--- a/cpp/include/milvus-storage/ffi_c.h
+++ b/cpp/include/milvus-storage/ffi_c.h
@@ -220,6 +220,26 @@ typedef struct LoonStatsLog {
 } LoonStatsLog;
 
 /**
+ * @brief C structure representing a single LOB file info
+ * Used for TEXT column compaction strategy decision (hole ratio calculation)
+ */
+typedef struct LoonLobFileInfo {
+  const char* path;         ///< Relative path to the LOB file
+  int64_t field_id;         ///< Field ID this LOB file belongs to
+  int64_t total_rows;       ///< Total number of rows in the LOB file
+  int64_t valid_rows;       ///< Number of valid (non-deleted) rows
+  int64_t file_size_bytes;  ///< Size of the LOB file in bytes
+} LoonLobFileInfo;
+
+/**
+ * @brief C structure representing LOB files in manifest
+ */
+typedef struct LoonLobFiles {
+  LoonLobFileInfo* files;  ///< Array of LOB file info
+  uint32_t num_files;      ///< Number of LOB files
+} LoonLobFiles;
+
+/**
  * @brief C structure representing a Manifest
  */
 typedef struct LoonManifest {
@@ -231,6 +251,9 @@ typedef struct LoonManifest {
 
   // Stats
   LoonStatsLog stats;
+
+  // LOB files (TEXT column metadata)
+  LoonLobFiles lob_files;
 } LoonManifest;
 
 /**
@@ -714,6 +737,16 @@ FFI_EXPORT LoonFFIResult loon_transaction_update_stat(LoonTransactionHandle hand
                                                       const char* const* metadata_values,
                                                       size_t metadata_len);
 
+/**
+ * @brief Add a LOB file info to the transaction updates
+ * Used during compaction REUSE_ALL mode to merge LOB file references
+ *
+ * @param handle Transaction handle
+ * @param lob_file LOB file info to add
+ * @return result of FFI
+ */
+FFI_EXPORT LoonFFIResult loon_transaction_add_lob_file(LoonTransactionHandle handle, const LoonLobFileInfo* lob_file);
+
 // ==================== End of Manifest C Interface ====================
 
 // ==================== Test-only Interface ====================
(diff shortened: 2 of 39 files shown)


cpp/include/milvus-storage/text_column/lob_reference.h
```diff
@@ -0,0 +1,115 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include <cstdint>
+#include <string>
+#include <vector>
+
+namespace milvus_storage::text_column {
```
zhengbuqian: how about rename this namespace to lob_column? text is only a type of lob.

zhengbuqian: same as the folder name

cpp/src/text_column/text_column_writer.cpp
```diff
@@ -0,0 +1,395 @@
...
+    }
+
+    arrow::BinaryBuilder builder;
+    ARROW_RETURN_NOT_OK(builder.Reserve(texts->length()));
+
+    for (int64_t i = 0; i < texts->length(); i++) {
+      if (texts->IsNull(i)) {
+        // preserve null values - write null to output array
+        ARROW_RETURN_NOT_OK(builder.AppendNull());
+        stats_.total_texts++;
```
zhengbuqian: should we increment written_rows_ for null rows?

cpp/src/text_column/text_column_writer.cpp
```diff
@@ -0,0 +1,395 @@
...
+    // check if we need to roll to a new file after flush
+    if (current_file_bytes_ >= config_.max_lob_file_bytes) {
+      ARROW_RETURN_NOT_OK(RollToNewFile());
+    }
+
+    return arrow::Status::OK();
+  }
+
+  // roll to a new LOB file
+  arrow::Status RollToNewFile() {
```
zhengbuqian: both claude and gemini suggest to avoid `RollToNewFile` and `FlushPendingTexts` to have recursive calls, which is correct currently but fragile to changes. is it possible to refactor to make the code more robust?

cpp/src/segment/segment_writer.cpp
```diff
@@ -0,0 +1,336 @@
...
+  std::vector<int> lob_column_indices_;
+
+  bool closed_;
+  int64_t written_rows_;
+
+  // Transaction for manifest management
+  std::unique_ptr<api::transaction::Transaction> transaction_;
+
+  // TEXT column writers, keyed by column index
+  std::map<int, std::unique_ptr<lob_column::LobColumnWriter>> lob_writers_;
```
zhengbuqian: can we use unordered_map instead? not seeing why we need the ordering.

cpp/src/lob_column/lob_column_writer.cpp
```diff
@@ -0,0 +1,378 @@
...
+    arrow::BinaryBuilder builder;
+    ARROW_RETURN_NOT_OK(builder.Reserve(texts->length()));
+
+    for (int64_t i = 0; i < texts->length(); i++) {
+      if (texts->IsNull(i)) {
+        ARROW_RETURN_NOT_OK(builder.AppendNull());
+        stats_.total_texts++;
+        written_rows_++;
+      } else {
+        auto text = texts->GetString(i);
```
zhengbuqian: text is copied here. use GetView?

cpp/src/lob_column/lob_reference.cpp
```diff
@@ -0,0 +1,87 @@
...
+#include <cstring>
+#include <filesystem>
+
+#include <boost/uuid/uuid.hpp>
+#include <boost/uuid/uuid_generators.hpp>
+#include <boost/uuid/uuid_io.hpp>
+
+namespace milvus_storage::lob_column {
+
+std::vector<uint8_t> EncodeInlineText(const std::string& text) {
```
zhengbuqian: this `std::vector<uint8_t>` is always of 24 bytes, with the data part allocated on heap and the overhead of vector. let's create a struct with 24 bytes length, to avoid the vector's overhead and to return the result in stack directly.

cpp/include/milvus-storage/lob_column/lob_reference.h
```diff
@@ -0,0 +1,125 @@
...
+ */
+std::string DecodeInlineText(const uint8_t* data, size_t data_size);
+
+/**
+ * @brief decode LOB reference from encoded data
+ * @param data encoded bytes (must be at least 24 bytes)
+ * @param file_id_str output: UUID string (36 chars), converted from binary
+ * @param row_offset output: row offset
+ */
+void DecodeLOBReference(const uint8_t* data, std::string* file_id_str, int32_t* row_offset);
```
zhengbuqian: std::string* file_id_str

can also create a 16B struct

cpp/src/transaction/transaction.cpp
```diff
@@ -221,6 +226,12 @@ arrow::Result<std::shared_ptr<Manifest>> applyUpdates(const std::shared_ptr<Mani
     resolved_stats[key] = stat;  // Override existing or add new
   }
 
+  // Prepare LOB files (copy from base + add new ones)
+  std::vector<LobFileInfo> resolved_lob_files = base_lob_files;
```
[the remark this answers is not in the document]
zhagnlu: solve it in next line

cpp/include/milvus-storage/common/layout.h
```diff
@@ -58,6 +58,9 @@ inline const std::string kDataPath = kDataDir + kSep;
 inline const std::string kDeltaPath = kDeltaDir + kSep;
 inline const std::string kStatsPath = kStatsDir + kSep;
 inline const std::string kIndexPath = kIndexDir + kSep;
+// LOB files live at partition level (sibling to {seg_id}/), so from the segment
+// base_path ({partition}/{seg_id}) we go up one level: {partition}/{seg_id}/../lobs/
+inline const std::string kLobPath = std::string("..") + kSep + "lobs" + kSep;
```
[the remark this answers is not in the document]
zhagnlu: `kLobPath = "../lobs/"` is only used during manifest serialization/deserialization, and it must be used after `lexically_normal()`. not  related with S3/GCP/Azure

cpp/include/milvus-storage/ffi_c.h
```diff
@@ -723,6 +756,101 @@ FFI_EXPORT LoonFFIResult loon_transaction_update_stat(LoonTransactionHandle hand
...
+                                                 LoonSegmentWriterHandle* out_handle);
+
+/**
+ * @brief Writes a record batch to the segment
+ *
+ * @param handle SegmentWriter handle
+ * @param array Arrow array representing the record batch
+ * @return result of FFI
+ */
+FFI_EXPORT LoonFFIResult loon_segment_writer_write(LoonSegmentWriterHandle handle, struct ArrowArray* array);
```
jiaqizho: current symbol also need export to `ffi_exports.map` and `ffi_exports_mac.map`. which python used.

Also need define the relative python object in `python/milvus_storage`. which ray-milvus will use it.

PR #415, opened 2026-02-10
fix: use std::call_once to eliminate TOCTOU race in S3 initialization

jiaqizho:
The old code checks IsS3Initialized() then calls InitializeS3(), but
  there's a window between the check and the call where another thread
  can slip through. When that happens the second thread gets back
  "S3 was already initialized" as an Invalid status, which we treat as
  a hard error and throw. This crashes Spark tasks that hit InitS3()
  concurrently through the FilesystemCache -> S3FileSystemProducer path.

  Replace the if-guard with a static std::once_flag and std::call_once
  so the init block runs exactly once and all other threads just wait
  for it to finish. This matches the same pattern already used inside
  AwsInstance::EnsureInitialized() in s3_global.cpp.

### DIFF
--- a/cpp/src/filesystem/s3/s3_filesystem_producer.cpp
+++ b/cpp/src/filesystem/s3/s3_filesystem_producer.cpp
@@ -16,6 +16,7 @@
 #include "milvus-storage/filesystem/fs.h"
 
 #include <cstdlib>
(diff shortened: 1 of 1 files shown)


congqixia (approved):
/lgtm

PR #426, opened 2026-02-26
enhance: do the sync read  when parallelism hint is 1

jiaqizho:
Pulled the core read logic out of the parallel lambdas into standalone methods (take_rows_from_files and read_chunks_from_files) so both the serial and parallel paths can share the same code. When parallelism is 1, we now skip the thread pool entirely and just call the method directly, which avoids unnecessary overhead.

Also added GetParallelism() to ThreadPoolHolder so PackedRecordBatchReader can pick up the configured thread pool size automatically instead of hardcoding 1.

Fixed a subtle issue in the futures collection loop: previously if one future returned an error, ARROW_ASSIGN_OR_RAISE would early-return while other tasks were still running with a captured this pointer. Now we wait for all futures to finish first, then check for errors.

### DIFF
--- a/cpp/src/format/column_group_lazy_reader.cpp
+++ b/cpp/src/format/column_group_lazy_reader.cpp
@@ -15,6 +15,7 @@
 #include "milvus-storage/format/column_group_lazy_reader.h"
 
 #include <memory>
+#include <mutex>
 #include <unordered_map>
 #include <vector>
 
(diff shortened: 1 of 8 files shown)


cpp/src/format/column_group_lazy_reader.cpp
```diff
@@ -169,47 +156,74 @@ arrow::Result<std::shared_ptr<arrow::Table>> ColumnGroupLazyReaderImpl::take(con
                                                  needed_columns_, key_retriever_));
     }
   }
+  return arrow::Status::OK();
+}
+
+arrow::Result<std::shared_ptr<arrow::Table>> ColumnGroupLazyReaderImpl::take_rows_from_files(
```
congqixia: in `take_rows_from_files` is it safe to access `loaded_format_readers_` without lock protection?

PR #434, opened 2026-03-04
feat: cache file size and footer size in manifest to skip S3 HEAD requests

jiaqizho:
When we open a file on S3, Arrow always does a HeadObject first just to figure out the file size. That's wasteful since we already know the size at write time. Now both Parquet and Vortex writers capture the file size and footer size when they close, and we store them in ColumnGroupFile in the manifest. On the read side, if the file size is known we pass it via FileInfo so Arrow skips the HEAD entirely. For the footer, if we know its exact size we pre-read the whole thing in one IO instead of Arrow's default two-step approach (read the 8-byte trailer first, then go back for the Thrift metadata).

Also refactored the Vortex read path to open a RandomAccessFile handle once and reuse it for all ReadAt calls, instead of re-opening per request. The reader handle is wrapped in Arc<ReaderHandle> so that if the read future gets cancelled, any in-flight spawn_blocking tasks still hold a valid reference and the underlying C++ object won't get freed from under them.

Updated the FFI layer across C++, Rust and Python to thread file_size and footer_size through. The C struct LoonColumnGroupFile and the Python cffi definition now both include these fields to stay in sync with the internal ColumnGroupFile. The coalescing window max_size was also tuned down to 1 MB to better match our actual read patterns.

### DIFF
--- a/cpp/include/milvus-storage/column_groups.h
+++ b/cpp/include/milvus-storage/column_groups.h
@@ -15,17 +15,42 @@
 #pragma once
 
 #include <string>
+#include <unordered_map>
 #include <vector>
+
+#include "milvus-storage/common/properties_convert.h"
+
 namespace milvus_storage::api {
+
+// Well-known property keys for ColumnGroupFile
+constexpr const char* kPropertyFileSize = "file_size";
+constexpr const char* kPropertyFooterSize = "footer_size";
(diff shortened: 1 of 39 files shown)


xiaofan-luan:
this seems to be a very smart idea. why we close this issue?

PR #437, opened 2026-03-04
fix: wrap Azure output stream with BufferedOutputStream to avoid BlockCountExceedsLimit

xiaocai2333:
Azure Block Blob has a hard limit of 100,000 uncommitted blocks per blob. Arrow's Azure ObjectAppendStream creates a new StageBlock for every Write() call without internal buffering. During large compaction outputs (e.g., 8GB), the Parquet writer produces many small writes (one per row group column page), easily exceeding the 100K block limit.

This fix wraps the output stream with arrow::io::BufferedOutputStream (default 10MB buffer) for non-local filesystems that don't implement UploadSizable (i.e., Azure). This reduces ~100K+ StageBlock calls to ~800 for 8GB data.

S3 and local filesystem paths are unaffected.

### DIFF
--- a/cpp/src/format/parquet/parquet_writer.cpp
+++ b/cpp/src/format/parquet/parquet_writer.cpp
@@ -31,6 +31,8 @@
 #include "milvus-storage/filesystem/fs.h"
 #include "milvus-storage/filesystem/upload_sizable.h"
 
+#include <arrow/io/buffered.h>
+
 namespace milvus_storage::parquet {
 
 static ::parquet::Compression::type convert_compression_type(const std::string& compression) {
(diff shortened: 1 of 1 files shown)


cpp/src/format/parquet/parquet_writer.cpp
```diff
@@ -187,23 +189,36 @@ arrow::Status ParquetFileWriter::init() {
...
     return arrow::Status::IOError(fmt::format("Failed to open output stream: {}", sink_result.status().ToString()));
   }
   sink_ = std::move(sink_result).ValueOrDie();
 
+  // For non-local filesystems without built-in upload buffering (e.g., Azure Blob Storage),
+  // wrap with BufferedOutputStream to reduce the number of remote write calls.
+  // Azure Block Blob has a limit of 100,000 uncommitted blocks per blob; without buffering,
+  // each small Parquet write becomes a separate StageBlock, easily exceeding this limit
+  // for large compaction outputs (e.g., 8GB data can produce >100K blocks).
+  if (needs_buffering && !IsLocalFileSystem(fs_)) {
```
jiaqizho: The core issue here is that azurefs doesn't perform batching; it directly calls `DoAppend` to create blocks. However, for the `s3filesystem`, batching occurs inside `sink_->write`?

Storage plans to port the `azurefs`. Therefore, please add the HOTFIX comment. After we port the `azurefs`, a buffer can be maintained internally, then we can abandoning `BufferedOutputStream` and reducing copying.

PR #446, opened 2026-03-13
[DNM]feat: introduce task-based async read path for take and get_chunks

jiaqizho:
The old parallel read path baked parallelism into the lower-level readers (ColumnGroupLazyReader::take and ColumnGroupReader::get_chunks), which made it hard to control task granularity from the outside. This commit pulls parallelism up to the caller level with a new task-based async design.

The core idea is splitting reads into natural tasks grouped by file, then letting the caller decide how to schedule them. ColumnGroupLazyReader now exposes get_natural_tasks() and take_async(), and ColumnGroupReader exposes get_natural_tasks() and get_chunks_async(). Both return folly::SemiFuture so we can use collectAll for concurrent execution.

FormatReader also gets default async wrappers (read_with_range_async and take_async) that just dispatch the sync methods onto the thread pool. Subclasses with native async I/O can override these later.

On the reader.cpp side, ReaderImpl::take_async and ChunkReaderImpl::get_chunks_async handle the full pipeline: collect natural tasks, split large ones to fill the thread pool, submit via async interface, then reorder results back to match the original input order.

Added tests covering async take with multi-CG reorder, out-of-range error propagation, and fault injection through the async path.

### DIFF
--- a/cpp/include/milvus-storage/format/vortex/vortex_format_reader.h
+++ b/cpp/include/milvus-storage/format/vortex/vortex_format_reader.h
@@ -15,6 +15,7 @@
 #pragma once
 
 #include <arrow/chunked_array.h>
+#include <folly/futures/Future.h>
 
 #include "vortex_bridge.h"  // from cpp/src/format/vortex/vx-bridge/src/include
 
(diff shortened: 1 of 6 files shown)


jiaqizho:
Backup this design. will do verify if i am free

PR #461, opened 2026-03-25
enhance: add version info to manifest commit error messages

jiaqizho:
The error messages in the manifest commit flow were missing version details, which made it hard to figure out what went wrong when a commit failed due to version conflicts.

Now all error paths in the commit process include read_version, latest_version, and committed_version where applicable. Also renamed the misleading "seen_version" param to "latest_version" in resolvers for clarity, and added debug-level logs at key points like version drift detection, successful commits, and conflict retries.


congqixia (commented):
shall we update doc comments in transaction.h
/lgtm

PR #468, opened 2026-03-31
enhance: refactor format dispatch into Format abstract factory

tedxu:
Refactor format dispatch into a Format abstract factory with
Format::get() singleton lookup, PlainFormat base for parquet/vortex,
and dedicated LanceFormat/IcebergFormat implementations. Each format
implements explore(), create_reader(), and create_writer().

Add convertFunc<std::string> specialization to properties_convert.h
for ColumnGroupFile::Get<std::string>() support.


cpp/src/ffi/exttable_c.cpp
```diff
@@ -295,63 +131,24 @@ LoonFFIResult loon_exttable_get_file_info(const char* format,
...
-    // Check file_path is a file
-    auto file_info_res = fs->GetFileInfo(resolved_path);
-    if (!file_info_res.ok()) {
-      RETURN_ERROR(LOON_ARROW_ERROR, file_info_res.status().ToString());
-    }
-    auto file_info = file_info_res.ValueOrDie();
-
-    if (file_info.type() == arrow::fs::FileType::NotFound) {
-      RETURN_ERROR(LOON_INVALID_ARGS, "File not found: ", file_path);
+    auto fmt_res = milvus_storage::Format::get(format_str);
```
jiaqizho, line 134: Before this refactor, `loon_exttable_get_file_info` explicitly rejected any format other than `parquet`/`vortex` with `LOON_INVALID_ARGS "Unsupported format: <fmt>, file_path: <path>"`. Because `Format::get()` now registers `lance-table` and `iceberg-table` unconditionally, callers that pass those format strings here no longer hit that clean rejection:

- `"lance-table"` + a plain file path falls through to `LanceFormat::create_reader` → `lance::ParseLanceUri(file.path)`, which surfaces as a cryptic `LOON_ARROW_ERROR` with `"Invalid uri format: <path>. Expected format: {base_path}?fragment_id=..."`.
- `"iceberg-table"` silently instantiates `IcebergFormatReader::open()` and performs positional-delete I/O — heavyweight side effects for an API whose only job is to count rows in a single file.

If this function is intended to stay parquet/vortex-only, consider re-adding a format whitelist here (or gating it at the factory) so we keep the old `LOON_INVALID_ARGS` contract and fail fast with a readable message. Otherwise, please document in the header that lance/iceberg inputs are now accepted and describe what they mean.

cpp/src/ffi/exttable_c.cpp
```diff
@@ -295,63 +131,24 @@ LoonFFIResult loon_exttable_get_file_info(const char* format,
...
+    if (!fmt_res.ok()) {
+      RETURN_ERROR(LOON_INVALID_ARGS, fmt_res.status().ToString());
     }
 
-    if (file_info.type() != arrow::fs::FileType::File) {
-      RETURN_ERROR(LOON_INVALID_ARGS, "Path is not a file: ", file_path);
+    // Create a ColumnGroupFile to pass to the reader factory
+    ColumnGroupFile cg_file{std::string(file_path), 0, 0, {}};
+    auto reader_res = fmt_res.ValueOrDie()->create_reader(nullptr, cg_file, properties_map, {}, nullptr);
```
jiaqizho, line 167: The old implementation called `fs->GetFileInfo(resolved_path)` before constructing the reader and mapped two cases to `LOON_INVALID_ARGS`:

- `FileType::NotFound` → `"File not found: <path>"`
- any non-`File` type (e.g. a directory) → `"Path is not a file: <path>"`

Those guards are gone, so both inputs now fail deep inside `create_reader()` → `open()` and surface as `LOON_ARROW_ERROR` (code 3) with a generic Arrow I/O message, instead of `LOON_INVALID_ARGS` (code 1). The corresponding assertions in `cpp/test/ffi/ffi_external_test.c` (`ck_assert_int_eq(rc.err_code, LOON_INVALID_ARGS)` for the directory and file-not-found tests) were removed in this PR, which confirms the error-code change is deliberate but silently weakens the public contract — any caller switching on `err_code == LOON_INVALID_ARGS` to distinguish bad input from I/O failures will now take the wrong branch.

Two ways to make the contract explicit again:
1. Restore the `GetFileInfo` pre-flight check so not-found / is-directory keep returning `LOON_INVALID_ARGS`.
2. Document the new mapping in `ffi_exttable_c.h` and update the tests to assert the new code (`LOON_ARROW_ERROR`) instead of dropping the assertions entirely.

cpp/src/ffi/exttable_c.cpp
```diff
@@ -295,63 +131,24 @@ LoonFFIResult loon_exttable_get_file_info(const char* format,
...
-      }
-      *out_num_of_rows = reader.rows();
-    } else {
-      RETURN_ERROR(LOON_INVALID_ARGS, "Unsupported format: " + format_str + ", file_path: " + file_path);
+    auto rg_infos_res = reader_res.ValueOrDie()->get_row_group_infos();
+    if (!rg_infos_res.ok()) {
+      RETURN_ERROR(LOON_ARROW_ERROR, rg_infos_res.status().ToString());
     }
+    auto& rg_infos = rg_infos_res.ValueOrDie();
+    *out_num_of_rows = rg_infos.empty() ? 0 : rg_infos.back().end_offset;
```
jiaqizho, line 177: For parquet, the pre-refactor path was a cheap footer read:

```cpp
auto parquet_reader = parquet::ParquetFileReader::Open(file_reader);
*out_num_of_rows = parquet_reader->metadata()->num_rows();
```

After the refactor, `create_reader(nullptr, cg_file, ..., {}, nullptr)` instantiates a full `ParquetFormatReader` and runs `open()`, which opens the Arrow `FileReader`, calls `GetSchema()`, and iterates all columns to build `needed_column_indices_` even though `needed_columns` is empty. `rg_infos.back().end_offset` is numerically equivalent to `metadata->num_rows()`, but the setup around it is substantially heavier than the old footer-only probe. If `loon_exttable_get_file_info` is called per file during a manifest scan or similar hot path, this is a real per-call overhead regression.

Consider exposing a lightweight `Format::row_count(file, properties)` (parquet implementing via footer metadata, vortex via `VortexFile::RowCount()`) so this function can stay cheap without losing the factory indirection.

cpp/src/ffi/exttable_c.cpp
jiaqizho: BTW, WP still use "File not found" to grep the file-not-exist error. after https://github.com/milvus-io/milvus-storage/pull/490, it will be changed.

cpp/src/ffi/exttable_c.cpp
tedxu: Is there any specific reason not to support lance and iceberg?

cpp/src/ffi/exttable_c.cpp
tedxu: Will fix.

PR #470, opened 2026-04-02
enhance: switch from ARROW_LOG to glog which align with milvus

jiaqizho:
Previously we removed glog to avoid large TLS issues, but now we need it back to align with milvus-common's logging approach. This brings back glog/0.7.1 as a dependency and recreates the storage log.h/log.cpp that were removed in #269. The log macros (LOG_STORAGE_DEBUG_, LOG_STORAGE_WARNING_, etc.) follow the same pattern as milvus core's Log.h, using VLOG(5) for debug and VLOG(6) for trace so log verbosity is controllable at runtime via --v flag.

For AWS SDK logging, replaced the ConsoleLogSystem with a custom GlogAwsLogger that inherits FormattedLogSystem and forwards formatted statements through LOG_STORAGE_INFO_, consistent with how milvus core bridges AWS logs via its AwsLogger in MinioChunkManager.

The reason we align with milvus is that milvus core already handles glog initialization and configuration (log level, output destination, etc.), so storage just needs to use glog directly and the log output will be managed by the host process. With ARROW_LOG, storage's logs were going through a separate logging path that milvus had no control over.

### DIFF
--- a/cpp/include/milvus-storage/common/log.h
+++ b/cpp/include/milvus-storage/common/log.h
@@ -1,0 +1,48 @@
+// Copyright 2023 Zilliz
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include <string>
+#include <typeinfo>
+#include <sys/types.h>
+#include <unistd.h>
+#include "glog/logging.h"
+#include <fmt/format.h>
+
+#if __GLIBC__ == 2 && __GLIBC_MINOR__ < 30
+#include <sys/syscall.h>
+#define gettid() syscall(SYS_gettid)
+#endif
+
+// GLOG has no debug and trace level, using VLOG to implement it.
+#define GLOG_DEBUG 5
+#define GLOG_TRACE 6
+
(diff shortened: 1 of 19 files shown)


cpp/src/filesystem/s3/s3_global.cpp
```diff
@@ -21,19 +21,32 @@
...
 
+class GlogAwsLogger : public Aws::Utils::Logging::FormattedLogSystem {
+  public:
+  explicit GlogAwsLogger(Aws::Utils::Logging::LogLevel log_level)
+      : Aws::Utils::Logging::FormattedLogSystem(log_level) {}
+  ~GlogAwsLogger() override = default;
+  void Flush() override {}
+
+  protected:
+  void ProcessFormattedStatement(Aws::String&& statement) override { LOG_STORAGE_INFO_ << "[Storage-AWS] " << statement; }
```
congqixia: All AWS SDK log messages are forwarded at INFO level regardless of the
  original AWS log level. The FormattedLogSystem base class already filters by
  log_level, but a WARN-level AWS message will still appear as INFO in glog.
  Consider mapping the AWS log level to the corresponding LOG_STORAGE_* macro,
  or at minimum document why INFO is acceptable.

cpp/include/milvus-storage/common/log.h
```diff
@@ -0,0 +1,52 @@
...
+#include <string_view>
+#include <typeinfo>
+#include <sys/types.h>
+#include <unistd.h>
+#include "glog/logging.h"
+
+#if __GLIBC__ == 2 && __GLIBC_MINOR__ < 30
+#include <sys/syscall.h>
+#define gettid() syscall(SYS_gettid)
+#endif
```
congqixia: All AWS SDK log messages are forwarded at INFO level regardless of the
  original AWS log level. The FormattedLogSystem base class already filters by
  log_level, but a WARN-level AWS message will still appear as INFO in glog.
  Consider mapping the AWS log level to the corresponding LOG_STORAGE_* macro,
  or at minimum document why INFO is acceptable.

PR #477, opened 2026-04-08
fix: include SONAME-versioned .so files in JNI fat jar

congqixia:
Package versioned shared libraries (libfoo.so.N) alongside bare .so names so dlopen can resolve SONAMEs embedded in libmilvus-storage-jni (e.g. libarrow_dataset.so.1700), and teach NativeLibraryLoader to recognize versioned suffixes when extracting natives.


jiaqizho (approved):
/lgtm

Unrelated to this PR, but honestly, the `.dll/.so`(or other hard-code) and even the entire `java/package_fat_jar.sh` script are not robust.....

PR #480, opened 2026-04-10
fix: GCP S3 compatibility with AWS SDK 1.11.692 checksum changes

jiaqizho:
After PR https://github.com/milvus-io/milvus-storage/pull/464, AWS SDK 1.11.692 changed its default checksum behavior, it now sends x-amz-checksum-* headers on all requests that support them (WHEN_SUPPORTED). GCP's S3-compatible API doesn't recognize these extra headers and rejects requests with "ExcessHeaderValues" for PutObject and "Missing required header Content-MD5 OR x-amz-checksum-crc32" for DeleteObjects.

Two things going on here. First, for GCP we set the SDK's requestChecksumCalculation to WHEN_REQUIRED so it stops tacking on checksum headers to every request like PutObject. Second, DeleteObjects is a special case, GCP actually requires a checksum for it, but the SDK doesn't consider it mandatory, so we explicitly set CRC32 on the request for GCP.

### DIFF
--- a/cpp/src/filesystem/s3/s3_filesystem.cpp
+++ b/cpp/src/filesystem/s3/s3_filesystem.cpp
@@ -1764,6 +1764,10 @@ class S3FileSystem::Impl : public std::enable_shared_from_this<S3FileSystem::Imp
       req.SetDelete(std::move(del));
       if (options().use_crc32c_checksum) {
         req.SetChecksumAlgorithm(S3Model::ChecksumAlgorithm::CRC32C);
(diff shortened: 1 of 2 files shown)


cpp/src/filesystem/s3/s3_filesystem.cpp
tinswzy, line 1: For the addition of x-amz-checksum-crc32, it might be necessary to confirm whether `s3_filesystem_producer.cpp@MakeRequest()`  needs to be adjusted accordingly

tinswzy (approved):
/lgtm

PR #482, opened 2026-04-13
feat: add Transaction::DropColumn for single-column schema evolution

congqixia:
Add DropColumn method to Transaction that removes a single column from its column group (removing the entire CG if it becomes empty), with automatic index cleanup. The operation is a noop when the column doesn't exist, consistent with DropIndex semantics for idempotency and concurrent transaction safety.

Key design: DropColumn always executes before AddColumnGroup validation in applyUpdates(), regardless of user call order. This enables Drop("x") + AddColumnGroup(containing "x") as an atomic column replacement pattern.

Changes:
- C++ core: Updates/Transaction classes, applyUpdates() ordering
- C FFI / JNI / Python bindings
- 7 test cases covering drop, empty CG removal, noop, index cleanup, and ordering guarantees


cpp/src/jni/manifest_jni.cpp
```diff
@@ -247,6 +247,30 @@ JNIEXPORT jlong JNICALL Java_io_milvus_storage_MilvusStorageTransaction_transact
   }
 }
 
+JNIEXPORT void JNICALL Java_io_milvus_storage_MilvusStorageTransaction_transactionDropColumn(JNIEnv* env,
```
jiaqizho, line 282: return jlong?

if current function failed. Then caller won't know in L264

cpp/src/transaction/transaction.cpp
```diff
@@ -83,6 +88,34 @@ arrow::Result<std::shared_ptr<Manifest>> applyUpdates(const std::shared_ptr<Mani
   auto& indexes = base->indexes();
   auto& lob_files = base->lobFiles();
 
+  // Apply dropped columns (MUST execute before AddColumnGroup validation)
+  // Noop if column doesn't exist — consistent with DropIndex, supports idempotency
+  for (const auto& col_name : updates.GetDroppedColumns()) {
+    for (auto it = cgs.begin(); it != cgs.end();) {
+      if (!*it) {
```
jiaqizho: won't got empty column grouo, if do have empty column group return logical error with arrow::error.

cpp/src/transaction/transaction.cpp
```diff
@@ -83,6 +88,34 @@ arrow::Result<std::shared_ptr<Manifest>> applyUpdates(const std::shared_ptr<Mani
...
+        ++it;
```
jiaqizho: then u don't need ++it in if..else just put it in L94

cpp/src/transaction/transaction.cpp
```diff
@@ -83,6 +88,34 @@ arrow::Result<std::shared_ptr<Manifest>> applyUpdates(const std::shared_ptr<Mani
   auto& indexes = base->indexes();
   auto& lob_files = base->lobFiles();
 
+  // Apply dropped columns (MUST execute before AddColumnGroup validation)
+  // Noop if column doesn't exist — consistent with DropIndex, supports idempotency
+  for (const auto& col_name : updates.GetDroppedColumns()) {
```
jiaqizho, line 93: Apply dropped columns after AddedColumnGroups ? 

Then caller can drop field after add in the same transaction.

cpp/src/jni/manifest_jni.cpp
congqixia: The current void return is intentional. Errors are propagated to Java/Scala callers via ThrowJavaExceptionFromFFIResult, which throws a Java exception. The caller catches failures through the exception mechanism, not a return code. This is consistent with other void JNI methods in the same file (transactionAbort, transactionDestroy). There's no meaningful value to return here — the C++ side DropColumn returns Transaction& for method chaining, not a result status.

cpp/src/transaction/transaction.cpp
congqixia: Can't move ++it into the for-loop increment expression. When cols.empty() is true, we call it = cgs.erase(it), which already advances the iterator to the next element. If ++it were also in the for-loop increment, it would skip an element after every erase. This is the standard erase-while-iterating pattern —  the current form is correct.

cpp/src/transaction/transaction.cpp
congqixia: The current ordering (Drop before Add) is intentional — it enables the column replacement use case: drop an existing column, then add a new column with the same name (e.g., changing column type). The test case DropColumnThenAddSameName validates exactly this scenario.                                             
If we reversed the order (Add before Drop), the Add step would fail with "Column 'x' already exists" when trying to replace a column. The "add then drop in the same transaction" scenario has limited practical value — why add a column just to immediately drop it? Column replacement (drop-then-add-same-name) is the real use case.

jiaqizho: You could decide which position to delete first, and then delete the `cgs` and `indexes` with two loops.

Using `foreach it++ and delete` makes the code difficult to read and maintain.

cpp/src/jni/manifest_jni.cpp
jiaqizho: checkout the logical 

```
// If column_groups is provided, apply updates based on update_id
    if (column_groups) {
      LoonColumnGroups* cgroups = reinterpret_cast<LoonColumnGroups*>(column_groups);

      if (update_id == 1) {
        // ADDFIELD: add each column group as a new field column group
        for (uint32_t i = 0; i < cgroups->num_of_column_groups; i++) {
          LoonFFIResult add_result = loon_transaction_add_column_group(handle, &cgroups->column_group_array[i]);
          if (!loon_ffi_is_success(&add_result)) {
            ThrowJavaExceptionFromFFIResult(env, &add_result);
            loon_ffi_free_result(&add_result);
            return -1;
          }
        }
      } else {
        // ADDFILES (default): append files to existing column groups
        LoonFFIResult append_result = loon_transaction_append_files(handle, cgroups);
        if (!loon_ffi_is_success(&append_result)) {
          ThrowJavaExceptionFromFFIResult(env, &append_result);
          loon_ffi_free_result(&append_result);
          return -1;
        }
      }
    }
```

The logic here seems inconsistent. Why is the update call occurring in `Java_io_milvus_storage_MilvusStorageTransaction_transactionCommit`? It seems the code here isn't being updated.

jiaqizho: And, in commit/begin, jni use the jlong as return value. But `drop column` use the exception? I believe the logic here should be consistent.

congqixia: ▎ Consistency here means unified error channel, not unified return type. All methods throw Java exceptions on error — callers use try-catch uniformly. Return  
  ▎ type reflects the method's semantic output: begin returns a handle, commit returns a version, getColumnGroups returns a pointer — these need jlong. 
  ▎ dropColumn / appendFiles / addColumnGroups / abort / destroy have no value to return, so they're void (same as Abort / Destroy which also throw on error).   
  ▎ Forcing a jlong status code on void-semantic methods would create a second error channel alongside exceptions, which is actually the inconsistent design. 
  ▎ This matches the Python binding where exceptions are the sole error channel.

jiaqizho (approved):
/lgtm

PR #483, opened 2026-04-13
fix: add back conditional write support in filesystem FFI

jiaqizho:
MetricsInputStream, MetricsRandomAccessFile, and MetricsOutputStream were introduced in 9c273ac (PR #390 "feat: add observable support to local filesystem") but only overrode a subset of virtual methods from the Arrow IO interfaces. The rest fell through to base class defaults that silently returned nullptr or NotImplemented.

This adds a `conditional` parameter to `loon_filesystem_open_writer` so callers can request a conditional write that fails if the file already exists. Under the hood it routes to `OpenConditionalOutputStream` when the flag is set, and returns the new `LOON_NOT_SUPPORT` error code if the filesystem doesn't implement it. The Rust bridge, Python bindings, and error code table are all updated to match.

### DIFF
--- a/cpp/ffi_exports_mac.map
+++ b/cpp/ffi_exports_mac.map
@@ -100,7 +100,6 @@ _loon_column_groups_debug_string
 # Manifest interface
 _loon_manifest_destroy
 _loon_manifest_debug_string
-_loon_reset_context
 
 # ThreadPool interface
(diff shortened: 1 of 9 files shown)


tinswzy (approved):
/lgtm

PR #485, opened 2026-04-14
enhance: make FFI tests run on cloud storage

jiaqizho:
Extracts shared test environment helpers into ffi_test_env.c/h so all FFI test suites can detect cloud mode via is_cloud_env() and configure filesystem properties through init_test_props() instead of hardcoding local paths. Each test file is refactored to use these shared helpers.

The Makefile now runs Test_FFI under both minio and azurite environments alongside the existing milvus_test runs, using sub-shells to isolate env vars between providers.

A couple of cloud-specific test adjustments worth noting: list_dir uses a concrete subdirectory instead of "." on S3 because MinIO rejects "." as a path prefix through SubTreeFileSystem, and get_file_stats metadata assertions now branch on is_cloud_env since S3 returns HTTP headers as metadata while Azure currently doesn't.

### DIFF
--- a/cpp/Makefile
+++ b/cpp/Makefile
@@ -73,13 +73,16 @@ test-all: build
 	# running milvus_test and Test_FFI without minio
 	./build/${build_type}/test/milvus_test && ./build/${build_type}/test/Test_FFI
 
+	# use the sub-shell to split the diffent cloud env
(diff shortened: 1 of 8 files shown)


tinswzy (approved):
/lgtm

PR #487, opened 2026-04-15
feat: [jni]add MilvusStorageColumnGroups for direct LoonColumnGroups construction

congqixia:
Expose a JNI entry point that builds a LoonColumnGroups* from caller-provided per-group arrays (columns, files, per-file row counts) without going through MilvusStorageManifest. This lets callers feed loon_reader_new directly when the column-group layout is already known from another source — required by the spark-connector's StorageV2 (non-manifest packed parquet) read path, where the layout comes from the snapshot AVRO + parquet footer kv-metadata and there is no .milvus_manifest file to resolve.

cpp/src/jni/column_groups_jni.cpp:
- Java_..._createFromGroups: validates inputs, allocates LoonColumnGroups + LoonColumnGroup[] + LoonColumnGroupFile[] using new[]/dup_cstr so the existing loon_column_groups_destroy reclaims everything; computes cumulative (start_index, end_index) per file from the row counts (the packed reader rejects negative end_index, so a "whole-file" sentinel is not viable). Wrapped in extern "C" — unlike sibling *_jni.cpp files these symbols have no declaration in ffi_jni.h to inherit C linkage from.
- Java_..._destroy: thin wrapper over loon_column_groups_destroy; safe on a zero pointer.

java/src/main/scala/io/milvus/storage/MilvusStorageColumnGroups.scala:
- Scala wrapper enforcing per-group array-length invariants before crossing the JNI boundary; destroy(0L) is a no-op.

CMakeLists.txt and sbt build files are unchanged: src/jni/*.cpp is already globbed, and the new Scala source lives in the existing source set.


cpp/src/jni/column_groups_jni.cpp
```diff
@@ -0,0 +1,228 @@
...
+                          std::to_string(num_rcs) + ") != files.length (" +
+                          std::to_string(num_files) + ")";
+        env->ThrowNew(exc, msg.c_str());
+        return 0;
+      }
+
+      LoonColumnGroup& out = cgroups->column_group_array[g];
+      out.num_of_columns = static_cast<uint32_t>(num_cols);
+      out.columns = new const char*[num_cols]{};
+      out.format = dup_cstr("parquet");
```
jiaqizho: use the `LOON_FORMAT_PARQUET` replace `parquet`

cpp/src/jni/column_groups_jni.cpp
```diff
@@ -0,0 +1,228 @@
...
+char* dup_cstr(const char* s) {
+  const size_t n = std::strlen(s);
+  char* p = new char[n + 1];
+  std::memcpy(p, s, n);
+  p[n] = '\0';
+  return p;
+}
+
+// Release a partially-constructed LoonColumnGroups on error paths.
+void destroy_partial(LoonColumnGroups* cg) {
```
jiaqizho: direct call `loon_column_groups_destroy`? which already check the `cg`  nullable

cpp/src/jni/column_groups_jni.cpp
```diff
@@ -0,0 +1,228 @@
...
+//   - the outer LoonColumnGroups is `new LoonColumnGroups`.
+
+#include "milvus-storage/ffi_c.h"
+#include "milvus-storage/ffi_jni.h"
+
+#include <cstring>
+#include <string>
+#include <vector>
+
+namespace {
```
jiaqizho: Could you add a test to prevent this section:
if the column group structure changed, but current logical have not been updated?

cpp/src/jni/column_groups_jni.cpp
```diff
@@ -0,0 +1,228 @@
+// Copyright 2026 Zilliz
```
jiaqizho: Can we change this file name?  how about `v2_column_groups_jni.cpp`. 

And should we create a new dir which  handle v2 compatibility issues?

PR #489, opened 2026-04-15
fix: make clang-format CI actually catch format violations

jiaqizho:
Turns out the clang-format CI check has been silently passing even when files were misformatted. The root cause was in cpp/Makefile: the check-format target used $(foreach ...) which expands all the find commands into a single shell line joined by semicolons, so the shell only ever returned the exit code of the last command. Any violations in ./src, ./include, or ./test got printed to stderr but swallowed as long as ./benchmark was clean. Rewrote both check-format and fix-format to use a proper shell loop with set -e so any failure actually fails the target.

On top of that, the CI workflow never pinned a clang-format version, so the ubuntu-22.04 runner was using the default clang-format 14 while most of us run 18 locally. That mismatch meant running make fix-format locally could still leave diffs that CI disagreed with (or vice versa). Added a step to install clang-format-18 from the official LLVM apt repo and point /usr/bin/clang-format at it via update-alternatives, so CI and local are on the same version.

### DIFF
--- a/.github/workflows/clang-format-ci.yml
+++ b/.github/workflows/clang-format-ci.yml
@@ -19,10 +19,19 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v4
-     
+
+      - name: install clang-format-18
+        run: |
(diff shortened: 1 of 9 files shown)


congqixia (approved):
/lgtm

PR #490, opened 2026-04-15
feat: add finer-grained FFI error codes for transactions and file-not-found

jiaqizho:
This commit adds three new FFI error codes so callers can actually tell what went wrong instead of getting a generic arrow error or logical error. Two of them cover the transaction commit path: one signals that we ran out of retries due to concurrent conflicts, the other signals that the resolver decided the conflict couldn't be merged. Both are plumbed through a new ExtendStatusDetail type code so the commit wrapper can unwrap the status and map it back cleanly instead of relying on string matching.

The third new code covers "path not found" on the filesystem side. Previously every call that touched a missing path returned a generic arrow error, or in some places the wrong category entirely, which was confusing and in one case actually buggy — one of the stat-style APIs silently wrote a garbage size into its out parameter when the path didn't exist, because arrow's "get info by path" overload reports missing paths via the result value rather than an error status. The fix handles both conventions: inspect the FileInfo type where arrow uses that convention, and check the errno detail on the status everywhere else. All filesystem entry points that touch an existing path now report missing-path failures with the dedicated code, so callers in Python / Java don't have to grep error messages.

While in there, a few error codes that were miscategorized got straightened out. Some arrow IO failures on the read path were being reported as logical errors, which is supposed to mean our own bugs, not downstream IO; those are now arrow errors. One path was even reporting arrow IO failures as invalid-argument errors, which is doubly wrong.

To make the retry-exhaustion case actually testable, there's a new fault injection point that pauses a writer right before it commits the manifest, which lets a second transaction sneak in and claim the target version first. Two new manifest tests exercise the two new transaction codes — one constructs a real concurrent commit conflict across two threads, the other exercises the fail-resolver path. A new filesystem test hits every affected API with a missing path and asserts the new code comes back.


congqixia (approved):
/lgtm

PR #491, opened 2026-04-15
feat: [jni] add MilvusPackedWriter for caller-path-controlled V2 writes

congqixia:
Expose the C++ PackedRecordBatchWriter over JNI so callers can emit parquet files at exact, caller-supplied paths while still getting milvus-storage's full private KV metadata in the footer. Required by the spark-connector's V2 backfill path: it must produce per-field binlogs at {rootPath}/insert_log/{coll}/{part}/{seg}/{fieldId}/{logId} for Milvus to pick them up via the snapshot AVRO — the existing MilvusStorageWriter JNI (wrapping Loon WriterImpl) generates UUID-suffixed paths under an internal _data/ prefix and is unsuitable for this layout.

Previously the spark side wrote these files via Apache AvroParquetWriter directly. That only set 2 of the 3 KV keys Milvus's strict reader (PackedFileMetadata::Make) requires — `storage_version` and `group_field_id_list` went in via withExtraMetaData, but `row_group_metadata` (milvus-storage-private, computed per row group at close time) was missing, so segment load failed at GroupChunkTranslator.cpp:150 with "missing key row_group_metadata". PackedRecordBatchWriter's flushRemainingBuffer() appends group_field_id_list and then ParquetFileWriter::Close() appends the remaining two keys, so routing through it produces byte-identical output to what milvus's own compaction worker writes.

cpp/include/milvus-storage/packed_writer_c.h (new):
- Four C FFI entry points: loon_packed_writer_new/write/close/destroy, mirroring what milvus's internal/storagev2/packed/packed_writer.go calls on the Go side. Column groups cross the FFI as a CSR-style flat encoding (group_offsets + group_indices) — simpler than mirroring milvus's opaque CColumnSplits type, which lives in milvus core and has never been ported here.

cpp/src/ffi/packed_writer_c.cpp (new):
- Reuses the existing plumbing used by loon_writer_new: ConvertFFIProperties for the properties map, FilesystemCache::getInstance().get() for the Arrow fs (the returned FileSystemProxy is a SubTreeFileSystem rooted at the bucket, so caller paths are bucket-relative keys), and PROPERTY_FS_MULTI_PART_UPLOAD_SIZE for StorageConfig. No new plumbing.
- PackedWriterHolder caches schema alongside the writer because PackedRecordBatchWriter does not expose schema() and ImportRecordBatch needs one on every write call.
- Validates CSR offsets (monotonic, in-bounds) and column indices against the schema before handing off to C++.

cpp/src/jni/packed_writer_jni.cpp (new):
- Java_io_milvus_storage_MilvusPackedWriter_writer{New,Write,Close,Destroy} — thin JNI wrappers around the FFI entry points, using the existing ThrowJavaExceptionFromFFIResult helper.

java/src/main/scala/io/milvus/storage/MilvusPackedWriter.scala (new):
- Scala handle class modelled on MilvusStorageWriter.scala. Flattens Array[Array[Int]] column groups into the CSR offsets/indices pair before the JNI call and validates length invariants.

CMakeLists.txt and sbt build files are unchanged: src/*.cpp and src/jni/*.cpp are already globbed, and the new Scala source lives in the existing source set.

### DIFF
--- a/cpp/include/milvus-storage/ffi_internal/v2_packed_writer_c.h
+++ b/cpp/include/milvus-storage/ffi_internal/v2_packed_writer_c.h
@@ -12,8 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#ifndef MILVUS_STORAGE_PACKED_WRITER_C_H_
-#define MILVUS_STORAGE_PACKED_WRITER_C_H_
+#ifndef MILVUS_STORAGE_FFI_INTERNAL_V2_PACKED_WRITER_C_H_
+#define MILVUS_STORAGE_FFI_INTERNAL_V2_PACKED_WRITER_C_H_
 
 #ifdef __cplusplus
 extern "C" {
@@ -84,4 +84,4 @@ FFI_EXPORT void loon_packed_writer_destroy(LoonPackedWriterHandle handle);
 }
 #endif
 
-#endif  // MILVUS_STORAGE_PACKED_WRITER_C_H_
+#endif  // MILVUS_STORAGE_FFI_INTERNAL_V2_PACKED_WRITER_C_H_
--- a/cpp/src/ffi/v2_packed_writer_c.cpp
+++ b/cpp/src/ffi/v2_packed_writer_c.cpp
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "milvus-storage/packed_writer_c.h"
+#include "milvus-storage/ffi_internal/v2_packed_writer_c.h"
 #include "milvus-storage/ffi_c.h"
 #include "milvus-storage/ffi_internal/result.h"
 #include "milvus-storage/properties.h"
@@ -86,6 +86,10 @@ LoonFFIResult loon_packed_writer_new(const char* const* paths,
     std::vector<std::vector<int>> column_groups(num_groups);
     if (group_offsets[0] != 0) {
       RETURN_ERROR(LOON_INVALID_ARGS, "group_offsets[0] must be 0, got ", group_offsets[0]);
+    }
+    if (group_offsets[num_groups] != total_indices) {
+      RETURN_ERROR(LOON_INVALID_ARGS, "group_offsets[num_groups] must equal total_indices, got ",
+                   group_offsets[num_groups], " vs ", total_indices);
     }
     for (int32_t i = 0; i < num_groups; ++i) {
       int32_t lo = group_offsets[i];
@@ -163,10 +167,16 @@ LoonFFIResult loon_packed_writer_write(LoonPackedWriterHandle handle, ArrowArray
     }
     auto status = holder->writer->Write(rb_result.ValueOrDie());
     if (!status.ok()) {
+      if (array->release) {
+        array->release(array);
+      }
       RETURN_ERROR(LOON_ARROW_ERROR, status.ToString());
     }
     RETURN_SUCCESS();
   } catch (std::exception& e) {
+    if (array && array->release) {
+      array->release(array);
+    }
     RETURN_EXCEPTION(e.what());
   }
   RETURN_UNREACHABLE();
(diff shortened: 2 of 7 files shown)


cpp/include/milvus-storage/ffi_internal/v2_packed_writer_c.h
```diff
@@ -0,0 +1,87 @@
+// Copyright 2025 Zilliz
```
jiaqizho, line 1: v2_packed_writer_c.h ? and check others?

java/src/main/scala/io/milvus/storage/MilvusPackedWriter.scala
```diff
@@ -0,0 +1,121 @@
...
+   * @param bufferSize    Max in-memory buffer in bytes (0 = native default).
+   */
+  def create(
+      paths: Array[String],
+      columnGroups: Array[Array[Int]],
+      schemaPtr: Long,
+      properties: MilvusStorageProperties,
+      bufferSize: Long = 0L
+  ): Unit = {
+    if (isDestroyed) throw new IllegalStateException("Writer has been destroyed")
```
jiaqizho, line 41: also check `if (writerHandle != 0)` ?

cpp/include/milvus-storage/ffi_internal/v2_packed_writer_c.h
jiaqizho, line 1: also move this file to cpp/include/milvus-storage/ffi_internal

cpp/src/ffi/v2_packed_writer_c.cpp
```diff
@@ -0,0 +1,197 @@
...
+    auto* holder = reinterpret_cast<PackedWriterHolder*>(handle);
+    auto rb_result = arrow::ImportRecordBatch(array, holder->schema);
+    if (!rb_result.ok()) {
+      if (array->release) {
+        array->release(array);
+      }
+      RETURN_ERROR(LOON_ARROW_ERROR, rb_result.status().ToString());
+    }
+    auto status = holder->writer->Write(rb_result.ValueOrDie());
+    if (!status.ok()) {
```
jiaqizho, line 169: missing 

```
if (array->release) {
array->release(array);
} 

```
if failed

cpp/src/ffi/v2_packed_writer_c.cpp
```diff
@@ -0,0 +1,197 @@
...
+      RETURN_ERROR(LOON_ARROW_ERROR, status.ToString());
+    }
+    RETURN_SUCCESS();
+  } catch (std::exception& e) {
```
jiaqizho, line 176: missing 

```
if (array && array->release) {
array->release(array);
} 
```

cpp/src/jni/v2_packed_writer_jni.cpp
```diff
@@ -0,0 +1,156 @@
...
+      env->ThrowNew(exc_class, "paths/group_offsets/group_indices must not be null");
+      return -1;
+    }
+
+    num_paths = env->GetArrayLength(jpaths);
+    owned_paths.reserve(static_cast<size_t>(num_paths));
+    path_ptrs.reserve(static_cast<size_t>(num_paths));
+    for (jsize i = 0; i < num_paths; ++i) {
+      auto jstr = static_cast<jstring>(env->GetObjectArrayElement(jpaths, i));
+      const char* utf = env->GetStringUTFChars(jstr, nullptr);
```
jiaqizho, line 59: `utf` maybe nullptr

cpp/src/ffi/v2_packed_writer_c.cpp
```diff
@@ -0,0 +1,197 @@
...
+                 "Invalid arguments: paths/group_offsets/group_indices/schema/properties/out_handle "
+                 "must not be null");
+  }
+  if (num_groups <= 0) {
+    RETURN_ERROR(LOON_INVALID_ARGS, "num_groups must be > 0, got ", num_groups);
+  }
+  if (total_indices < 0) {
+    RETURN_ERROR(LOON_INVALID_ARGS, "total_indices must be >= 0, got ", total_indices);
+  }
+
```
jiaqizho, line 71: Add the check with `group_offsets` and `group_indices`. Prevent exceeding the boundaries during use.

PR #492, opened 2026-04-17
fix: unbreak macOS build after Conan 2.x migration

jiaqizho:
A few things were broken on macOS after the Conan 2.x migration.

First, conanfile.py was setting self.options["arrow"].with_jemalloc inside requirements(), which Conan 2.x flatly rejects with "Dependencies options were defined incorrectly". Moved it into configure() where dep option overrides actually belong. While I was there, the old del self.options["folly"].use_sse4_2 also blows up with KeyError if the folly recipe no longer exposes that option, so swapped it for rm_safe which just silently no-ops in that case.

Second, the openssl package() step was dying on arm64 macOS when Conan tried to rewrite rpaths via install_name_tool, with the classic "larger updated load commands do not fit" error. The openssl binary just wasn't linked with enough Mach-O header padding. Exporting LDFLAGS=-headerpad_max_install_names in the Makefile on Darwin gets inherited by the openssl subprocess build so its binary actually has room for install_name_tool to add rpaths later.

### DIFF
--- a/cpp/Makefile
+++ b/cpp/Makefile
@@ -21,6 +21,9 @@ libcxx_setting = -s compiler.cppstd=17
 uname_s := $(shell uname -s)
 ifeq ($(uname_s),Darwin)
 	libcxx_setting = -s compiler.libcxx=libc++ -s compiler.cppstd=17
+	# Ensure enough Mach-O header padding so Conan's fix_apple_shared_install_name
+	# (install_name_tool) can rewrite rpaths in dependency binaries like openssl.
+	export LDFLAGS := $(LDFLAGS) -headerpad_max_install_names
 endif
 
 # Common conan flags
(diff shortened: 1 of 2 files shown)


cpp/Makefile
```diff
@@ -21,6 +21,9 @@ libcxx_setting = -s compiler.cppstd=17
 uname_s := $(shell uname -s)
 ifeq ($(uname_s),Darwin)
 	libcxx_setting = -s compiler.libcxx=libc++ -s compiler.cppstd=17
+	# Ensure enough Mach-O header padding so Conan's fix_apple_shared_install_name
```
liliu-z, line 24: `:=` overwrites any externally provided LDFLAGS, which can break ASAN builds and other configurations passing flags via LDFLAGS. Should append instead: `export LDFLAGS := $(LDFLAGS) -headerpad_max_install_names`.

PR #493, opened 2026-04-18
feat: [ffi] per-batch RecordBatchReader with offset-0 export for JVM consumers

congqixia:
The Spark connector consumes V2 packed-parquet segments through the existing ArrowArrayStream-based `loon_get_record_batch_reader` path. That path does not work correctly with Arrow Java's `ArrowReader`:

1. `ArrowReader.loadNextBatch` imports every batch into the SAME open `VectorSchemaRoot` and ignores `ArrowArray.offset`. When `PackedRecordBatchReader::ReadNext` returns a sliced batch (the remainder it keeps in its per-chunk queue via `rb->Slice(min_rows)`), every post-slice batch re-reads from buffer position 0, producing a row-cycle of period `min_rows` — rows 0..min_rows-1's values repeat forever while `current_offset_` advances.

2. Milvus segcore's Go/cgo binding (`segcore/packed_reader_c.cpp ReadNext`) sidesteps this by exporting a fresh ArrowArray+ArrowSchema pair per batch. This PR gives the FFI layer the same shape so JVM callers can mirror that proven scheme.

Changes:

* New per-batch RecordBatchReader FFI:
- `loon_record_batch_reader_new(reader, predicate, out_handle)`
- `loon_record_batch_reader_read_next(handle, out_array, out_schema)`
- `loon_record_batch_reader_destroy(handle)` Owns a `shared_ptr<arrow::RecordBatchReader>` on the native side; each `read_next` call exports one batch via `arrow::ExportRecordBatch` into caller-allocated C Data structs. EOF is signalled by leaving `out_array->release == nullptr`. Symbols added to both `ffi_exports.map` and `ffi_exports_mac.map`.

* Materialize sliced columns to offset=0 inside `read_next`. Before exporting, scan each column's `offset()` and, for any column with a non-zero offset, call `arrow::Concatenate({col})` to copy the slice into fresh contiguous buffers. Non-sliced columns pass through unchanged. This preserves correctness for any JVM-style consumer that cannot honour `ArrowArray.offset` — the offset field is always zero across this boundary.

* JNI bindings (`recordBatchReaderNew`, `recordBatchReaderReadNext`, `recordBatchReaderDestroy`) plus matching `MilvusStorageReader` `destroyRecordBatchReaderScala`).

* `reader_jni.cpp` correctness:
- Wrap the entire JNI function body in `extern "C" { ... }`. Some functions were only declared in `ffi_jni.h` (which already has an `extern "C"` block) while newer additions were not, and the compiler was silently giving them C++ linkage — the resulting mangled symbols made the JVM throw `UnsatisfiedLinkError` at runtime (`Java_io_milvus_storage_MilvusStorageReader_<method>` could not be resolved).
- Sync stale declarations in `ffi_jni.h`: `getChunkReader` and `take` gained a `jobjectArray needed_columns` parameter in the implementation but the header was never updated, so callers got two differently-linked declarations for the same symbol.
- Add declarations for the three new `recordBatchReader*` JNI entry points inside the existing `extern "C"` block.

Verification:

* `nm -D libmilvus-storage-jni.so` now shows every `Java_io_milvus_storage_MilvusStorageReader_*` symbol as a plain C export; no `_Z...` mangled variants remain.
* End-to-end backfill against a V2 snapshot (Spark connector → `openRecordBatchReaderScala` →
`readNextBatchScala` per batch) reads all 20480 rows of a 5-group segment with monotonically-correct primary keys:
`row_offset=8192 → ID=8192`, `row_offset=10000 → ID=10000`. Previous ArrowArrayStream-based path exhibited
`ID = row_offset mod 8192`.


jiaqizho:
The V2 and V3 both have this problem. Reproduced the problem with follow test(gen by AI):

OffsetCycleReproTest.scala
```
package io.milvus.storage

import org.apache.arrow.c.{ArrowArray, ArrowArrayStream, ArrowSchema, Data}
import org.apache.arrow.memory.RootAllocator
import org.apache.arrow.vector.BigIntVector
import org.apache.arrow.vector.VectorSchemaRoot
import org.apache.arrow.vector.types.pojo.{ArrowType, Field, FieldType, Schema}
import org.scalatest.BeforeAndAfterAll
import org.scalatest.flatspec.AnyFlatSpec
import org.scalatest.matchers.should.Matchers

import java.io.File
import java.util.{HashMap => JHashMap}
import scala.collection.JavaConverters._

/**
 * Reproduces the ArrowArray.offset cycle bug that PR #493 fixes on the
 * producer side. Mirrors the C++ test at
 * cpp/test/ffi/per_batch_offset_test.cpp: one V2 packed parquet file with
 * exactly 10 row groups, each carrying 131072 int64 rows. The packed
 * reader's default record-batch limit is 8192, so it slices every row
 * group 16 times via rb->Slice(min_rows); the ArrowArray handed back to
 * Java across the FFI boundary carries a non-zero offset for 15 of every
 * 16 batches.
 *
 * Two consumer shapes are exercised:
 *   1. Per-batch ArrowArray imported via Data.importVectorSchemaRoot —
 *      fresh root each time. This is what MilvusStorageIntegrationTest
 *      already uses, and what the spark-milvus
 *      MilvusLoonV2PartitionReader does today.
 *   2. Stream-based ArrowReader via Data.importArrayStream with a
 *      SHARED VectorSchemaRoot across loadNextBatch() calls. This is the
 *      shape described in the PR #493 message as the offending path.
 *
 * If Arrow Java's C-data importer honours ArrowArray.offset, both
 * consumer shapes see monotonic IDs 0..N-1. If it does not, we expect
 * `id[row_offset] == row_offset mod 8192` — duplicated rows.
 */
class OffsetCycleReproTest extends AnyFlatSpec with Matchers with BeforeAndAfterAll {

  private val TEST_BASE_PATH = "/tmp/milvus_storage_offset_cycle_test"
  private val FILE_PATH = s"$TEST_BASE_PATH/v2.parquet"

  // Match the C++ reproducer exactly.
  private val NUM_BATCHES = 10
  // 131072 = 1 MiB / 8 B per int64 → exactly one row group per batch.
  private val ROWS_PER_BATCH = 131072
  private val TOTAL_ROWS = NUM_BATCHES.toLong * ROWS_PER_BATCH

  private val allocator = new RootAllocator(Long.MaxValue)
  private var libraryLoaded = false

  override def beforeAll(): Unit = {
    try {
      NativeLibraryLoader.loadLibrary()
      libraryLoaded = true
      val dir = new File(TEST_BASE_PATH)
      if (dir.exists()) deleteRecursively(dir)
      dir.mkdirs()
      writeV2File()
    } catch {
      case e: Throwable =>
        libraryLoaded = false
        println(s"beforeAll failed: ${e.getMessage}")
        e.printStackTrace()
    }
  }

  override def afterAll(): Unit = {
    val dir = new File(TEST_BASE_PATH)
    if (dir.exists()) deleteRecursively(dir)
    allocator.close()
  }

  private def deleteRecursively(f: File): Unit = {
    if (f.isDirectory) f.listFiles().foreach(deleteRecursively)
    f.delete()
  }

  /** Single-column `id` schema with the PARQUET:field_id=100 metadata
   * that milvus-storage writers require. */
  private def makeSchema(): Schema = {
    val meta = Map("PARQUET:field_id" -> "100").asJava
    val fields = List(
      new Field("id", new FieldType(false, new ArrowType.Int(64, true), null, meta), null)
    ).asJava
    new Schema(fields)
  }

  /** Write N batches × ROWS_PER_BATCH monotonic int64 values into a
   * VectorSchemaRoot, export as a single ArrowArray + ArrowSchema, and
   * hand them to MilvusPackedWriter. Close triggers the 10 row-group
   * flush inside ParquetFileWriter::Flush. */
  private def writeV2File(): Unit = {
    val schema = makeSchema()
    val schemaPtrObj = ArrowSchema.allocateNew(allocator)
    Data.exportSchema(allocator, schema, null, schemaPtrObj)

    val props = new MilvusStorageProperties()
    val m = new JHashMap[String, String]()
    m.put("fs.storage_type", "local")
    m.put("fs.root_path", "/")
    props.create(m)

    val writer = new MilvusPackedWriter()
    writer.create(
      paths = Array(FILE_PATH),
      columnGroups = Array(Array(0)),
      schemaPtr = schemaPtrObj.memoryAddress(),
      properties = props,
      bufferSize = 64L * 1024 * 1024
    )

    // MilvusPackedWriter buffers batches and reads them lazily (flush
    // happens inside close()), so the exported ArrowArray/Schema wrappers
    // MUST stay alive until writer.close() returns — closing them earlier
    // triggers their release callback while C++ still holds the buffer
    // pointer, which ends up serving all-zero data on the reader side.
    val keepAlive = new scala.collection.mutable.ArrayBuffer[(VectorSchemaRoot, ArrowArray, ArrowSchema)]()
    try {
      for (b <- 0 until NUM_BATCHES) {
        val root = VectorSchemaRoot.create(schema, allocator)
        val vec = root.getVector("id").asInstanceOf[BigIntVector]
        vec.setInitialCapacity(ROWS_PER_BATCH)
        vec.allocateNew()
        val start = b.toLong * ROWS_PER_BATCH
        var i = 0
        while (i < ROWS_PER_BATCH) {
          vec.set(i, start + i)
          i += 1
        }
        vec.setValueCount(ROWS_PER_BATCH)
        root.setRowCount(ROWS_PER_BATCH)

        val arr = ArrowArray.allocateNew(allocator)
        val sch = ArrowSchema.allocateNew(allocator)
        Data.exportVectorSchemaRoot(allocator, root, null, arr, sch)
        writer.write(arr.memoryAddress())
        keepAlive.append((root, arr, sch))
      }
      writer.close()
    } finally {
      writer.destroy()
      // Now safe to release Java-side buffers.
      keepAlive.foreach { case (root, arr, sch) =>
        arr.close()
        sch.close()
        root.close()
      }
      props.free()
      schemaPtrObj.close()
    }
  }

  /** Build a fresh LoonColumnGroups pointer for the single-file layout. */
  private def openReader(): (MilvusStorageReader, Long, Long, MilvusStorageProperties) = {
    val cgs = MilvusStorageColumnGroups.createFromGroups(
      columnsPerGroup = Array(Array("id")),
      filesPerGroup = Array(Array(FILE_PATH)),
      fileRowCountsPerGroup = Array(Array(TOTAL_ROWS))
    )

    val schemaPtrObj = ArrowSchema.allocateNew(allocator)
    Data.exportSchema(allocator, makeSchema(), null, schemaPtrObj)

    val props = new MilvusStorageProperties()
    val m = new JHashMap[String, String]()
    m.put("fs.storage_type", "local")
    m.put("fs.root_path", "/")
    props.create(m)

    val reader = new MilvusStorageReader()
    reader.create(cgs, schemaPtrObj.memoryAddress(), Array("id"), props)
    (reader, cgs, schemaPtrObj.memoryAddress(), props)
  }

  // ============================================================
  // Variant 1: per-batch ArrowArray + Data.importVectorSchemaRoot
  // ============================================================
  "V2 packed file" should "read back with per-batch fresh VectorSchemaRoot" in {
    assume(libraryLoaded, "Native library must be loaded")

    val (reader, cgs, readerSchemaPtr, props) = openReader()
    val streamPtr = reader.getRecordBatchReaderScala(null)

    val readSchema = makeSchema()
    var totalRows = 0L
    var expected = 0L
    var firstMismatchExpected = -1L
    var firstMismatchActual = -1L
    var monotonic = true
    var numBatches = 0
    // Snapshot the first few batches' row counts + head values so we can
    // reason about what the Java importer is actually seeing.
    val batchSizes = new scala.collection.mutable.ArrayBuffer[Int]()
    val firstRowsPerBatch = new scala.collection.mutable.ArrayBuffer[Seq[Long]]()

    try {
      var keepGoing = true
      while (keepGoing) {
        val arrayPtr = ArrowUtils.readNextBatch(streamPtr)
        if (arrayPtr == 0L) {
          keepGoing = false
        } else {
          numBatches += 1
          val arrow = ArrowArray.wrap(arrayPtr)
          val sch = ArrowSchema.allocateNew(allocator)
          try {
            Data.exportSchema(allocator, readSchema, null, sch)
            val root = Data.importVectorSchemaRoot(allocator, arrow, sch, null)
            try {
              val rowCount = root.getRowCount
              val vec = root.getVector(0).asInstanceOf[BigIntVector]
              if (batchSizes.size < 5) {
                batchSizes.append(rowCount)
                val headK = math.min(rowCount, 4)
                firstRowsPerBatch.append((0 until headK).map(j => vec.get(j)))
              }
              var i = 0
              while (i < rowCount) {
                val v = vec.get(i)
                if (v != expected && monotonic) {
                  firstMismatchExpected = expected
                  firstMismatchActual = v
                  monotonic = false
                }
                expected += 1
                i += 1
              }
              totalRows += rowCount
            } finally {
              root.close()
            }
          } finally {
            sch.close()
          }
        }
      }
    } finally {
      ArrowUtils.releaseArrowStream(streamPtr, true)
      reader.destroy()
      MilvusStorageColumnGroups.destroy(cgs)
      ArrowUtils.releaseArrowSchema(readerSchemaPtr, false)
      val w = ArrowSchema.wrap(readerSchemaPtr); w.close()
      props.free()
    }

    println(
      s"[per-batch] numBatches=$numBatches totalRows=$totalRows monotonic=$monotonic " +
        s"firstMismatch expected=$firstMismatchExpected actual=$firstMismatchActual"
    )
    batchSizes.zip(firstRowsPerBatch).zipWithIndex.foreach { case ((sz, head), idx) =>
      println(s"  batch[$idx] size=$sz head=${head.mkString(",")}")
    }

    totalRows should be(TOTAL_ROWS)
    monotonic should be(true)
  }

  // ============================================================
  // Variant 2: ArrowArrayStream + ArrowReader with shared root
  // ============================================================
  it should "read back via ArrowReader with a shared VectorSchemaRoot" in {
    assume(libraryLoaded, "Native library must be loaded")

    val (reader, cgs, readerSchemaPtr, props) = openReader()
    val streamPtr = reader.getRecordBatchReaderScala(null)

    var totalRows = 0L
    var expected = 0L
    var firstMismatchExpected = -1L
    var firstMismatchActual = -1L
    var monotonic = true
    var numBatches = 0

    val stream = ArrowArrayStream.wrap(streamPtr)
    try {
      val arrowReader = Data.importArrayStream(allocator, stream)
      try {
        while (arrowReader.loadNextBatch()) {
          numBatches += 1
          val root = arrowReader.getVectorSchemaRoot
          val rowCount = root.getRowCount
          val vec = root.getVector(0).asInstanceOf[BigIntVector]
          var i = 0
          while (i < rowCount) {
            val v = vec.get(i)
            if (v != expected && monotonic) {
              firstMismatchExpected = expected
              firstMismatchActual = v
              monotonic = false
            }
            expected += 1
            i += 1
          }
          totalRows += rowCount
        }
      } finally {
        arrowReader.close()
      }
    } finally {
      reader.destroy()
      MilvusStorageColumnGroups.destroy(cgs)
      ArrowUtils.releaseArrowSchema(readerSchemaPtr, false)
      val w = ArrowSchema.wrap(readerSchemaPtr); w.close()
      props.free()
    }

    println(
      s"[shared-root] numBatches=$numBatches totalRows=$totalRows monotonic=$monotonic " +
        s"firstMismatch expected=$firstMismatchExpected actual=$firstMismatchActual"
    )

    // This is the PR #493 bug path. If Arrow Java's VectorLoader ignores
    // ArrowArray.offset (the PR's diagnosis), `monotonic` is false and
    // the first mismatch is at row 8192 with actual value 0 — i.e. the
    // ID cycle repeats every 8192 rows (the reader's default batch limit).
    totalRows should be(TOTAL_ROWS)
    monotonic should be(true)
  }
}
```

jiaqizho:
Can u just move this logical into `java/src/main/scala/io/milvus/storage/ArrowUtils.scala` ? 

This is a problem specific to Java and does not affect other languages. I do not intend to provide a separate FFI interface that includes Concatenate (memory copy). Also if changed in java, V3 no need change the interface.

jiaqizho:
> Can u just move this logical into `java/src/main/scala/io/milvus/storage/ArrowUtils.scala` ?
> 
> This is a problem specific to Java and does not affect other languages. I do not intend to provide a separate FFI interface that includes Concatenate (memory copy). Also if changed in java, V3 no need change the interface.

or move to the `ffi_jni.h`

java/src/main/scala/io/milvus/storage/MilvusStorageReader.scala
jiaqizho: remove this function. java should never use this function.

cpp/src/jni/reader_jni.cpp
```diff
@@ -14,13 +14,160 @@
...
+            auto concat_result = arrow::Concatenate({col}, arrow::default_memory_pool());
+            if (!concat_result.ok()) {
+              RETURN_ERROR(LOON_ARROW_ERROR, concat_result.status().ToString());
+            }
+            fresh_cols.push_back(concat_result.ValueOrDie());
+          }
+        }
+        batch = arrow::RecordBatch::Make(batch->schema(), batch->num_rows(), fresh_cols);
+      }
+    }
```
jiaqizho:
```
else {  // batch == nullptr
      out_array->release = nullptr;
      out_schema->release = nullptr;
      RETURN_SUCCESS();
}
```

jiaqizho (approved):
/lgtm

PR #494, opened 2026-04-18
feat: [spark-backfill][ffi] per-batch RecordBatchReader with offset-0 export for JVM consumer

congqixia:
The Spark connector consumes V2 packed-parquet segments through the existing ArrowArrayStream-based `loon_get_record_batch_reader` path. That path does not work correctly with Arrow Java's `ArrowReader`:

1. `ArrowReader.loadNextBatch` imports every batch into the SAME open `VectorSchemaRoot` and ignores `ArrowArray.offset`. When `PackedRecordBatchReader::ReadNext` returns a sliced batch (the remainder it keeps in its per-chunk queue via `rb->Slice(min_rows)`), every post-slice batch re-reads from buffer position 0, producing a row-cycle of period `min_rows` — rows 0..min_rows-1's values repeat forever while `current_offset_` advances.

2. Milvus segcore's Go/cgo binding (`segcore/packed_reader_c.cpp ReadNext`) sidesteps this by exporting a fresh ArrowArray+ArrowSchema pair per batch. This PR gives the FFI layer the same shape so JVM callers can mirror that proven scheme.

Changes:

* New per-batch RecordBatchReader FFI:
- `loon_record_batch_reader_new(reader, predicate, out_handle)`
- `loon_record_batch_reader_read_next(handle, out_array, out_schema)`
- `loon_record_batch_reader_destroy(handle)` Owns a `shared_ptr<arrow::RecordBatchReader>` on the native side; each `read_next` call exports one batch via `arrow::ExportRecordBatch` into caller-allocated C Data structs. EOF is signalled by leaving `out_array->release == nullptr`. Symbols added to both `ffi_exports.map` and `ffi_exports_mac.map`.

* Materialize sliced columns to offset=0 inside `read_next`. Before exporting, scan each column's `offset()` and, for any column with a non-zero offset, call `arrow::Concatenate({col})` to copy the slice into fresh contiguous buffers. Non-sliced columns pass through unchanged. This preserves correctness for any JVM-style consumer that cannot honour `ArrowArray.offset` — the offset field is always zero across this boundary.

* JNI bindings (`recordBatchReaderNew`, `recordBatchReaderReadNext`, `recordBatchReaderDestroy`) plus matching `MilvusStorageReader` `destroyRecordBatchReaderScala`).

* `reader_jni.cpp` correctness:
- Wrap the entire JNI function body in `extern "C" { ... }`. Some functions were only declared in `ffi_jni.h` (which already has an `extern "C"` block) while newer additions were not, and the compiler was silently giving them C++ linkage — the resulting mangled symbols made the JVM throw `UnsatisfiedLinkError` at runtime (`Java_io_milvus_storage_MilvusStorageReader_<method>` could not be resolved).
- Sync stale declarations in `ffi_jni.h`: `getChunkReader` and `take` gained a `jobjectArray needed_columns` parameter in the implementation but the header was never updated, so callers got two differently-linked declarations for the same symbol.
- Add declarations for the three new `recordBatchReader*` JNI entry points inside the existing `extern "C"` block.

Verification:

* `nm -D libmilvus-storage-jni.so` now shows every `Java_io_milvus_storage_MilvusStorageReader_*` symbol as a plain C export; no `_Z...` mangled variants remain.
* End-to-end backfill against a V2 snapshot (Spark connector → `openRecordBatchReaderScala` →
`readNextBatchScala` per batch) reads all 20480 rows of a 5-group segment with monotonically-correct primary keys:
`row_offset=8192 → ID=8192`, `row_offset=10000 → ID=10000`. Previous ArrowArrayStream-based path exhibited
`ID = row_offset mod 8192`.


jiaqizho:
Pls mark [spark-backfill] in title..

Although merify bot isn't fully completed yet, labeling the titles to differentiate between merging master and other branches will help with filtering.

PR #496, opened 2026-04-21
feat: add GCP cross-tenant access via service account impersonation

jiaqizho:
This adds cross-tenant GCS support by letting us impersonate a service account that lives in another GCP project. It's basically the GCP equivalent of the AWS STS AssumeRole flow we already shipped — Milvus runs in tenant A, the bucket sits in tenant B, and instead of baking static keys into the config we chain through an impersonation target.

On the C++ side there's a new GcpCredentialProvider plus a registry, so the filesystem producer can look up and reuse impersonated creds instead of the producer itself being the only thing that knows how to build GCP credentials. This also cleans up the producer a bit now that the AKSK path and the impersonation path go through the same abstraction.

The Rust bridge gets a new gcp_impersonation module that wires the same behavior into the Iceberg and Lance readers, so external tables pointing at a cross-tenant GCS bucket just work end-to-end without callers having to plumb anything extra.

Tests cover both the filesystem path and the external table path, including the ARN-style config we standardized on. The full design writeup with the alternatives we considered lives in docs/gcp-cross-tenant-impersonation-design.md.


weiliu1031:
## Code Review — 2 issues to address before merge

Verified by cloning the PR branch and reading lance-io / object_store sources.

### 1. `load_frequency` validation inconsistency between C++ and Rust

Same config value produces different behavior on the two read paths:

- Global validator (`cpp/src/properties.cpp:394`) accepts `[900, 43200]`
- C++ IAM impersonation (`cpp/src/filesystem/gcp/gcp_credential_provider.cpp:101-103`) silently clamps to `3600`
- Rust Lance path (`cpp/src/format/bridge/rust/src/lance_bridgeimpl.rs:464-471`) rejects `>3600` with `invalid_input`

A user configuring `load_frequency=7200` passes global validation, then:
- The C++ GCP filesystem producer clamps to 3600 and continues
- The Rust `open_dataset` / `write_dataset` path returns an error

Same properties blob, divergent outcomes, hard to debug.

**Suggested fix**: reject oversize values in `BuildGcpProviderFromConfig` and drop the `std::min` clamp in `MakeImpersonationPublicCreds`. Aligns with `AssumeRoleConfig::parse`'s reject style.

```cpp
arrow::Result<std::shared_ptr<GcpCredentialProvider>> BuildGcpProviderFromConfig(
    const ArrowFileSystemConfig& config) {
  if (config.use_iam && !config.gcp_target_service_account.empty()) {
    if (config.load_frequency > 3600) {
      return arrow::Status::Invalid(
          "gcp_target_service_account requires load_frequency in [900, 3600], got ",
          config.load_frequency);
    }
    return std::make_shared<IamImpersonateProvider>(
        config.gcp_target_service_account, config.load_frequency);
  }
  ...
}
```

---

### 2. `reqwest::Client` has no timeout configured

`cpp/src/format/bridge/rust/src/gcp_impersonation.rs:204` and `:252` both construct the HTTP client with `reqwest::Client::new()`. Default builder has neither `timeout` nor `connect_timeout` set.

Failure modes this exposes:
1. GCE metadata server (`169.254.169.254`) stalls during VM migration, node pressure, or network partition → `get_credential` blocks indefinitely
2. `iamcredentials.googleapis.com` returns 5xx but keeps the TCP connection alive → same outcome
3. `object_store` calls `get_credential` on every request → **all concurrent requests on the dataset stall**, with no log output and no surfaced error

`tokio::sync::RwLock` queues `read().await` behind a pending writer to prevent writer starvation. A stuck refresher holding the write lock therefore blocks every reader in the fast path, not just the one issuing the IAM call.

**Suggested fix** (both call sites):
```rust
let http_client = reqwest::Client::builder()
    .timeout(Duration::from_secs(30))
    .connect_timeout(Duration::from_secs(10))
    .build()
    .expect("reqwest client build");
```

Normal metadata requests complete in ~10ms, IAM calls in ~500ms. A 30s cap gives ample margin for jitter while still surfacing stalls through `object_store::Error::Generic` → lance query error, instead of a silent hang.

PR #498, opened 2026-04-22
fix: PackedRecordBatchReader returns wrong column data when projection reorders columns

jiaqizho:
Whenever the projection order for a column group differs from the write order, PackedRecordBatchReader was silently returning data from the wrong column. The output schema had the right field names, but the arrays under them were shuffled around. Nothing crashed, nothing logged — just garbage values in the wrong slots.

The fix flips the outer iteration so we walk needed_columns_ and keep names that live in the current cg. That produces the exact same idx definition ColumnGroupReader uses, so columnMap and out_field_map_ line up with the rb's actual layout. No other paths needed touching — ChunkReader was already correct because it returns the rb verbatim, and take() uses a separate name-keyed merge that never hit this path.

Added tests that would have caught this originally, plus coverage for adjacent projection and schema-evolution scenarios we had holes in. ColumnOrderReorderedSchemaAndProjection now verifies values instead of only schema names, which is exactly why this bug slipped through for so long. New tesWhenever the projection order for a column group differs from the write order, PackedRecordBatchReader was silently returning data from the wrong column. The output schema had the right field names, but the arrays under them were shuffled around. Nothing crashed, nothing logged — just garbage values in the wrong slots.

The bug sat in PackedRecordBatchReader::open(). columns_after_projection was built by iterating cg->columns (disk/write order), so the (cg_index, idx_in_columns) pairs stashed in columnMap were disk-order indices. But the RecordBatch coming back from ColumnGroupReader has its columns in needed_columns_ (projection) order — that's what column_group_reader.cpp does when it filters. ReadNext then used a disk-order index to address a projects cover numeric-looking field names, multi-CG reorder with data verification, interleaved null-fill columns, ChunkReader intersection semantics, per-call needed_columns override on get_chunk_reader and take, and a negative case for projecting a name absent from the read schema. Also dropped a diagnostic test on the ParquetFileWriter side to pin down how it handles record batches whose column order disagrees with the writer schema.

### DIFF
--- a/cpp/src/reader.cpp
+++ b/cpp/src/reader.cpp
@@ -117,25 +117,27 @@ class PackedRecordBatchReader final : public arrow::RecordBatchReader {
     }
     assert(already_set_total_rows);
 
-    // build the columns after projection in column groups
+    // build the columns after projection in column groups.
+    // Must match the column order that ColumnGroupReader produces: it filters
+    // needed_columns_ against each CG's columns (see column_group_reader.cpp),
+    // so the resulting RecordBatch columns follow needed_columns_ order, not
+    // cg->columns (write) order. Iterate needed_columns_ here to keep the
+    // (cg_index, idx_in_columns) pairs in sync with rb->column(idx).
     // ex.
-    //   needed columns: [A, B, C, F]
+    //   needed columns: [F, A, B, C]
     //   column group 0: [A, C, E]
     //   column group 1: [B, D, F]
     //   columns after projection:
     //   column group 0: [A, C]
-    //   column group 1: [B, F]
+    //   column group 1: [F, B]
     //
     std::vector<std::vector<std::string_view>> columns_after_projection(column_groups_.size());
-    {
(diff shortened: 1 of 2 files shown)


jiaqizho:
### Code review

Focused review of projection + schema-evolution across all paths (PackedRecordBatchReader, ChunkReader, take, per-call overrides, null-fill, numeric-looking names, cross-CG reorder). The rewritten loop in `PackedRecordBatchReader::open()` now iterates `needed_columns_` and filters by `cg->columns`, which matches the order `ColumnGroupReader::create` uses when it builds `filtered_columns` — so `columnMap`'s `idx_in_columns` and `rb->column(idx)` stay in sync. The duplicate-name guard in `resolve_needed_columns` is placed before the early-return and covers every entry point (get_record_batch_reader / get_chunk_reader / take, including per-call overrides).

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

congqixia (approved):
/lgtm

PR #500, opened 2026-04-23
fix: OSS/COS/OBS compatibility with AWS SDK 1.11.692 checksum changes

jiaqizho:
Follow-up to #480. Turns out the same aws-sdk-cpp 1.11.692 checksum mess that broke GCP also broke Aliyun OSS, Tencent COS and Huawei OBS, the error strings just looked different enough that it wasn't obvious until we actually ran against each of them (Aliyun complained about aws-chunked, Huawei threw XAmzContentSHA256Mismatch, Tencent griped about Content-MD5).

For PutObject the fix is identical to what we did for GCP: flip requestChecksumCalculation to WHEN_REQUIRED so the SDK stops tacking on x-amz-checksum-* headers and aws-chunked streaming that none of these providers understand. That single switch clears up the PutObject failures across all three.

DeleteObjects is where things get weird. GCP accepts x-amz-checksum-crc32 so #480 just sets the ChecksumAlgorithm and moves on. Aliyun, Tencent and Huawei silently ignore the whole x-amz-checksum-* family and only honor Content-MD5, which the new SDK no longer computes automatically, and MD5 isn't even an option in the ChecksumAlgorithm enum anymore. So we compute the MD5 ourselves off req.SerializePayload() and inject it as a custom header. That's safe because AmazonSerializableWebServiceRequest::GetBody wraps SerializePayload directly into the body stream, so the bytes we hash are exactly what the SDK sends on the wire.

### DIFF
--- a/cpp/src/filesystem/s3/s3_client.cpp
+++ b/cpp/src/filesystem/s3/s3_client.cpp
@@ -553,10 +553,12 @@ arrow::Result<std::shared_ptr<S3ClientHolder>> ClientBuilder::BuildClient(
 
   const bool use_virtual_addressing = options_.endpoint_override.empty() || options_.force_virtual_addressing;
 
-  // GCP's S3-compatible API does not accept the extra x-amz-checksum-* headers
-  // that AWS SDK >= 1.11.x sends by default (WHEN_SUPPORTED). Restrict to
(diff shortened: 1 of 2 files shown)


jiaqizho:
### Code review

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #504, opened 2026-04-26
feat: add property to force non-view schema in vortex schemaless reads

jiaqizho:
Vortex maps DType::Utf8 / DType::Binary to Utf8View / BinaryView when nothing tells it otherwise, which is its preferred Arrow output but not what milvus consumes downstream.  Upstream's stance (vortex-data/vortex#4522) is that this is by design and callers should pass a schema. So instead of waiting on them, this adds a milvus-storage-side knob: PROPERTY_READER_VORTEX_SCHEMA_NON_VIEW, default false. When it's true AND the caller doesn't supply a read schema, the bridge synthesizes a plain Utf8 / Binary arrow schema (mirroring vortex's own to_arrow_dtype but swapping the two view branches) and feeds that to into_record_batch_reader. Vortex honors it and emits 3-buffer Utf8 / plain Binary. GetFileSchema gets the same treatment so it agrees with what reads actually produce.

If a caller passes their own read schema, the property is ignored — explicit schema always wins. Default-off means existing callers see zero behavior change.

### DIFF
--- a/cpp/include/milvus-storage/format/vortex/vortex_format_reader.h
+++ b/cpp/include/milvus-storage/format/vortex/vortex_format_reader.h
@@ -86,6 +86,10 @@ class VortexFormatReader final : public FormatReader, public std::enable_shared_
   std::shared_ptr<arrow::Schema> file_schema_;  // always derived from file in open()
 
   uint64_t logical_chunk_rows_;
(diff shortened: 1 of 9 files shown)


jiaqizho:
milvus already fixed, closed.

PR #507, opened 2026-04-30
[DNM]feat: add Azure cross-tenant support for external tables

jiaqizho:
This adds the C++ side for Azure cross-tenant access. The new fs.azure_client_id and fs.azure_tenant_id properties are parsed into the filesystem config, and AzureFileSystem can now use our Managed Identity to exchange for a customer-tenant storage bearer without holding the customer's account key.

For Iceberg, ABFS/ABFSS reads can now go through a custom ADLS storage factory when cross-tenant options are present. The bridge handles the Managed Identity to AAD token exchange, injects the bearer into ADLS requests, and keeps the normal Azure path unchanged when cross-tenant mode is not enabled.

For Lance, az:// reads now get a per-call custom object store session with a bearer-token credential provider. This lets Lance read customer Azure storage through the same cross-tenant flow while filter ing out account keys, SAS tokens, and other credentials that could bypass the intended path.

### DIFF
--- a/cpp/include/milvus-storage/filesystem/azure/azure_cross_tenant_credential.h
+++ b/cpp/include/milvus-storage/filesystem/azure/azure_cross_tenant_credential.h
@@ -1,0 +1,96 @@
+// Copyright 2025 Zilliz
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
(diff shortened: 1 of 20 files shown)


jiaqizho:
current azure iceberg support need iceberg-rust 0.9(rust 1.92)

jiaqizho:
An alternative authentication path was selected; closed

PR #508, opened 2026-05-06
enhance: tighten FFI validation and writer schema errors

jiaqizho:
This tightens FFI-side argument validation so bad inputs are caught consistently instead of slipping into asserts, null dereferences, or silent defaults.

Writer FFI paths now import record batches with the writer's declared schema, so schema mismatches between the writer and incoming record batch are returned as structured FFI errors instead of surfacing later in the write path.

This also makes delta log num_entries use int64_t across the C and Python FFI surface, rejects non-positive counts, and returns proper status errors from the bridge layer when imported manifest data is malformed.


cpp/src/ffi/bridge.cpp
```diff
@@ -180,6 +212,12 @@ arrow::Status manifest_export(const std::shared_ptr<milvus_storage::api::Manifes
                               LoonManifest** out_cmanifest) {
   assert(manifest != nullptr && out_cmanifest != nullptr);
```
liliu-z: The `manifest_export` function still uses `assert(manifest)` which is stripped in Release builds, leaving a null dereference. The `manifest` pointer is used before the `try` block, so any null access crashes without returning an error. Replace the assert with a proper `Invalid` status return and initialize `*out_cmanifest = nullptr` before the early validation loop.

cpp/src/ffi/bridge.cpp
```diff
@@ -94,29 +94,59 @@ static void export_column_group(const ColumnGroup* cg, LoonColumnGroup* ccg) {
...
-static void import_column_group_file(const LoonColumnGroupFile* in_ccgf, ColumnGroupFile* cgf) {
-  assert(in_ccgf != nullptr && cgf != nullptr);
+static arrow::Status import_column_group_file(const LoonColumnGroupFile* in_ccgf, ColumnGroupFile* cgf) {
+  if (!in_ccgf || !cgf) {
+    return arrow::Status::Invalid("column group file and output must not be null");
+  }
+  if (!in_ccgf->path) {
+    return arrow::Status::Invalid("column group file path is null");
+  }
+  if (in_ccgf->start_index < 0 || in_ccgf->end_index < 0 || in_ccgf->start_index > in_ccgf->end_index) {
```
liliu-z: The new negative row-bound validation in `bridge.cpp:104` and `column_groups_c.cpp:50` rejects `-1/-1` values, but external-table discovery (`format.cpp:91`) still emits `-1/-1` as a sentinel. Any FFI consumer that round-trips external-table manifests will hit a validation error. Either allow `-1/-1` through as a valid sentinel or stop emitting it upstream.

cpp/src/ffi/column_groups_c.cpp
```diff
@@ -30,25 +30,26 @@ LoonFFIResult loon_column_groups_create(const char** columns,
...
       cg->columns.emplace_back(columns[col_idx]);
     }
 
     cg->files.reserve(file_lens);
     for (size_t file_idx = 0; file_idx < file_lens; file_idx++) {
-      if (!paths[file_idx]) {
-        RETURN_ERROR(LOON_INVALID_ARGS, "Path is null [index=" + std::to_string(file_idx) + "]");
-      }
+      RETURN_ERROR_IF(!paths[file_idx], LOON_INVALID_ARGS, "Path is null [index=" + std::to_string(file_idx) + "]");
+      RETURN_ERROR_IF(
```
liliu-z: Same row-range sentinel incompatibility as `bridge.cpp:104`. This validation site also rejects the `-1/-1` sentinel emitted by external-table discovery in `format.cpp:91`.

cpp/src/ffi/segment_writer_c.cpp
```diff
@@ -64,45 +64,35 @@ LoonFFIResult loon_segment_writer_new(ArrowSchema* schema_raw,
                                       const LoonSegmentWriterConfig* config,
                                       const LoonProperties* properties,
                                       LoonSegmentWriterHandle* out_handle) {
-  if (!schema_raw || !config || !properties || !out_handle) {
-    RETURN_ERROR(LOON_INVALID_ARGS,
-                 "Invalid arguments: schema_raw, config, properties, and out_handle must not be null");
-  }
+  RETURN_ERROR_IF(!schema_raw || !config || !properties || !out_handle, LOON_INVALID_ARGS,
+                  "Invalid arguments: schema_raw, config, properties, and out_handle must not be null");
+  RETURN_ERROR_IF(config->num_lob_columns > 0 && !config->lob_columns, LOON_INVALID_ARGS,
```
liliu-z, line 80: In `loon_segment_writer_close`, `malloc` at line 166 and `strdup` at line 168 are not checked for failure. If either returns null, the function continues writing into invalid memory or produces a partially populated `out_output`. The output struct should be zero-initialized at entry, allocations checked, and values committed to `out_output` only after all allocations succeed. Only publish `num_lob_files` after the whole LOB copy succeeds.

cpp/src/ffi/segment_writer_c.cpp
```diff
@@ -64,45 +64,35 @@ LoonFFIResult loon_segment_writer_new(ArrowSchema* schema_raw,
...
+                  "Invalid arguments: config.lob_columns must not be null when num_lob_columns > 0");
```
liliu-z, line 81: The `strdup` call at line 168 is not checked for null. If it fails, the output struct will contain a null pointer where a valid string is expected, leading to undefined behavior in consumers.

cpp/include/milvus-storage/ffi_c.h
```diff
@@ -207,7 +207,7 @@ typedef struct LoonColumnGroups {
  */
 typedef struct LoonDeltaLogs {
   const char** delta_log_paths;
-  uint32_t* delta_log_num_entries;
+  int64_t* delta_log_num_entries;
```
liliu-z, line 210: `loon_segment_write_output_free` now frees `column_groups`, but the header doesn't document this ownership. Callers who separately destroy `column_groups` before calling `_free` will double-free. Add an explicit docstring stating that `LoonSegmentWriteOutput` owns `column_groups` and callers must not separately destroy it.

cpp/src/ffi/reader_c.cpp
```diff
@@ -374,42 +353,52 @@ static inline std::shared_ptr<std::vector<std::string>> convert_needed_columns(c
   return std::make_shared<std::vector<std::string>>(result);
 }
 
+static inline std::optional<std::string> validate_needed_columns(const char* const* strings, size_t count) {
```
liliu-z, line 349: The `if (strings[i])` branch inside `convert_needed_columns` is now dead code because `validate_needed_columns` already rejects null entries before this point is reached. This is also a behavior change: previously null projection entries were silently skipped, now they return `LOON_INVALID_ARGS`. Delete the dead branch or document the intentional behavior tightening.

cpp/src/ffi/properties_c.cpp
```diff
@@ -101,22 +101,17 @@ LoonFFIResult loon_properties_create(const char* const* keys,
...
-      RETURN_ERROR(LOON_INVALID_ARGS, "Invalid keys/values");
-    }
+    RETURN_ERROR_IF(count == 0 || !keys || !values, LOON_INVALID_ARGS, "Invalid keys/values");
 
     properties->properties = static_cast<LoonProperty*>(malloc(sizeof(LoonProperty) * count));
-    if (!properties->properties) {
-      RETURN_ERROR(LOON_MEMORY_ERROR, "Failed to malloc [size=", sizeof(LoonProperty) * count, "]");
-    }
+    RETURN_ERROR_IF(!properties->properties, LOON_MEMORY_ERROR, "Failed to malloc [size=", sizeof(LoonProperty) * count,
+                    "]");
```
liliu-z, line 114: `ConvertFFIProperties` tightening changes behavior: null property arrays or entries now return `LOON_INVALID_PROPERTIES` instead of the previous silent no-op/assert behavior. This is correct but is a breaking change that should be documented in release notes.

cpp/src/ffi/segment_writer_c.cpp
```diff
@@ -164,24 +146,19 @@ LoonFFIResult loon_segment_writer_flush(LoonSegmentWriterHandle handle) {
...
     auto result = writer->Close();
-    if (!result.ok()) {
-      RETURN_ERROR(LOON_ARROW_ERROR, result.status().ToString());
-    }
+    RETURN_ERROR_IF(!result.ok(), LOON_ARROW_ERROR, result.status().ToString());
 
     auto output = std::move(result).ValueOrDie();
 
     auto cgs = output.column_groups;
     auto export_st = milvus_storage::column_groups_export(*cgs, &out_output->column_groups);
```
liliu-z: `column_groups_export` writes via `**out_column_groups`, so the "populate into temporaries" approach cannot trivially wrap it. The correct fix is: do all `malloc`/`strdup` into local stack variables first, keep `column_groups_export` at the end, then commit all values to `out_output` only on full success.

python/milvus_storage/transaction.py
```diff
@@ -187,34 +187,45 @@ def add_column_group(self, column_group: ColumnGroup) -> None:
...
-            path_buffers = []
+            buffers.append(c_files)
             for i, f in enumerate(column_group.files):
                 path_c = self._ffi.new("char[]", f.path.encode("utf-8"))
-                path_buffers.append(path_c)
+                buffers.append(path_c)
                 c_files[i].path = path_c
                 c_files[i].start_index = f.start_index
                 c_files[i].end_index = f.end_index
                 if f.metadata:
```
liliu-z: The Python metadata API uses `metadata: bytes` which can produce lossy or UTF-8-crashing behavior when round-tripping through the C FFI. Multiple call sites are affected: `transaction.py:215`, `manifest.py:203`, `manifest.py:271`, `manifest.py:456`. Either migrate to `properties: Dict[str, str]` or add safe encoding with a clear restriction.

python/tests/test_transaction.py
```diff
@@ -136,6 +136,33 @@ def test_transaction_append_files(temp_dir, sample_schema, properties):
...
+    with Writer(temp_dir, sample_schema, properties=properties) as writer:
+        writer.write(data)
+        column_groups = writer.close()
+
+    try:
+        column_group = column_groups.to_list()[0]
+        with Transaction(temp_dir, properties=properties) as txn:
+            txn.add_column_group(column_group)
+            version = txn.commit()
+
```
liliu-z, line 173: The new `test_transaction_add_column_group` test writes a column group and asserts `commit()` returns a version, but never reads the manifest back to verify metadata survived. The test would pass even if the C side silently dropped properties. Combined with `_from_c` ignoring `lob_files`, no Python test verifies any end-to-end manifest round-trip.

liliu-z:
**cpp/include/milvus-storage/common/ffi_c.h:0** The `num_entries` field in `LoonDeltaLogs` changed from `uint32_t*` to `int64_t*`. Any downstream C or Python consumer compiled against the old header will silently misinterpret the data, causing corruption. Requires lockstep release with Python bindings (and any other C consumer). Java/JNI is unaffected since it does not currently map DeltaLog structs.

liliu-z:
**cpp/src/common/reader_c.cpp:0** The PR aims to tighten FFI validation by replacing `assert` with proper error codes, but `reader_c.cpp` still contains raw `assert()` calls. These will crash in release builds (or be compiled out entirely with NDEBUG), violating the PR's own contract. The same issue exists in `writer_c.cpp`.

liliu-z:
**cpp/src/common/reader_c.cpp:0** `open_reader` now calls `GetSize()` eagerly to produce an early `NotFound` error. On S3/Azure/GCS this is a synchronous HEAD request on every reader open. The result is immediately discarded (not cached for later ReadAt). This is a real performance regression for cloud deployments.

liliu-z:
**cpp/src/common/bridge.cpp:474** `manifest_import` constructs a `Manifest` with an empty `std::vector<Index>{}`, silently dropping any index data. An `export → import` round-trip will lose indexes without warning. This is consistent with the FFI surface not exposing indexes, but callers expecting a faithful clone will get silently corrupted data.

jiaqizho:
Responses to the new review comments:

**`LoonDeltaLogs.delta_log_num_entries` ABI change**

Good catch. This is an ABI change. `delta_log_num_entries` was changed to `int64_t*` to match the C++ `Manifest::DeltaLog::num_entries` type and the persisted Avro `long` schema. The Python binding in this PR is updated accordingly, but C/Python consumers need to be released/rebuilt in lockstep against the new header. Java/JNI is unaffected since it does not map this struct.

**`open_reader` eager `GetSize()`**

This has been addressed in the lower filesystem layer after the rebase: input-file open/size handling is lazy now, so cloud filesystems do not issue an extra synchronous HEAD on every reader open. No separate FFI-level change is needed here.

**`manifest_import` index round-trip**

Correct. The current C FFI `LoonManifest` surface does not expose index metadata yet, so index round-trip is not supported through this API. Supporting indexes needs an explicit FFI schema/API extension in a separate change; this PR keeps the existing FFI manifest scope.

jiaqizho:
> **cpp/src/common/reader_c.cpp:0** The PR aims to tighten FFI validation by replacing `assert` with proper error codes, but `reader_c.cpp` still contains raw `assert()` calls. These will crash in release builds (or be compiled out entirely with NDEBUG), violating the PR's own contract. The same issue exists in `writer_c.cpp`.

This was actually expected, as assert and input checks do not belong to the same category.

PR #511, opened 2026-05-08
feat: add predicate pushdown for Vortex format

tedxu:
See #512

Implement predicate pushdown for Vortex files using sqlparser-rs
to parse SQL WHERE-clause predicates into native Vortex expressions.
The predicate string from get_record_batch_reader(predicate) is
parsed once in Rust, converted to a vortex::expr::Expression, and
applied via ScanBuilder::WithFilter() for zone-map pruning and
row-level filtering. No expression tree crosses the FFI boundary.

Changes:
- Rust: sqlparser-based predicate parser (comparisons, AND/OR/NOT,
  IN lists, quoted column names for all-digit fields)
- CXX bridge: expr::parse_predicate() for one-time parse,
  ScanBuilder integration
- C++: predicate plumbing through Reader -> PackedRecordBatchReader
  -> ColumnGroupReader -> FormatReader::set_predicate()
- VortexFormatReader: parse once in set_predicate(), apply by-ref
  in read() via WithFilter()
- Single CG only; multi-CG logs warning and ignores predicate
- Parquet/Lance/Iceberg: no-op (default set_predicate)
- Handle empty filtered results from Vortex scan
- loon read: add --predicate and --verbose flags
- Benchmark: sorted vs unsorted, int64 vs string predicates,
  filesystem I/O metrics

Benchmark highlights (40K rows, Vortex, local SSD):
- Sorted + 10% selectivity: 75% I/O reduction, 2.5x faster
- String predicates within 9-14% of int64 on sorted data
- Unsorted data: predicate adds overhead with no pruning benefit

### DIFF
--- a/cpp/benchmark/benchmark_predicate.cpp
+++ b/cpp/benchmark/benchmark_predicate.cpp
@@ -1,0 +1,267 @@
+// Copyright 2024 Zilliz
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "benchmark_format_common.h"
+
+#include <iomanip>
+#include <sstream>
+#include <thread>
+#include <arrow/table.h>
+#include "milvus-storage/thread_pool.h"
+#include "milvus-storage/filesystem/observable.h"
+
+namespace milvus_storage::benchmark {
+
+using namespace milvus_storage::api;
+
+//=============================================================================
+// Predicate Pushdown Benchmark
+//
+// Args: [sorted, selectivity_pct, predicate_col]
+//   sorted: 0 = unsorted (random), 1 = sorted (sequential)
+//   selectivity_pct: 0 = no predicate, 10 = keep 10%, 50 = keep 50%, 90 = keep 90%
+//   predicate_col: 0 = int64 (id), 1 = string (name)
+//=============================================================================
+
+/// Zero-pad an integer to a fixed width string
+static std::string ZeroPad(int64_t val, int width) {
+  std::ostringstream oss;
+  oss << std::setw(width) << std::setfill('0') << val;
+  return oss.str();
+}
+
+class PredicateBenchmark : public FormatBenchFixtureBase<> {
+  public:
+  void SetUp(::benchmark::State& st) override {
+    FormatBenchFixtureBase<>::SetUp(st);
+    schema_ = GetLoaderSchema();
+  }
+
+  void TearDown(::benchmark::State& st) override {
+    schema_.reset();
+    ThreadPoolHolder::Release();
+    FormatBenchFixtureBase<>::TearDown(st);
+  }
+
+  protected:
+  /// Build a RecordBatch with zero-padded names.
+  /// Both sorted and unsorted use the SAME values (id=0..N-1, name="name_00000"..),
+  /// so predicates have identical selectivity. The only difference is row order:
+  /// sorted = sequential, unsorted = shuffled. This isolates the effect of
+  /// data ordering (zone-map pruning) from predicate selectivity.
+  arrow::Result<std::shared_ptr<arrow::RecordBatch>> CreatePaddedBatch(const std::shared_ptr<arrow::Schema>& schema,
+                                                                       const DataSizeConfig& config,
+                                                                       bool sorted) {
+    int64_t num_rows = static_cast<int64_t>(config.num_rows);
+    int pad_width = static_cast<int>(std::to_string(num_rows - 1).size());
+
+    // Generate row indices: 0..N-1, then optionally shuffle
+    std::vector<int64_t> indices(num_rows);
+    std::iota(indices.begin(), indices.end(), 0);
+    if (!sorted) {
+      std::mt19937 rng(42);  // fixed seed for reproducibility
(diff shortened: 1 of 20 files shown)


cpp/src/format/bridge/rust/src/vortex_bridge.cpp
```diff
@@ -98,6 +98,14 @@ Expr select(const std::vector<std::string_view>& fields, Expr child) {
   return Expr(ffi::select(rs_fields, std::move(child).IntoImpl()));
 }
 
+Expr parse_predicate(const std::string& predicate) {
+  try {
+    return Expr(ffi::parse_predicate_string(rust::Str(predicate.data(), predicate.length())));
+  } catch (const rust::cxxbridge1::Error& e) {
```
jiaqizho: It will be a exception if current `predicate` not support.

In fact, storage does not provide an "executor" function. It should be allowed to fail, and failures should not be propagated via exceptions.

cpp/src/reader.cpp
jiaqizho, line 745: also add `const std::string& /*predicate*/` ?

cpp/src/reader.cpp
```diff
@@ -89,11 +92,19 @@ class PackedRecordBatchReader final : public arrow::RecordBatchReader {
         milvus_storage::api::GetValueNoError<int64_t>(properties_, PROPERTY_READER_RECORD_BATCH_MAX_ROWS);
     parallelism_ = milvus_storage::ThreadPoolHolder::GetParallelism();
 
+    // Predicate pushdown only supported for single column group.
+    // Multi-CG filtering requires row-index tracking for cross-CG alignment.
+    if (!predicate_.empty() && column_groups_.size() > 1) {
```
jiaqizho, line 97: We can support Multi-CG filtering if lower level reader return the mask. Still can align the rows

cpp/src/format/vortex/vortex_format_reader.cpp
```diff
@@ -215,6 +227,10 @@ arrow::Result<ArrowArrayStream> VortexFormatReader::read(uint64_t row_start, uin
     scan_builder.WithOutputSchema(c_arrow_schema);
   }
 
+  if (parsed_predicate_) {
+    scan_builder.WithFilter(*parsed_predicate_);
```
jiaqizho, line 272: I think the main issue with the current PR is that it performs full filtering rather than sparse filtering. The performance of full filtering is much worse than sparse filtering. My understanding is that, for “filter pushdown,” the storage layer only needs to provide sparse filtering capability at this stage.

cpp/benchmark/benchmark_predicate.cpp
```diff
@@ -0,0 +1,267 @@
...
+  arrow::Status PrepareTestData(const std::string& format,
+                                bool sorted,
+                                std::shared_ptr<ColumnGroups>& out_cgs,
+                                std::string& out_path,
+                                int64_t& out_num_rows) {
+    out_path = GetUniquePath(format + "_predicate_test");
+
+    ARROW_ASSIGN_OR_RAISE(auto policy, CreateSinglePolicy(format, schema_));
+
+    auto writer = Writer::create(out_path, schema_, std::move(policy), properties_);
```
jiaqizho, line 117: default writer won't record any STATISTICS by default.

Use `PROPERTY_WRITER_VORTEX_ENABLE_STATISTICS` to open it?

cpp/src/format/vortex/vortex_format_reader.cpp
jiaqizho: Performing two complete filter, especially with a low filter rate, will reduce the end-to-end performance.

cpp/src/format/bridge/rust/src/predicate_parser.rs
```diff
@@ -0,0 +1,318 @@
...
+/// - `"age > 30"`
+/// - `"status IN (1, 2, 3)"`
+/// - `"(a > 1 OR b > 2) AND c = 3"`
+pub fn parse_predicate(predicate: &str) -> Result<Expression> {
+    let predicate = predicate.trim();
+    if predicate.is_empty() {
+        bail!("empty predicate");
+    }
+
+    let sql = format!("SELECT * FROM t WHERE {predicate}");
```
jiaqizho: try ?
```
let dialect = GenericDialect {};
  let mut parser = Parser::new(&dialect)
      .try_with_sql(predicate)
      .map_err(|e| anyhow::anyhow!("failed to tokenize predicate: {e}"))?;
```
I don't think that is a good practices to hard code `SELECT * FROM t WHERE`?

cpp/src/reader.cpp
tedxu: The chunk reader, per our previous design, does not support predicate. Filtering on chunks may break the caching preconditions.

cpp/src/format/bridge/rust/src/predicate_parser.rs
```diff
@@ -0,0 +1,318 @@
...
+                .selection
+                .ok_or_else(|| anyhow::anyhow!("no WHERE clause found")),
+            _ => bail!("unexpected query body"),
+        },
+        _ => bail!("unexpected statement type"),
+    }
+}
+
+/// Convert a sqlparser AST expression into a Vortex expression.
+fn convert_expr(sql_expr: &SqlExpr) -> Result<Expression> {
```
jiaqizho: If we do not validate whether the columns referenced by the predicate actually exist, the current scan will always fail once Vortex tries to bind/evaluate the expression. In this case, we should not treat it as a hard scan failure. The correct behavior should be to fall back to an unfiltered scan.

cpp/src/format/vortex/vortex_format_reader.cpp
tedxu: Vortex's with_filter() couples zone-map pruning with row-level evaluation in the patched-in version; there is no prune-only knob today. Refreshed benchmark numbers (stats now default-on after this PR) still show net wins on selective sorted data, e.g. 75% I/O reduction and 2.2× speedup at 10% selectivity vs the unsorted baseline. If/when Vortex exposes a prune-only path, we'll switch. Tracking as a follow-up in #524.

PR #514, opened 2026-05-08
enhance: expose Arrow reader properties

sparknack:


jiaqizho:
#516  do the similar changes, closed.

PR #518, opened 2026-05-11
fix: use-after-free in segment take streams

jiaqizho:
Stack:

```
AddressSanitizer:DEADLYSIGNAL
=================================================================
==12195==ERROR: AddressSanitizer: SEGV on unknown address 0x00000001c8e8 (pc 0x000123246764 bp 0x00016b2c4180 sp 0x00016b2c40b0 T0)
==12195==The signal is caused by a READ memory access.
    #0 0x000123246764 in arrow::TableBatchReader::ReadNext(std::__1::shared_ptr<arrow::RecordBatch>*)+0xf0 (libmilvus-storage.dylib:arm64+0x9d06764)
    #1 0x0001232a3398 in arrow::(anonymous namespace)::ExportedArrayStream<arrow::RecordBatchReader, false>::StaticGetNext(ArrowArrayStream*, ArrowArray*)+0x40 (libmilvus-storage.dylib:arm64+0x9d63398)
    #2 0x000104b718e0 in assert_stream_first_batch ffi_segment_test.c:108
    #3 0x000104b6edd4 in run_segment_suite ffi_segment_test.c:363
    #4 0x000104b72f14 in main ffi_test_main.c:41
    #5 0x00019be3dd50  (<unknown module>)
```

The logical in segment_reader_c.cpp:L181

```
auto batch_reader = std::make_shared<arrow::TableBatchReader>(*table);
```

The segment reader FFI take path built TableBatchReader from *table, which only kept a reference to the local Table object. Once loon_segment_reader_take returned, the exported Arrow stream could read from freed memory and crash in TableBatchReader::ReadNext. Build the reader from the shared_ptr instead so the table stays alive for the stream lifetime.

### DIFF
--- a/cpp/src/ffi/segment_reader_c.cpp
+++ b/cpp/src/ffi/segment_reader_c.cpp
@@ -178,7 +178,7 @@ LoonFFIResult loon_segment_reader_take(LoonSegmentReaderHandle handle,
     auto table = std::move(result).ValueOrDie();
 
     // Convert Table to RecordBatchReader, then export as ArrowArrayStream
-    auto batch_reader = std::make_shared<arrow::TableBatchReader>(*table);
+    auto batch_reader = std::make_shared<arrow::TableBatchReader>(std::move(table));
     auto status = arrow::ExportRecordBatchReader(batch_reader, out_stream);
     if (!status.ok()) {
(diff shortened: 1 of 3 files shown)


jiaqizho:
@zhagnlu PTAL

PR #519, opened 2026-05-11
Remove --update to reduce conan check

yhmo:
Conan will no longer proactively check remotes for newer recipe/package revisions when a matching dependency already exists in the local cache.

- If a dependency is truly missing from cache, Conan will still download it from
  remotes.
- This change reduces random failures caused by remote checks or newly published revisions.


jiaqizho:
Older logical add `--build=arrow` to flush the arrow build cache(which not build with azure, and will got undefined symbol). For now, we change to use the conan 2.x, it's fine to remove this logical.

PR #525, opened 2026-05-12
feat: Add a row group zone map layout for Vortex V2

jiaqizho:
The default `ZoneMapLayout` keeps stats under the data layout. That works for a simple zoned layout, but it is not a great fit for our V2 row group layout. It can leave holes between columns, and each column's stats may end up scattered around the physical data layout instead of being grouped together as one pruning-friendly structure.

```
ZoneMapLayout childs:

  - 0: data
  - 1: stats

  which produce interleaved physical layout:

  - RG0 data
  - RG0 zones
  - RG1 data
  - RG1 zones
    ...
```

This adds a dedicated `RowGroupZoneMapLayout` for Vortex V2. The main row group data stays in the existing V2 chunked layout, while all row group zonemap data is written into a compact sidecar layout. That keeps the data path clean and gives scans a single place to read per-row-group stats for pruning.

```
  milvus.v2_zoned_row_group
    data: current V2 ChunkedLayout<RG StructLayout>
    zones: Struct/container
      col_a: Vortex-compatible ZoneMap table, one row per RG
      col_b: Vortex-compatible ZoneMap table, one row per RG
      ...
      __rg_boundaries: __rg_start/__rg_len table, one row per RG
```

The boundary table lets the reader map row-group-level pruning results back to the requested row range, so we can prune whole row groups without changing the normal scan/projection behavior. This also wires the new layout into the Rust bridge and adds V2 coverage for row group splitting, zonemap layout shape, and filter pruning.

### DIFF
--- a/cpp/src/format/bridge/rust/include/vortex_bridge.h
+++ b/cpp/src/format/bridge/rust/include/vortex_bridge.h
@@ -233,6 +233,10 @@ class VortexFile {
   // otherwise, return vector with uncompressed sizes
   std::vector<uint64_t> GetUncompressedSizes() const;
 
+  std::string RootLayoutEncoding() const;
+  uint64_t RowGroupZoneMapCount() const;
+  bool RowGroupZoneMapDataBeforeZones() const;
+
   private:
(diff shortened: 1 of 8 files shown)


jiaqizho:
base on !473

PR #531, opened 2026-05-14
enhance: make input file opens lazy and count real reads

jiaqizho:
This change makes input file opens lazy. Before this, opening an input file on S3 or Azure could immediately trigger remote metadata calls like HEAD or GetProperties just to build the file handle. Now those calls are deferred until the caller actually needs size or metadata, and normal range reads can cache the object size from the read response when the provider returns enough information. This keeps simple opens cheap while still preserving the same behavior once size, metadata, or actual data is requested.

It also aligns the read metrics behavior across S3, Azure, and local filesystems. Opening an input stream or file no longer counts as a read; read_count and read_bytes are updated only after bytes are successfully returned to the caller. That makes the metrics reflect real read IO instead of handle creation, and avoids inflating counts for lazy opens or EOF reads.

The tests cover lazy opens, size caching, real read counting, EOF handling, and the updated provider expectations.

### DIFF
--- a/cpp/include/milvus-storage/filesystem/observable.h
+++ b/cpp/include/milvus-storage/filesystem/observable.h
@@ -165,6 +165,7 @@ class MetricsInputStream : public arrow::io::InputStream {
   arrow::Result<int64_t> Read(int64_t nbytes, void* out) override {
     ARROW_ASSIGN_OR_RAISE(auto bytes_read, stream_->Read(nbytes, out));
     if (bytes_read > 0) {
+      metrics_->IncrementReadCount();
       metrics_->IncrementReadBytes(bytes_read);
     }
     return bytes_read;
@@ -173,6 +174,7 @@ class MetricsInputStream : public arrow::io::InputStream {
   arrow::Result<std::shared_ptr<arrow::Buffer>> Read(int64_t nbytes) override {
     ARROW_ASSIGN_OR_RAISE(auto buffer, stream_->Read(nbytes));
     if (buffer && buffer->size() > 0) {
+      metrics_->IncrementReadCount();
       metrics_->IncrementReadBytes(buffer->size());
     }
     return buffer;
@@ -214,6 +216,7 @@ class MetricsRandomAccessFile : public arrow::io::RandomAccessFile {
   arrow::Result<int64_t> Read(int64_t nbytes, void* out) override {
     ARROW_ASSIGN_OR_RAISE(auto bytes_read, file_->Read(nbytes, out));
     if (bytes_read > 0) {
+      metrics_->IncrementReadCount();
       metrics_->IncrementReadBytes(bytes_read);
     }
     return bytes_read;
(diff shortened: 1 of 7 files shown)


jiaqizho:
### Code review

cpp/src/filesystem/azure/azurefs.cc
```diff
@@ -931,9 +917,17 @@ class ObjectInputFile final : public io::RandomAccessFile {
...
+      auto result =
+          blob_client_->DownloadTo(reinterpret_cast<uint8_t*>(out), nbytes,
+                                   download_options);
+      const auto bytes_read = result.Value.ContentRange.Length.Value();
+      if (bytes_read < 0 || bytes_read > nbytes) {
+        return Status::IOError("Unexpected DownloadTo Content-Range length ",
+                               bytes_read, " for range read of ", nbytes,
+                               " bytes");
+      }
+      RETURN_NOT_OK(CacheContentLengthFromRead(result.Value));
```
tedxu, line 929: Consider cache the size only if it is absent.

cpp/src/filesystem/s3/s3_filesystem.cpp
```diff
@@ -535,9 +555,7 @@ class ObjectInputFile final : public arrow::io::RandomAccessFile {
     FIU_RETURN_ON(FIUKEY_S3FS_READAT_FAIL,
                   arrow::Status::IOError(fmt::format("Injected fault: {}", FIUKEY_S3FS_READAT_FAIL)));
     ARROW_RETURN_NOT_OK(CheckClosed());
-    ARROW_RETURN_NOT_OK(CheckPosition(position, "read"));
```
tedxu, line 538: Why is the `CheckPosition` call removed?

cpp/include/milvus-storage/filesystem/observable.h
```diff
@@ -165,6 +165,7 @@ class MetricsInputStream : public arrow::io::InputStream {
   arrow::Result<int64_t> Read(int64_t nbytes, void* out) override {
     ARROW_ASSIGN_OR_RAISE(auto bytes_read, stream_->Read(nbytes, out));
     if (bytes_read > 0) {
+      metrics_->IncrementReadCount();
```
tedxu, line 168: +1 to add the read count metrics, but why is that gated by `bytes read > 0`?

cpp/src/filesystem/s3/s3_filesystem.cpp
jiaqizho: no longer use the `CheckPosition` to check. Used `ARROW_ASSIGN_OR_RAISE(nbytes, GetReadSize(position, nbytes));` which contain the `CheckPosition` logical

cpp/include/milvus-storage/filesystem/observable.h
jiaqizho: If `bytes_read == 0` the `read`  will be success. But we won't do the s3/azure request. So i think that is fine we not increment the read count?

cpp/src/filesystem/azure/azurefs.cc
jiaqizho: ack

PR #533, opened 2026-05-15
enhance: support FixedSizeBinary in nested types for Vortex

jiaqizho:
Vortex does not currently accept Arrow FixedSizeBinary directly, but callers can still pass schemas that contain it, including nested shapes like List<FixedSizeBinary(512)>. We should handle that at the milvus-storage bridge boundary instead of asking callers to rewrite their schemas or patching the vendored Vortex code.

This change teaches the Vortex bridge to translate FixedSizeBinary into FixedSizeList<UInt8> before data enters Vortex, and to translate it back when data is read with the original caller schema. The schema conversion now follows the same recursive shape as Vortex's Arrow dtype handling for list, large list, list view, fixed-size list, struct, and dictionary types, while only overriding the unsupported FixedSizeBinary leaf. When no conversion is needed, the bridge keeps using the original schema and array path.

The array conversion mirrors the schema conversion so nested data stays consistent with the converted Vortex dtype. FixedSizeBinary and FixedSizeList<UInt8> share the same physical byte layout, so the conversion rebuilds Arrow metadata and child arrays while reusing the underlying buffers instead of copying vector payloads.

### DIFF
--- a/cpp/src/format/bridge/rust/src/vortex_bridgeimpl.rs
+++ b/cpp/src/format/bridge/rust/src/vortex_bridgeimpl.rs
@@ -12,7 +12,7 @@ use arrow_array::ffi::FFI_ArrowSchema;
 use arrow_array::ffi_stream::FFI_ArrowArrayStream;
 use arrow_array::{
     Array, ArrayRef as ArrowArrayRef, FixedSizeBinaryArray, FixedSizeListArray, RecordBatch,
-    RecordBatchReader, UInt8Array,
+    RecordBatchReader, UInt8Array, make_array,
 };
 use arrow_data::ArrayData;
 use arrow_data::ffi::FFI_ArrowArray;
@@ -307,50 +307,67 @@ pub(crate) fn select(fields: Vec<String>, child: Box<Expr>) -> Box<Expr> {
  * Both types have identical memory layout, enabling zero-copy conversion.
  */
 
-/// Check if a DataType is FixedSizeBinary
-fn is_fixed_size_binary(dt: &DataType) -> bool {
-    matches!(dt, DataType::FixedSizeBinary(_))
+fn to_vortex_field(field: &Field) -> Option<Arc<Field>> {
+    to_vortex_data_type(field.data_type())
+        .map(|data_type| Arc::new(field.clone().with_data_type(data_type)))
 }
 
-/// Check if schema contains any FixedSizeBinary fields
-fn schema_has_fixed_size_binary(schema: &Schema) -> bool {
-    schema
-        .fields()
-        .iter()
-        .any(|f| is_fixed_size_binary(f.data_type()))
-}
-
-/// Convert FixedSizeBinary type to FixedSizeList<u8>
-fn fixed_size_binary_to_list_type(byte_width: i32) -> DataType {
-    DataType::FixedSizeList(
-        Arc::new(Field::new("item", DataType::UInt8, false)),
-        byte_width,
-    )
+fn to_vortex_data_type(dt: &DataType) -> Option<DataType> {
+    match dt {
+        DataType::FixedSizeBinary(byte_width) => Some(DataType::FixedSizeList(
+            Arc::new(Field::new("item", DataType::UInt8, false)),
+            *byte_width,
+        )),
+        DataType::List(field) => to_vortex_field(field).map(DataType::List),
+        DataType::LargeList(field) => to_vortex_field(field).map(DataType::LargeList),
+        DataType::ListView(field) => to_vortex_field(field).map(DataType::ListView),
+        DataType::LargeListView(field) => to_vortex_field(field).map(DataType::LargeListView),
+        DataType::FixedSizeList(field, list_size) => to_vortex_field(field)
+            .map(|field| DataType::FixedSizeList(field, *list_size)),
+        DataType::Struct(fields) => {
+            let mut changed = false;
+            let converted_fields: Vec<_> = fields
+                .iter()
+                .map(|field| match to_vortex_field(field) {
+                    Some(converted_field) => {
+                        changed = true;
+                        converted_field
+                    }
+                    None => field.clone(),
+                })
+                .collect();
+            if changed {
+                Some(DataType::Struct(converted_fields.into()))
+            } else {
+                None
+            }
+        }
+        DataType::Dictionary(key_type, value_type) => to_vortex_data_type(value_type)
+            .map(|value_type| DataType::Dictionary(key_type.clone(), Box::new(value_type))),
+        _ => None,
+    }
 }
 
 /// Convert schema: replace FixedSizeBinary with FixedSizeList<u8>
-fn convert_schema_for_vortex(schema: &Schema) -> Schema {
-    if !schema_has_fixed_size_binary(schema) {
-        return schema.clone();
-    }
-
+fn convert_schema_for_vortex(schema: &Schema) -> Option<Schema> {
+    let mut changed = false;
     let new_fields: Vec<_> = schema
         .fields()
         .iter()
-        .map(|field| {
-            if let DataType::FixedSizeBinary(byte_width) = field.data_type() {
-                Arc::new(Field::new(
-                    field.name(),
-                    fixed_size_binary_to_list_type(*byte_width),
-                    field.is_nullable(),
-                ))
-            } else {
-                field.clone()
+        .map(|field| match to_vortex_field(field) {
+            Some(converted_field) => {
+                changed = true;
+                converted_field
             }
+            None => field.clone(),
         })
         .collect();
 
-    Schema::new_with_metadata(new_fields, schema.metadata().clone())
+    if changed {
+        Some(Schema::new_with_metadata(new_fields, schema.metadata().clone()))
+    } else {
+        None
+    }
 }
 
 /// Convert FixedSizeBinary array to FixedSizeList<u8> array (zero-copy)
@@ -360,10 +377,12 @@ fn convert_fixed_size_binary_to_list(array: &FixedSizeBinaryArray) -> ArrowArray
     // Get the underlying data buffer directly (zero-copy via Arc)
     let data = array.to_data();
     let values_buffer = data.buffers()[0].clone();
+    let values_offset = data.offset() * byte_width as usize;
 
     // Create UInt8Array from the buffer directly (zero-copy)
     let child_data = ArrayData::builder(DataType::UInt8)
         .len(array.len() * byte_width as usize)
+        .offset(values_offset)
         .add_buffer(values_buffer)
         .build()
         .expect("Failed to build UInt8 ArrayData");
@@ -382,7 +401,10 @@ fn convert_fixed_size_binary_to_list(array: &FixedSizeBinaryArray) -> ArrowArray
 }
 
 /// Convert FixedSizeList<u8> array to FixedSizeBinary array (zero-copy)
-fn convert_list_to_fixed_size_binary(array: &FixedSizeListArray, byte_width: i32) -> ArrowArrayRef {
+fn convert_list_to_fixed_size_binary(
+    array: &FixedSizeListArray,
+    byte_width: i32,
+) -> ArrowArrayRef {
     let values = array.values();
 
     // Get the u8 child array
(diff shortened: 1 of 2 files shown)


cpp/src/format/bridge/rust/src/vortex_bridgeimpl.rs
```diff
@@ -394,101 +416,233 @@ fn convert_list_to_fixed_size_binary(array: &FixedSizeListArray, byte_width: i32
     // Get the underlying buffer directly (zero-copy via Arc)
     let u8_data = u8_array.to_data();
     let values_buffer = u8_data.buffers()[0].clone();
+    let byte_width_usize = byte_width as usize;
```
liliu-z, line 603: The offset arithmetic in `convert_list_to_fixed_size_binary` (line 419–425) and `convert_fixed_size_binary_to_list` (line 380) is correct on paper, but the new test only exercises unsliced arrays where `offset == 0`, making old and new code indistinguishable. The entire purpose of the offset code is sliced arrays. Add at least one test that slices a `List<FSB>` before writing, and another that exercises the FSL→FSB direction on a sliced result, so the next refactor cannot silently break this.

cpp/src/format/bridge/rust/src/vortex_bridgeimpl.rs
```diff
@@ -394,101 +416,233 @@ fn convert_list_to_fixed_size_binary(array: &FixedSizeListArray, byte_width: i32
...
+    assert_eq!(
+        u8_data.offset() % byte_width_usize,
```
liliu-z: The `assert_eq!(u8_data.offset() % byte_width_usize, 0, ...)` at line 421 will SIGABRT a Milvus process if a downstream producer hands back a `FixedSizeList<u8>` whose child is sliced at a non-aligned boundary. Rust panics across `extern "C"` are undefined behavior. This applies to ~8 new `expect(...)` calls added by this PR as well (`rebuild_array_with_children`, `to_vortex_data_type(...).expect(...)` in the array converter, etc.). Return `VortexError` or `ArrowError` and let it propagate — the surrounding functions already use `Result`.

cpp/src/format/bridge/rust/src/vortex_bridgeimpl.rs
```diff
@@ -394,101 +416,233 @@ fn convert_list_to_fixed_size_binary(array: &FixedSizeListArray, byte_width: i32
...
+                            changed = true;
+                            converted_child.to_data()
+                        }
+                        None => child.clone(),
+                    }
+                })
+                .collect();
+            if changed {
+                let converted_type = to_vortex_data_type(array.data_type())
+                    .expect("Expected converted Arrow struct type for Vortex");
```
liliu-z: `to_vortex_arrow_array` calls `to_vortex_data_type(array.data_type()).expect(...)` after already recursively converting the child array. The runtime array type and the schema type are assumed identical — they usually are, but coupling two parallel match trees via `expect` is fragile. Consider returning the converted child type alongside the child array from the recursion (e.g., `Option<(ArrowArrayRef, DataType)>`) so the parent builds its type directly without re-walking. This would eliminate the `expect` points in both the List and Struct branches.

cpp/src/format/bridge/rust/src/vortex_bridgeimpl.rs
```diff
@@ -307,50 +307,67 @@ pub(crate) fn select(fields: Vec<String>, child: Box<Expr>) -> Box<Expr> {
  * Both types have identical memory layout, enabling zero-copy conversion.
  */
 
-/// Check if a DataType is FixedSizeBinary
-fn is_fixed_size_binary(dt: &DataType) -> bool {
-    matches!(dt, DataType::FixedSizeBinary(_))
+fn to_vortex_field(field: &Field) -> Option<Arc<Field>> {
```
liliu-z: `to_vortex_data_type` returns `None` for any container it doesn't enumerate (Map, Union, RunEndEncoded). If a caller passes e.g. `Map<…, FixedSizeBinary>`, the schema reaches Vortex unchanged and `RustDType::from_arrow` will fail with a less helpful error. Add an explicit error or document the unsupported set in the file-level comment at line 302.

cpp/src/format/bridge/rust/src/vortex_bridgeimpl.rs
```diff
@@ -307,50 +307,67 @@ pub(crate) fn select(fields: Vec<String>, child: Box<Expr>) -> Box<Expr> {
...
+                        converted_field
+                    }
+                    None => field.clone(),
+                })
+                .collect();
+            if changed {
+                Some(DataType::Struct(converted_fields.into()))
+            } else {
+                None
+            }
```
liliu-z, line 432: The PR claims support for LargeList, ListView, LargeListView, FixedSizeList<FSB>, Struct{FSB}, Dictionary<*, FSB>, and arbitrary nesting. The single test covers only one shape: `List<FSB(512)>`. Missing coverage for: (1) `Struct{ fsb: FSB(N), other: Int32 }` — exercises the `changed` flag path, (2) `List<List<FSB>>` or `List<Struct{FSB}>` — exercises multi-level recursion, (3) nullable FSB column with some nulls — exercises null-buffer propagation, (4) sliced writes (see offset issue). These are exactly the code paths where the risky new conversion logic lives.

cpp/src/format/bridge/rust/src/vortex_bridgeimpl.rs
```diff
@@ -394,101 +416,233 @@ fn convert_list_to_fixed_size_binary(array: &FixedSizeListArray, byte_width: i32
...
+            let child_array = make_array(child_data[0].clone());
+            let converted_child = to_arrow_array(&child_array, field.data_type())?;
+            Some(rebuild_array_with_children(
+                array,
+                target_type.clone(),
+                vec![converted_child.to_data()],
+                "Failed to build Arrow nested array converted from Vortex",
+            ))
+        }
+        DataType::Struct(fields) => {
```
liliu-z: The `.zip(original_schema.fields().iter())` in `convert_record_batch_from_vortex` (and similarly in the struct conversion) truncates to the shorter side if `batch.num_columns() != original_schema.fields().len()`. `RecordBatch::try_new` then panics via `expect` with a confusing error. Validate child count matches field count upfront and return a clear error.

cpp/src/format/bridge/rust/src/vortex_bridgeimpl.rs
```diff
@@ -307,188 +307,573 @@ pub(crate) fn select(fields: Vec<String>, child: Box<Expr>) -> Box<Expr> {
...
         byte_width,
         Arc::new(child_array),
         nulls,
-    ))
+    )?))
 }
 
 /// Convert FixedSizeList<u8> array to FixedSizeBinary array (zero-copy)
-fn convert_list_to_fixed_size_binary(array: &FixedSizeListArray, byte_width: i32) -> ArrowArrayRef {
+fn convert_list_to_fixed_size_binary(
```
liliu-z, line 562: The function takes `byte_width` from the caller's target schema and an `array: &FixedSizeListArray` from Vortex, but never checks that `array.value_length() == byte_width`. If the caller's schema disagrees with the file's stored width (e.g. schema evolution bug, corrupted metadata), the conversion silently reinterprets the byte stream. Worst case: source width > target width reads past each row's stride into the next row's bytes — no panic, just corrupted vectors. Fix: add an explicit `assert_eq!(array.value_length(), byte_width as usize)` or return an error before using the child buffer.

cpp/src/format/bridge/rust/src/vortex_bridgeimpl.rs
```diff
@@ -307,188 +307,573 @@ pub(crate) fn select(fields: Vec<String>, child: Box<Expr>) -> Box<Expr> {
...
+                    "Failed to build Arrow struct array converted from Vortex",
+                )?))
             } else {
-                col.clone()
+                Ok(None)
             }
-        })
-        .collect();
+        }
+        DataType::Dictionary(_, value_type) => {
```
liliu-z: In the `to_arrow_array` Dictionary branch, when `child_data.is_empty()` the code returns `Ok(None)`. Valid Arrow dictionary data always has exactly one values child. Returning `None` here causes the schema to say `Dictionary<_, FSB>` while the batch column is missing, leading to an unhelpful `RecordBatch::try_new` error far from the root cause. This should be an explicit error indicating malformed dictionary data.

cpp/src/format/bridge/rust/src/vortex_bridgeimpl.rs
```diff
@@ -307,188 +307,573 @@ pub(crate) fn select(fields: Vec<String>, child: Box<Expr>) -> Box<Expr> {
  * Both types have identical memory layout, enabling zero-copy conversion.
  */
 
-/// Check if a DataType is FixedSizeBinary
-fn is_fixed_size_binary(dt: &DataType) -> bool {
-    matches!(dt, DataType::FixedSizeBinary(_))
+#[derive(Clone)]
```
liliu-z, line 310: In `scan_builder_into_stream`, `needs_conversion = vortex_schema != original_schema` relies on schema inequality as a proxy for whether FSB conversion was applied. While currently correct because both schemas are derived together, this is fragile — schemas could differ for non-FSB reasons in the future. The right signal is an explicit `needs_conversion: bool` returned from `convert_schema_for_vortex`.

liliu-z:
**cpp/src/format/bridge/rust/src/lib.rs:482** Arrow dictionary key types are constrained to integer types by specification, so `contains_fixed_size_binary(key_type)` can never return true. This recursion is unreachable and inconsistent with `build_vortex_array_conversion`, which does not walk the key type.

liliu-z:
**rust/lance-encoding/src/vortex_bridge.rs:0** Add a `plan: Option<VortexSchemaConversion>` field to `ConvertingRecordBatchReader`. Build it in `scan_builder_into_stream` and use it in `next()` to drive column conversion, achieving read/write symmetry and eliminating runtime DataType inference and unnecessary recursion into non-FSB nested columns.

liliu-z:
**rust/lance-encoding/src/vortex_bridge.rs:0** The Dictionary branch in the read-side `to_arrow_array` has no test coverage and the write side blocks Dictionary input. This is unreachable dead code. Delete it, or if kept, add an end-to-end test to prove it works.

liliu-z:
**rust/lance-encoding/src/vortex_bridge.rs:0** When `to_arrow_array` encounters Map, Union, or RunEndEncoded types on the read path, it should return an explicit `unsupported_container_error` like the write side does, instead of silently returning `Ok(None)`.

liliu-z:
**cpp/src/vortex/tests/test_vortex_writer.cpp:0** Add a C++ end-to-end test that slices a FixedSizeBinary array and then writes it, to exercise the `values_offset = data.offset() * byte_width` arithmetic and prove there are no offset calculation blind spots.

liliu-z:
**rust/lance-encoding/src/vortex_bridge.rs:0** The `inner_writer.take().unwrap()` is historical debt unrelated to this PR's FSB conversion work. File a separate issue to track replacing it with proper error handling.

cpp/src/format/bridge/rust/src/vortex_bridgeimpl.rs
```diff
@@ -307,188 +307,564 @@ pub(crate) fn select(fields: Vec<String>, child: Box<Expr>) -> Box<Expr> {
...
```
liliu-z, line 310: When converting `FixedSizeList<UInt8>` to `FixedSizeBinary`, if the child UInt8 array has a null bitmap, it is silently dropped. `FixedSizeBinary` has no per-byte null semantics, so the now-undefined byte slots are reinterpreted as valid data. The write path produces non-nullable children so round-trips are fine, but reading external `FixedSizeList<UInt8>` data with child nulls causes silent data corruption. Add a check that rejects conversion when the child array has `null_count != 0` (or when any null buffer is attached).

cpp/src/format/bridge/rust/src/vortex_bridgeimpl.rs
```diff
@@ -307,188 +307,564 @@ pub(crate) fn select(fields: Vec<String>, child: Box<Expr>) -> Box<Expr> {
...
-/// Convert FixedSizeBinary type to FixedSizeList<u8>
-fn fixed_size_binary_to_list_type(byte_width: i32) -> DataType {
-    DataType::FixedSizeList(
-        Arc::new(Field::new("item", DataType::UInt8, false)),
-        byte_width,
-    )
+struct VortexSchemaConversion {
+    schema: Schema,
+    fields: Vec<Option<VortexArrayConversion>>,
 }
```
liliu-z, line 339: The `VortexSchemaConversion` plan is computed in `scan_builder_with_schema` but not stored, so it must be recomputed from scratch in `scan_builder_into_stream`. Store `Option<VortexSchemaConversion>` on `VortexScanBuilder` to avoid the redundant schema walk.

cpp/test/format/vortex/vortex_basic_test.cpp
```diff
@@ -112,6 +114,198 @@ INSTANTIATE_TEST_SUITE_P(V1V2,
...
+                                      int64_t source_row_offset,
+                                      int64_t num_rows,
+                                      int vectors_per_row,
+                                      int vector_width) {
+  ASSERT_NE(list_array, nullptr);
+  ASSERT_EQ(num_rows, list_array->length());
+
+  for (int64_t row = 0; row < num_rows; ++row) {
+    ASSERT_FALSE(list_array->IsNull(row));
+    auto values = std::dynamic_pointer_cast<arrow::FixedSizeBinaryArray>(list_array->value_slice(row));
```
liliu-z, line 233: The test suite covers `List<FixedSizeBinary>` but omits `LargeList<FixedSizeBinary>` and `FixedSizeList<FixedSizeBinary>` as outer containers. These are valid Arrow types that callers can produce and should be tested to prevent regressions in the recursive conversion path.

cpp/test/format/vortex/vortex_basic_test.cpp
```diff
@@ -112,6 +114,198 @@ INSTANTIATE_TEST_SUITE_P(V1V2,
                            return "V" + std::to_string(info.param);
                          });
 
+namespace {
```
liliu-z, line 117: The width-mismatch test asserts on an exact error message string from arrow-rs. When arrow-rs updates its error wording, this test will break. The failure mode is noisy (test failure, not silent pass), but still brittle. Use a substring or pattern match instead.

liliu-z:
**cpp/src/format/vortex/vortex_writer.cpp:65** `VortexFileWriter::Flush()` is declared `arrow::Status`, but the underlying cxx bridge call (`VortexWriter::Write`) propagates Rust errors as `VortexException`. This means a Rust `Err` becomes an exception while a Flush-level error becomes a Status — two failure channels in one function. Not introduced by this PR, but the new Dictionary-rejection test now relies on the exception channel, making the duality more visible. Worth a follow-up to unify error reporting (e.g., wrap the cxx call in try/catch converting to Status).

cpp/src/format/bridge/rust/src/vortex_bridgeimpl.rs
```diff
@@ -307,188 +307,574 @@ pub(crate) fn select(fields: Vec<String>, child: Box<Expr>) -> Box<Expr> {
  * Both types have identical memory layout, enabling zero-copy conversion.
  */
 
-/// Check if a DataType is FixedSizeBinary
-fn is_fixed_size_binary(dt: &DataType) -> bool {
-    matches!(dt, DataType::FixedSizeBinary(_))
+#[derive(Clone)]
```
liliu-z, line 310: Multiplications like `array.len() * byte_width` and `data.offset() * byte_width` can overflow silently on large arrays or wide FSB types, leading to incorrect slicing or out-of-bounds access. Use `checked_mul` and propagate the error instead of relying on wrapping arithmetic.

cpp/test/format/vortex/vortex_basic_test.cpp
```diff
@@ -124,6 +401,362 @@ TEST_P(VortexBasicTest, TestBasicWrite) {
...
+  ASSERT_AND_ASSIGN(auto cgfile, vx_writer.Close());
+  ASSERT_EQ(num_rows, cgfile.end_index);
+
+  auto read_schema = arrow::schema({arrow::field("embedding", arrow::fixed_size_binary(read_width), false,
+                                                 arrow::key_value_metadata({ARROW_FIELD_ID_KEY}, {"100"}))});
+  auto vx_reader = vortex::VortexFormatReader(file_system_, read_schema, test_file_name_, properties_,
+                                              std::vector<std::string>{"embedding"});
+  ASSERT_STATUS_OK(vx_reader.open());
+
+  auto read_result = vx_reader.blocking_read(0, num_rows);
```
liliu-z, line 726: Both negative tests at lines 726 and 754 only check `ASSERT_FALSE(read_result.ok())`, which passes for any failure — including unrelated crashes. Assert that the error message contains expected text (e.g. the mismatched width or null-related reason) so the tests actually validate the bridge conversion path.

cpp/test/format/vortex/vortex_basic_test.cpp
```diff
@@ -124,6 +401,362 @@ TEST_P(VortexBasicTest, TestBasicWrite) {
...
+  ASSERT_AND_ASSIGN(auto cgfile, vx_writer.Close());
+  ASSERT_EQ(num_rows, cgfile.end_index);
+
+  auto read_schema = arrow::schema({arrow::field("embedding", arrow::fixed_size_binary(vector_width), false,
+                                                 arrow::key_value_metadata({ARROW_FIELD_ID_KEY}, {"100"}))});
+  auto vx_reader = vortex::VortexFormatReader(file_system_, read_schema, test_file_name_, properties_,
+                                              std::vector<std::string>{"embedding"});
+  ASSERT_STATUS_OK(vx_reader.open());
+
+  auto read_result = vx_reader.blocking_read(0, num_rows);
```
liliu-z, line 754: Same issue as line 726 — `ASSERT_FALSE(read_result.ok())` does not verify the failure came from the bridge conversion. Assert on the error message content.

cpp/src/format/bridge/rust/src/vortex_bridgeimpl.rs
```diff
@@ -307,188 +307,574 @@ pub(crate) fn select(fields: Vec<String>, child: Box<Expr>) -> Box<Expr> {
...
-/// Convert FixedSizeBinary type to FixedSizeList<u8>
-fn fixed_size_binary_to_list_type(byte_width: i32) -> DataType {
-    DataType::FixedSizeList(
-        Arc::new(Field::new("item", DataType::UInt8, false)),
-        byte_width,
-    )
+struct VortexSchemaConversion {
+    schema: Schema,
+    fields: Vec<Option<VortexArrayConversion>>,
 }
```
liliu-z, line 339: When no read schema is provided, the scan builder returns the physical Vortex schema, so FixedSizeBinary fields come back as FixedSizeList<UInt8>. Nested FSB support expands the surface area of this surprise. Document this behaviour on `get_schema()` and the no-schema `scan_builder` path.

cpp/src/format/bridge/rust/src/vortex_bridgeimpl.rs
```diff
@@ -307,188 +307,574 @@ pub(crate) fn select(fields: Vec<String>, child: Box<Expr>) -> Box<Expr> {
...
 
-/// Convert FixedSizeBinary type to FixedSizeList<u8>
-fn fixed_size_binary_to_list_type(byte_width: i32) -> DataType {
-    DataType::FixedSizeList(
-        Arc::new(Field::new("item", DataType::UInt8, false)),
-        byte_width,
-    )
+struct VortexSchemaConversion {
+    schema: Schema,
```
liliu-z, line 336: Every `write()` call recomputes a fresh `VortexSchemaConversion` from its incoming schema. Subsequent batches' converted schemas are not checked against the first batch's schema used to initialize the writer. If a caller sends batches with structurally different FSB widths at the same field path, Vortex catches the mismatch late. Pre-existing issue, but nested conversion makes it easier to trigger.

cpp/src/format/bridge/rust/src/vortex_bridgeimpl.rs
jiaqizho: 16 EiB, won't overflow.

cpp/test/format/vortex/vortex_basic_test.cpp
jiaqizho: over-design here...

cpp/test/format/vortex/vortex_basic_test.cpp
jiaqizho: over-design here...

cpp/src/format/bridge/rust/src/vortex_bridgeimpl.rs
jiaqizho: over-design here...

cpp/src/format/bridge/rust/src/vortex_bridgeimpl.rs
jiaqizho: no need check it twice.

PR #536, opened 2026-05-19
fix: prefer build-local runtime libraries for Python CI

jiaqizho:
Python CI previously built LD_LIBRARY_PATH by scanning every lib directory under `~/.conan2/`. When the GitHub Actions Conan cache contained multiple builds of the same dependency, the dynamic linker could resolve libmilvus-storage.so against a stale or incompatible shared library.

The observed failure happened during pytest dlopen: libmilvus-storage.so was loaded with a mismatched libfolly, which failed to resolve fLI::FLAGS_folly_iothreadpoolexecutor_max_read_at_once. This was a runtime library resolution issue, not a Python test or PR logic failure.

The Conan build already copies the shared libraries for the current build into cpp/build/Release/libs. Use that build-local libs directory, plus cpp/build/Release, as the runtime library path instead of the entire Conan cache. Also preload liblzma from the build output and add an ldd -r check so missing dependencies or unresolved symbols fail before pytest.

### DIFF
--- a/.github/workflows/python-ci.yml
+++ b/.github/workflows/python-ci.yml
@@ -103,7 +103,7 @@ jobs:
           # Use the dependency libraries copied for this build. Scanning the
           # whole Conan cache can put stale/incompatible libraries before the
           # ones that match libmilvus-storage.so.
-          echo "LD_LIBRARY_PATH=$BUILD_LIB_DIR:$BUILD_DIR:${LD_LIBRARY_PATH}" >> $GITHUB_ENV
+          echo "LD_LIBRARY_PATH=$BUILD_LIB_DIR:$BUILD_DIR${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}" >> $GITHUB_ENV
 
           # Preload the liblzma copied for this build to override system liblzma.
           BUILD_LZMA=$(find "$BUILD_LIB_DIR" -name "liblzma.so*" -type f 2>/dev/null | sort | head -n 1)
@@ -144,6 +144,8 @@ jobs:
 
       - name: Verify library dependencies
         run: |
+          set -o pipefail
+
           echo "LD_LIBRARY_PATH: $LD_LIBRARY_PATH"
           echo "libmilvus-storage dependency resolution:"
           ldd -r cpp/build/Release/libmilvus-storage.so 2>&1 | tee /tmp/libmilvus-storage.ldd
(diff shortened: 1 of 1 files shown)


.github/workflows/python-ci.yml
```diff
@@ -145,6 +146,12 @@ jobs:
       - name: Verify library dependencies
         run: |
           echo "LD_LIBRARY_PATH: $LD_LIBRARY_PATH"
+          echo "libmilvus-storage dependency resolution:"
+          ldd -r cpp/build/Release/libmilvus-storage.so 2>&1 | tee /tmp/libmilvus-storage.ldd
+          if grep -q "not found\\|undefined symbol" /tmp/libmilvus-storage.ldd; then
+            echo "Runtime dependency resolution failed"
```
liliu-z, line 153: The new `exit 1` on line 153 is designed to turn confusing pytest dlopen tracebacks into a hard CI failure. However, the job is marked `continue-on-error: true` at line 43, so branch protection won't enforce the failure. The new signal you've added won't actually gate anything until this is addressed. Consider a follow-up that either drops `continue-on-error` or splits the verify step into its own required job.

.github/workflows/python-ci.yml
```diff
@@ -145,6 +145,12 @@ jobs:
       - name: Verify library dependencies
         run: |
           echo "LD_LIBRARY_PATH: $LD_LIBRARY_PATH"
+          echo "libmilvus-storage dependency resolution:"
```
liliu-z, line 150: `find ~/.conan2/p -name "libglog.so*" -exec ldd {} \;` still searches the Conan cache for diagnostics, which is inconsistent with the rest of the step now using `$BUILD_LIB_DIR`. Either redirect it to scan `$BUILD_LIB_DIR` or remove it — as-is it's misleading noise.

.github/workflows/python-ci.yml
```diff
@@ -145,11 +145,17 @@ jobs:
...
+          ldd -r cpp/build/Release/libmilvus-storage.so 2>&1 | tee /tmp/libmilvus-storage.ldd
```
liliu-z, line 151: GitHub Actions' default shell is `bash -e`, not `bash -eo pipefail`. If `ldd -r` itself fails (e.g. the .so is missing or LD_PRELOAD blows up), `tee` exits 0 and the pipeline succeeds. The subsequent `grep` then runs on a possibly-empty file and passes silently. Add `set -o pipefail` at the top of this step, or check `${PIPESTATUS[0]}` after the pipeline. Alternatively, add an explicit `test -f cpp/build/Release/libmilvus-storage.so` guard before running `ldd`.

PR #539, opened 2026-05-25
fix: make writer close safe after partial column group initialization

jiaqizho:
Storage writes can fail after the writer has already planned multiple column groups but before every column group writer has been created. Milvus still calls close on that writer as a cleanup step, so close needs to handle this partially initialized state instead of assuming the metadata list and writer list always have the same size.

This change makes close work with the writers that actually exist, detects impossible internal state where there are more writers than groups, and returns an empty column group list when initialization did not finish. That keeps the cleanup path from producing invalid manifest data while still releasing the resources owned by the initialized column group writers.

The FFI close comment is also clarified to match this behavior: after a previous writer operation has failed, a successful close is only a cleanup result, and the returned column groups must not be treated as a valid dataset manifest.


congqixia (approved):
/lgtm

Issue #541, opened 2026-05-26
[Bug]: Vortex take scales linearly in K with no segment-level decode sharing


tedxu:
Vortex `take` shows two related symptoms on the same file: wall time
scales linearly with K (the number of requested rows) with no plateau,
and the same row group / segment is decoded once per requested index
inside it rather than once total. As a result, taking 10,000 rows
takes ~1067x longer than reading the whole 40,960-row file.

### Environment

- milvus-storage upstream main `8dd9fe1`
- Build: `RelWithDebInfo`, `USE_ASAN=False`, `-fno-omit-frame-pointer`, gcc-11
- Vortex via the patched 0.56.0 in `cpp/src/format/bridge/rust/_vortex_patched/`
- Host: 8-core x86_64 Linux KVM guest, local FS, single thread
- Default reader/writer properties:
  - `writer.vortex.format_version=2` (row-group layout, current default)
  - `writer.compression=zstd`, level 5
  - `reader.record_batch_max_rows=8192`
  - `DEFAULT_MAX_ROW_GROUP_SIZE = 1 MB`
- Workload: 40,960 rows × 128-dim float32 vector + scalar columns (synthetic
  data loader, Medium config)

### Observation 1 — take wall is linear in K, no plateau

Random index distribution, seed = 42, mean of 3 reps:

| K     | Parquet wall (ms) | Vortex wall (ms) | Ratio V/P |
|-------|-------------------|------------------|-----------|
| 10    | 53.7              | 10.9             | 0.20x     |
| 100   | 162.8             | 117.5            | 0.72x     |
| 1000  | 165.0             | 1,242.7          | 7.53x     |
| 10000 | 182.7             | 12,853.4         | 70.35x    |

Vortex fits `wall_ms = -18 + 1.287 * K` with **R² = 1.000**. Parquet
does not fit linear (R² = 0.27) and plateaus near ~165 ms once the
random sample touches all row groups.

### Observation 2 — locality is not exploited

K = 1000, three index distributions, mean of 3 reps:

| Distribution                | Vortex (ms) | Parquet (ms) |
|-----------------------------|-------------|--------------|
| Sequential (rows 0..999)    | 1,464       | 8.3          |
| Random (uniform)            | 1,146       | 163          |
| Clustered (5-row clumps)    | 1,180       | 164          |

Parquet sequential is **20x faster** than Parquet random for the same K
because the 1,000 contiguous rows fit in one row group. Vortex
[...]

PR #542, opened 2026-05-27
fix: share segment decodes across indices in Vortex take

tedxu:
See #541

Vortex take scaled linearly in K with no segment-level decode
sharing. The bridge unconditionally calls
with_split_row_indices(false) on every scan builder, which makes
the patched scan_builder.rs turn each requested index into its
own single-row split. Every per-index task then independently
decodes the segment containing its index, so K indices over S
segments cost O(K) decodes instead of O(min(K, S)).

This change keeps the per-index path only when K <= S, by
computing the file's natural split count at builder construction
and consulting it inside with_include_by_index. Above the
threshold attempt_split_ranges merges contiguous indices and
sparse selections fall back to Natural splits, so each segment
is decoded at most once. Range reads are unaffected:
Selection::All never enters the per-index branch.

Verified on the issue's reproducer (40,960 rows, 128-dim
float32, S ~= 20):

  K        before (ms)  after (ms)  speedup
  10           10.9        10.1     1.08x
  100         117.5         5.33     22x
  1000      1,242.7         6.25    199x
  10000    12,853.4        12.4    1036x

Peak RSS at K=10000 drops from ~10 GB to ~7 MB. Sequential
locality at K=1000 is now exploited (2.89 ms vs 6.62 ms random).
All Vortex take unit tests pass for both V1 and V2 layouts.

### DIFF
--- a/cpp/src/format/bridge/rust/src/vortex_bridgeimpl.rs
+++ b/cpp/src/format/bridge/rust/src/vortex_bridgeimpl.rs
@@ -1029,12 +1029,14 @@ impl VortexFile {
     }
 
     pub(crate) fn scan_builder(&self) -> Result<Box<VortexScanBuilder>> {
+        let num_natural_splits = self.inner.splits().map_err(VortexError::from)?.len();
         Ok(Box::new(VortexScanBuilder {
-            inner: self.inner.scan()?.with_split_row_indices(false),
+            inner: self.inner.scan()?,
             output_schema: None,
             original_schema: None,
             conversion_plan: None,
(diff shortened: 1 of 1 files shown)


jiaqizho:
`with_split_row_indices` determines whether we need to call `attempt_split_ranges`. `attempt_split_ranges` merges request ranges by **row distinctness**. Currently, the condition for `with_split_row_indices` to be true is `include_by_index.len() > self.num_natural_splits`; this may lead to IO read amplification.

Offline-talk, the current PR is merged first.

PR #546, opened 2026-05-28
feat: add vortex local format support

aoiasd:
## What this PR does

Adds the storage-side Vortex local format support used by Milvus. The goal is to let Milvus plan and load Vortex data at local-format cell granularity while keeping the existing generic Vortex reader behavior unchanged.

Main additions:

- Adds `VortexFooterReader` to read Vortex footer and optional zonemap metadata through Vortex-level semantics.
- Adds `VortexPlanner` to convert row ranges, offsets, and pushdown-capable predicates into selected cells and Vortex read plans.
- Adds `VortexTranslater` and shared cell metadata helpers so Milvus can map local-format cells to Vortex physical units consistently.
- Extends `VortexFormatReader` with Vortex-specific `read_with_plan` APIs for local-format reads; normal `read`, `take`, and range-read paths continue to use the existing reader behavior.
- Extends the Vortex C++/Rust bridge with the metadata, row-range, row-id projection, and split-row-index controls needed by planned local-format reads.
- Adds the reader property for Vortex split-row-index behavior, defaulting to `auto` so existing behavior is preserved unless explicitly overridden.

This PR is intended to expose the metadata and planned-read primitives required by the Milvus local format path without changing non-local-format Vortex reader semantics.


cpp/include/milvus-storage/format/vortex/vortex_footer_reader.h
```diff
@@ -0,0 +1,135 @@
...
+  public:
+  virtual ~VortexRangeFile() = default;
+
+  VortexRangeFile(const VortexRangeFile&) = delete;
+  VortexRangeFile& operator=(const VortexRangeFile&) = delete;
+
+  VortexRangeFile() = default;
+
+  virtual void Resize(uint64_t size) = 0;
+  virtual uint64_t size() const = 0;
```
jiaqizho: size() -> Size()

cpp/include/milvus-storage/format/vortex/vortex_footer_reader.h
```diff
@@ -0,0 +1,135 @@
...
+  virtual arrow::Status WriteAt(uint64_t offset, std::shared_ptr<arrow::Buffer> data) = 0;
```
jiaqizho: -> (const uint64_t &offset, const std::shared_ptr<arrow::Buffer> &data)

cpp/include/milvus-storage/format/vortex/vortex_footer_reader.h
```diff
@@ -0,0 +1,135 @@
...
+                                                                   std::string_view path);
+
+arrow::Status FillVortexRangeFile(const std::shared_ptr<arrow::io::RandomAccessFile>& source_file,
+                                  const std::shared_ptr<VortexRangeFile>& range_file,
+                                  uint64_t offset,
+                                  uint64_t length);
+
+std::vector<ByteRange> MergeByteRanges(std::vector<ByteRange> ranges);
+
+enum class VortexPhysicalGranularity : uint64_t {
```
jiaqizho: `: uint16_t ` is fine. no need uint64_t

cpp/include/milvus-storage/format/vortex/vortex_format_reader.h
```diff
@@ -33,6 +45,7 @@ class VortexFormatReader final : public FormatReader, public std::enable_shared_
                      const std::vector<std::string>& needed_columns,
                      uint64_t file_size = 0,
                      uint64_t footer_size = 0);
+  ~VortexFormatReader() override;
```
jiaqizho, line 49: `VortexFormatReader::~VortexFormatReader() = default;` in cpp

what for?

cpp/include/milvus-storage/format/vortex/vortex_format_reader.h
```diff
@@ -60,10 +73,10 @@ class VortexFormatReader final : public FormatReader, public std::enable_shared_
   void set_predicate(const std::string& predicate) override;
 
   // get the row ranges(splits) of the file
-  inline std::vector<uint64_t> row_ranges() const { return vxfile_->Splits(); }
+  std::vector<uint64_t> row_ranges() const;
 
   // get the total rows of the file
-  inline size_t rows() const { return vxfile_->RowCount(); }
+  size_t rows() const;
```
jiaqizho, line 80: why remove the `inline` ?

cpp/include/milvus-storage/format/vortex/vortex_footer_reader.h
```diff
@@ -0,0 +1,135 @@
...
+  virtual arrow::Result<int64_t> ReadAt(int64_t position, int64_t nbytes, void* out) const = 0;
+  virtual arrow::Result<std::shared_ptr<arrow::Buffer>> ReadAt(int64_t position, int64_t nbytes) const = 0;
+  virtual void Punch(uint64_t offset, uint64_t length) = 0;
+};
+
+class VortexRangeFileProvider {
+  public:
+  virtual ~VortexRangeFileProvider() = default;
+
+  virtual arrow::Result<std::shared_ptr<VortexRangeFile>> GetVortexRangeFile(std::string_view path) const = 0;
```
jiaqizho: please check all std::string_view. consider replace with the const std::string & path. 

1. The lifecycle of `std::string_view` is completely identical to that of the original `std::string`. Excessive use of `std::string_view` as a parameter can make the lifecycle unnecessarily complex.

2. Passing `const std::string &` is essentially the same as passing `const std::string_view`.

3. We should only consider using `const std::string_view &` as a parameter in a single method with clear semantics that is not passed down.

4. `std::string_view` should never be used as a member variable.

cpp/include/milvus-storage/ffi_filesystem_c.h
```diff
@@ -55,6 +56,18 @@ typedef struct LoonFileInfoList {
   uint32_t count;
 } LoonFileInfoList;
 
+/**
+ * A borrowed read buffer returned by reader_readat_buffer.
+ *
+ * The data pointer is valid until handle is destroyed by
+ * loon_filesystem_reader_buffer_destroy.
+ */
+typedef struct LoonFileSystemReadBuffer {
```
jiaqizho: we really need this struct to hold the life-cycle ?

```
virtual Result<int64_t> ReadAt(int64_t position, int64_t nbytes, void* out);
```

replace with this function? then the caller will be the owner of out-buffer.

The FFI function is `loon_filesystem_reader_readat`.

jiaqizho: So the point is, just replace the `loon_filesystem_reader_readat_buffer` with `loon_filesystem_reader_readat`. which will make code more clear.

cpp/include/milvus-storage/format/vortex/vortex_translater.h
```diff
@@ -0,0 +1,113 @@
...
+#include <memory>
+#include <string>
+#include <vector>
+
+#include <arrow/filesystem/filesystem.h>
+#include <arrow/result.h>
+
+#include "milvus-storage/format/vortex/vortex_planner.h"
+
+#ifndef MILVUS_STORAGE_WITH_MILVUS_COMMON
```
jiaqizho: we do import the milvus-common , what this logical for?

cpp/include/milvus-storage/format/vortex/vortex_planner.h
```diff
@@ -0,0 +1,128 @@
...
+};
+
+struct VortexReadRange {
+  VortexPhysicalGranularity granularity = VortexPhysicalGranularity::kUnknown;
+  uint64_t physical_unit_index = 0;
+  uint64_t row_start = 0;
+  uint64_t row_end = 0;
+};
+
+struct VortexReadPlan {
```
jiaqizho, line 32: `vortex_planner.h` exposes a lot of implementation detail. Only the public read-plan shape needs to be visible to callers of `VortexFormatReader::read_with_plan`; the cell metadata/build/validation helpers look internal to planner/translator and should preferably move to vortex_planner.cpp or behind a pImpl/internal header. This reduces ABI/header surface and avoids making planner internals part of the public API accidentally.

jiaqizho: also check `vortex_translater.h` and `vortex_footer_reader.h`

cpp/include/milvus-storage/format/vortex/vortex_footer_reader.h
```diff
@@ -0,0 +1,135 @@
...
+  std::vector<VortexFlatSegmentRange> flat_segments;
+};
+
+struct VortexFieldLayout {
+  VortexPhysicalGranularity granularity = VortexPhysicalGranularity::kUnknown;
+  std::vector<VortexFlatUnit> flats;
+  std::vector<VortexRowGroupUnit> row_groups;
+};
+
+class VortexFooterReader {
```
jiaqizho, line 31: You must know one thing: header files in C++ are what coders tell callers what functionality they have, so you should move any unnecessary structs/functions to the cpp file as much as possible.

cpp/src/format/vortex/vortex_format_reader.cpp
```diff
@@ -57,21 +60,131 @@ uint8_t arrow_type_to_tag(const arrow::DataType& dt) {
       return 5;  // Other
   }
 }
```
jiaqizho, line 62: why removed the static inline ? please check other place.

The internal function(in cpp, not the member function of class) should declared as `static inline` or `static`

cpp/include/milvus-storage/format/vortex/vortex_format_reader.h
aoiasd: This is needed because `VortexFormatReader` owns `std::unique_ptr<VortexFile>`, while `VortexFile` is only forward-declared in the header. Defining the destructor in the cpp makes `VortexFile` a complete type when `std::default_delete<VortexFile>` is instantiated. Keeping it defaulted in the header can fail to compile for translation units that include this header without the full `VortexFile` definition.

cpp/src/format/bridge/rust/src/vortex_bridgeimpl.rs
```diff
@@ -1159,6 +1171,92 @@ impl VortexFile {
 
         Ok(max_data_offset < min_zones_offset)
     }
+
+    pub(crate) fn row_group_zone_map_segment_ids(&self) -> Result<Vec<u64>> {
+        let root = self.inner.footer().layout();
+        if root.encoding_id().as_ref() != "milvus.v2_zoned_row_group" {
```
jiaqizho: use the `LAYOUT_ID` which defined in `cpp/src/format/bridge/rust/src/vortex_layout_strategy_v2.rs`

cpp/src/format/bridge/rust/src/vortex_bridgeimpl.rs
```diff
@@ -1159,6 +1171,92 @@ impl VortexFile {
...
+            return Ok(Vec::new());
+        }
+
+        let zones_child = root.child(1)?;
```
jiaqizho: verify len of root, and use the error as return. no need use `?` here.

cpp/src/format/bridge/rust/src/vortex_bridgeimpl.rs
```diff
@@ -1159,6 +1171,92 @@ impl VortexFile {
...
+        let mut segment_ids = Vec::new();
+        collect_layout_segment_ids(&zones_child, &mut segment_ids)?;
```
jiaqizho: We should use error path return whenever possible, rather than causing it to panic directly.

`?` will make process direct panic.

cpp/src/format/bridge/rust/src/vortex_bridgeimpl.rs
```diff
@@ -1159,6 +1171,92 @@ impl VortexFile {
...
+            vec![footer_start, file_size - footer_start]
+        }
+    }
+
+    /// Returns [offset, length] for a given flat segment ID.
+    pub(crate) fn segment_bytes(&self, flat_segment_id: u64) -> Vec<u64> {
+        let footer = self.inner.footer();
+        let segment_map = footer.segment_map();
+        let idx = flat_segment_id as usize;
+        if idx < segment_map.len() {
```
jiaqizho: `if idx >= segment_map.len() ` just return as error.

cpp/src/format/bridge/rust/src/vortex_bridgeimpl.rs
jiaqizho: and check others.

cpp/src/format/bridge/rust/src/vortex_bridgeimpl.rs
```diff
@@ -1159,6 +1171,92 @@ impl VortexFile {
...
+    ///
+    /// V2 units use row-group granularity. V1 units use field/flat granularity.
+    pub(crate) fn field_layout_units(&self, field_name: &str) -> Result<Vec<u64>> {
+        let footer = self.inner.footer();
+        let root = footer.layout();
+
+        let units = if root.encoding_id().as_ref() == "milvus.v2_zoned_row_group" {
+            build_v2_row_group_units(&root, field_name)?
+        } else {
+            build_v1_flat_units(&root, field_name)?
```
jiaqizho, line 1233: in `row_group_zone_map_segment_ids`, it won't collect the row zone_map segment ids in V1. 

Is the current logic actually valid? Could you write a relevant gtest to verify this (using TEST_P, with the parameters being the layouts of vortex V1 and V2)?

cpp/src/format/vortex/vortex_planner.cpp
```diff
@@ -0,0 +1,547 @@
...
+  for (auto cell_id : cell_ids) {
+    const auto& meta = cell_metas_[cell_id];
+    if (kept_row_group_set.find(meta.physical_unit_index) != kept_row_group_set.end()) {
+      kept_cell_ids.emplace_back(cell_id);
+    }
+  }
+  return kept_cell_ids;
+}
+
+arrow::Result<VortexPlan> VortexPlanner::PlanForOffsets(const std::vector<int64_t>& offsets) const {
```
jiaqizho, line 441: current `offests` arg require sorted. u need check the `offests` is sorted. 

If input is `[150, 10]` Then the `VortexPlan` will output the plan with `[10, 150]`.

cpp/include/milvus-storage/format/vortex/vortex_planner.h
```diff
@@ -0,0 +1,108 @@
...
+  VortexReadSelection selection = VortexReadSelection::kRowRange;
```
jiaqizho: C-style: 
```
struct VortexReadPlan {
  VortexReadSelection selection = VortexReadSelection::kRowRange;
  union {
     struct {
         uint64_t row_start = 0;
         uint64_t row_end = 0;
        std::vector<VortexReadRange> row_ranges;
     } range_scan;
     struct {
        std::vector<int64_t> row_indices;
     } take;
  } plan;
  std::string predicate;
  bool apply_predicate = true;
}
```

C++ style
```
  struct VortexReadPlan {
    struct RangeScan {
      uint64_t row_start = 0;
      uint64_t row_end = 0;
      std::vector<VortexReadRange> row_ranges;
    };

    struct Take {
      std::vector<int64_t> row_indices;
    };

    std::variant<RangeScan, Take> scan = RangeScan{};
    std::string predicate;
    bool apply_predicate = true;
  };
```

DONOT pile everything together, wasting memory, and not all scenarios will be assigned values.

cpp/include/milvus-storage/format/vortex/vortex_planner.h
```diff
@@ -0,0 +1,108 @@
...
+
+enum class VortexReadSelection {
+  kRowRange,
+  kRowRanges,
+  kRowIndices,
+};
+
+struct VortexReadRange {
+  VortexPhysicalGranularity granularity = VortexPhysicalGranularity::kUnknown;
+  uint64_t physical_unit_index = 0;
```
jiaqizho: In fact, we did not use this parameter.

cpp/include/milvus-storage/format/vortex/vortex_planner.h
jiaqizho: consider these defined is fine!
```
  struct VortexPlan {
    std::vector<uint64_t> cell_ids;  // cache/load only
    VortexReadPlan read_plan;        // scan only
  };

  struct VortexReadPlan {
    struct RangeScan {
      std::vector<RowRange> ranges;
    };

    struct Take {
      std::vector<int64_t> row_indices;
    };

    std::variant<RangeScan, Take> op;
    std::string predicate;
    bool apply_predicate = true;
  };

  struct RowRange {
    uint64_t start;
    uint64_t end;
  };
```

cpp/include/milvus-storage/format/vortex/vortex_types.h
```diff
@@ -0,0 +1,70 @@
...
+  uint64_t row_offset = 0;
+  uint64_t row_count = 0;
+  std::vector<VortexFlatSegmentRange> flat_segments;
+};
+
+struct VortexFieldLayout {
+  VortexPhysicalGranularity granularity = VortexPhysicalGranularity::kUnknown;
+  std::vector<VortexFlatUnit> flats;
+  std::vector<VortexRowGroupUnit> row_groups;
+};
```
jiaqizho, line 93: consider these defined, and move the define in vortex_planner into this file.
```
struct BufferRange {
    uint64_t offset = 0;
    uint64_t length = 0;
  };

  struct RowRange {
    uint64_t start = 0;
    uint64_t end = 0;
  };

  struct VortexFieldLayout {
    enum class UnitKind : uint8_t {
      kFlat = 1,
      kRowGroup = 2,
    };

    struct SegmentRange {
      uint64_t segment_id = 0;
      BufferRange bytes;
    };

    struct PhysicalUnit {
      UnitKind kind = UnitKind::kFlat;
      uint64_t phy_id = 0;
      RowRange rows;
      std::vector<SegmentRange> segments;
    };

    std::vector<PhysicalUnit> phy_units;
  };

  struct VortexCellMeta {
    uint64_t cell_id = 0;
    VortexFieldLayout::UnitKind unit_kind = VortexFieldLayout::UnitKind::kFlat;
    uint64_t unit_id = 0;
    RowRange rows;
    std::vector<VortexFieldLayout::SegmentRange> segments;
    uint64_t memory_bytes = 0;
    uint64_t storage_bytes = 0;
  };
```

cpp/include/milvus-storage/format/vortex/vortex_footer_reader.h
```diff
@@ -0,0 +1,67 @@
+// Copyright 2023 Zilliz
```
jiaqizho, line 1: Here I think we should first make the API boundary explicit: which types are intended to be public API for Milvus, and which types are only internal implementation details. By “public” I mean logically visible/consumable by Milvus, not necessarily just physically located under cpp/include.

For example, VortexPlanner, VortexPlan, and VortexReadPlan may be part of the Milvus-facing API if Milvus is expected to plan reads and pass plans into the storage reader. But layout parsing types such as VortexFieldLayout, physical units, segment ranges, and footer-reader internals should probably stay internal. Otherwise we expose Vortex layout details too early and make future refactoring harder.

I also think the current change seems to be missing the special in-memory/sparse filesystem implementation. VortexRangeFileProvider defines a contract, and VortexFooterReader/VortexTranslater expect sparse_fs to provide GetVortexRangeFile(). However, I only see the interface, not an implementation of a filesystem that owns these range files and supports Resize, WriteAt, ReadAt, and Punch.

So the design needs to clarify who provides this filesystem. If Milvus is expected to provide it, that should be part of the public API contract. If storage owns it, then we likely need an internal VortexSparseFileSystem / in-memory range filesystem implementation in this PR.

jiaqizho:
image

codecov is required CI.

cpp/conanfile.py
```diff
@@ -114,6 +119,8 @@ def configure(self):
             self.options.rm_safe("fPIC")
         if self.settings.arch not in ("x86_64", "x86"):
             self.options["folly"].rm_safe("use_sse4_2")
+        if self.settings.os == "Linux":
```
jiaqizho: no need check linux.  Just 
```
self.options["opentelemetry-cpp"].shared = True
```

cpp/src/format/parquet/parquet_format_reader.cpp
```diff
@@ -87,9 +85,8 @@ arrow::Result<std::vector<RowGroupInfo>> ParquetFormatReader::create_row_group_i
   size_t offset = 0;
   for (int i = 0; i < metadata->num_row_groups(); ++i) {
     auto row_group_meta = metadata->RowGroup(i);
-    row_group_infos.emplace_back(RowGroupInfo{.start_offset = offset,
```
jiaqizho, line 90: why?

cpp/src/format/parquet/parquet_format_reader.cpp
```diff
@@ -60,9 +60,7 @@ static arrow::Result<std::vector<RowGroupInfo>> try_build_row_group_infos(
   size_t offset = 0;
   for (size_t i = 0; i < row_group_metadatas.size(); ++i) {
     auto row_group_metadata = row_group_metadatas.Get(i);
-    row_group_infos.emplace_back(RowGroupInfo{.start_offset = offset,
```
jiaqizho, line 63: why?don't do that, Implicit construction is prone to errors.

cpp/src/format/vortex/vortex_format_reader.cpp
```diff
@@ -92,15 +199,12 @@ static std::vector<RowGroupInfo> create_row_group_infos(uint64_t memory_usage_in
     return {};
   }
 
-  std::vector<RowGroupInfo> result(row_ranges.size());
+  std::vector<RowGroupInfo> result;
+  result.reserve(row_ranges.size());
   uint64_t last_offset = 0;
 
   for (size_t i = 0; i < row_ranges.size(); i++) {
-    result[i] = RowGroupInfo{
```
jiaqizho, line 99: don't do that

cpp/src/format/bridge/rust/src/lib.rs
```diff
@@ -290,7 +300,10 @@ pub mod vortex_ffi {
         fn with_filter_ref(self: &mut VortexScanBuilder, filter: &Expr);
         fn with_projection(self: &mut VortexScanBuilder, projection: Box<Expr>);
         fn with_projection_ref(self: &mut VortexScanBuilder, projection: &Expr);
+        fn with_row_indices_projection(self: &mut VortexScanBuilder, field_name: &str);
+        fn with_split_row_indices(self: &mut VortexScanBuilder, split_row_indices: bool);
```
jiaqizho, line 304:
```
  let merge_into_ranges = self
      .split_row_indices_override
      .unwrap_or(include_by_index.len() > self.num_natural_splits);
```

In fact, `split_row_indices` will not be forcibly overwritten the default logical. So after properties changed, won't be work

jiaqizho: let us make split_row_indices can overwrite the default logical is fine.

cpp/src/format/bridge/rust/src/vortex_bridgeimpl.rs
```diff
@@ -1236,19 +1485,52 @@ impl VortexScanBuilder {
...
+        });
+    }
+
     pub(crate) fn with_row_range(&mut self, row_range_start: u64, row_range_end: u64) {
         self.row_range = Some(row_range_start..row_range_end);
+        self.row_ranges = None;
+    }
+
+    pub(crate) fn with_row_ranges(&mut self, starts: &[u64], ends: &[u64]) {
+        assert_eq!(starts.len(), ends.len());
```
jiaqizho, line 1510: Does the caching layer not allow empty segments?

cpp/src/format/vortex/vortex_footer_reader.cpp
```diff
@@ -0,0 +1,510 @@
...
+    ParsedFieldUnit unit;
+    unit.physical_unit_index = raw_offsets[cursor++];
+    unit.row_offset = raw_offsets[cursor++];
+    unit.row_count = raw_offsets[cursor++];
+    const auto num_segments = raw_offsets[cursor++];
+
+    if (cursor + num_segments > raw_offsets.size()) {
+      return arrow::Status::Invalid(fmt::format("Truncated vortex segment list for {}", field_name));
+    }
+    if (num_segments == 0) {
```
jiaqizho: Do u have try open a vortex writer without write anything then closed? Please add this gtest to check this. 

In fact, the storage side **should not assume that** we do not have empty files to write to. This is permitted.

jiaqizho:
fill pr desc pls

PR #554, opened 2026-06-09
enhance: add macOS C++ CI and tighten cache writes

jiaqizho:
The cache changes are meant to stop every PR from creating its own Conan and Rust cache under `refs/pull/.../merge`, while still letting PR jobs reuse caches that already exist. The CI workflows now use explicit restore and save actions: restore runs for normal jobs, but save only runs on main when the primary key was not hit. The cache keys also include runner OS and architecture, so Linux and macOS builds do not share incompatible Conan or Rust artifacts. Release jobs are restore-only, so they can consume existing caches without publishing new ones.

This also adds a dedicated macOS C++ build job to cpp-ci.yml. It uses the normal cpp build path and runs make build, which keeps the mac build aligned with how the project is expected to build locally. The job is build-only and does not join the existing Linux unit-test artifact flow, so the current Linux build, upload, and test setup stays unchanged while macOS compatibility gets checked separately.


.github/workflows/cpp-ci.yml
```diff
@@ -75,6 +95,137 @@ jobs:
...
+              exit $?
+            fi
+            sleep 1
+          done
+          ps -o pid,ppid,stat,etime,command -p "$pid" > build/Release/test-diagnostics/milvus_test.ps.txt || true
+          sample "$pid" 10 -file build/Release/test-diagnostics/milvus_test.sample.txt || true
+          kill "$pid" || true
+          wait "$pid" || true
+          echo "milvus_test --gtest_list_tests did not finish within 300 seconds" >&2
+          exit 124
```
liliu-z: The new mac-build job's `List gtest tests` step ends in `exit 124` with no `continue-on-error`, so when `milvus_test --gtest_list_tests` hangs past the 300s watchdog the step fails and turns the whole job red. This already happened on run 27319194364 (sha 076d7d2a), where the step's conclusion was `failure`; the latest run is green only because ASAN was disabled. If mac-build is (or becomes) a merge-gating check, an intermittent hang will block merges even though nothing is broken. Add `continue-on-error: true` to this step (which also keeps `success()` true for the later cache-save steps) or keep mac-build non-required.

.github/workflows/cpp-ci.yml
```diff
@@ -75,6 +95,137 @@ jobs:
...
+      - name: Upload gtest diagnostics
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: mac-gtest-diagnostics
+          path: cpp/build/Release/test-diagnostics/
+          if-no-files-found: ignore
+
+      - name: Save rust build cache
+        if: github.ref == 'refs/heads/main' && steps.rust-cache.outputs.cache-hit != 'true'
```
liliu-z: The two mac-build cache-save steps (lines 213 and 222) use `if: github.ref == 'refs/heads/main' && steps.*-cache.outputs.cache-hit != 'true'` with no status function, so Actions ANDs in an implicit `success()`. They sit after the `List gtest tests` watchdog that `exit 124`s on a hang (line 202), so on a main run where the listing hangs `success()` is false and both saves are skipped — and the `continue-on-error: true` on the save steps doesn't help, since a skipped step isn't a failed one. The macOS os/arch-keyed Conan/Rust caches then stay unseeded, forcing from-source Conan rebuilds on macos-26 until a later main run both builds and passes the listing. Gate the saves with `always() && github.ref == 'refs/heads/main' && …`, move them above the listing step, or add `continue-on-error: true` to `List gtest tests`.

.github/workflows/cpp-ci.yml
jiaqizho: ack

.github/workflows/cpp-ci.yml
jiaqizho: ack

.github/workflows/cpp-mac-ci.yml
```diff
@@ -0,0 +1,148 @@
+name: cpp-macos-optional
+
+on:
+  push:
+    paths:
```
yhmo, line 5: Middle ---- This path filter means the new macOS workflow will not run for most C++ changes, nor for edits to this workflow file itself. The job builds the whole C++ tree, but it only triggers on `cpp/conanfile.py` and `.github/workflows/cpp-ci.yml`, so changes under `cpp/**`, CMake files, Rust bridge files, tests, or follow-up fixes to `.github/workflows/cpp-mac-ci.yml` can bypass the macOS check entirely. Please include the mac workflow file itself and the C++ build inputs this job is meant to validate, likely mirroring the Linux C++ workflow paths.

cpp/conanfile.py
```diff
@@ -125,6 +125,9 @@ def configure(self):
             self.options["arrow"].with_jemalloc = False
         else:
             self.options["arrow"].with_jemalloc = self.options.with_jemalloc
+        # macOS cannot build opentelemetry-cpp with shared=True.
+        if self.settings.os == "Linux":
+            self.options["opentelemetry-cpp"].shared = True
```
yhmo, line 130: Middle ---- This Linux-only override changes more than the macOS case described by the comment. Before this PR, `opentelemetry-cpp/*:shared` was always `True`; now the default is `False` and only Linux flips it back to `True`, so Windows and any other non-Linux target silently switch to static linkage too. If the intent is only to avoid the macOS shared build issue, please branch on `Macos` instead, or document/validate that static linkage is intentional for every non-Linux platform.

cpp/test/CMakeLists.txt
```diff
@@ -39,9 +38,6 @@ target_link_libraries(
   milvus_test PRIVATE milvus-storage GTest::gtest_main
 )
 
-include(GoogleTest)
-gtest_discover_tests(milvus_test DISCOVERY_TIMEOUT 120)
```
yhmo, line 43: Low ---- Removing `gtest_discover_tests(milvus_test ...)` means `milvus_test` is no longer registered with CTest. The Linux workflow still invokes the binary directly, so CI can remain green, but local/dev flows that use `ctest` will silently stop running these GoogleTest cases. If the goal is only to avoid build-time discovery on macOS, please keep test registration via a platform-gated `gtest_discover_tests` or an explicit `add_test` replacement.

.github/workflows/cpp-mac-ci.yml
jiaqizho: expected.

cpp/conanfile.py
jiaqizho: expected.

cpp/test/CMakeLists.txt
jiaqizho: no need this logical any more

Issue #555, opened 2026-06-10
Support AliveRecordStream For milvus-storage to support Milvus Predicate Delete


MrPresent-Han:
Support AliveRecordStream For milvus-storage to support Milvus Predicate Delete

xiaofan-luan:
what is aliveRecordSream, can you give some detailed explain and design?

PR #556, opened 2026-06-10
Support preidcate delete part5(#555)

MrPresent-Han:
related: https://github.com/milvus-io/milvus-storage/issues/555

### DIFF
--- a/cpp/include/milvus-storage/ffi_c.h
+++ b/cpp/include/milvus-storage/ffi_c.h
@@ -206,11 +206,22 @@ typedef struct LoonColumnGroups {
 } LoonColumnGroups;
 
 /**
- * @brief C structure representing delta logs
+ * @brief C delta log type values matching api::DeltaLogType.
+ */
+typedef enum LoonDeltaLogType {
+  LOON_DELTA_LOG_TYPE_PRIMARY_KEY = 0,
+  LOON_DELTA_LOG_TYPE_POSITIONAL = 1,
+  // Value 2 is retired: equality delete was never shipped; use LOON_DELTA_LOG_TYPE_PREDICATE.
+  LOON_DELTA_LOG_TYPE_PREDICATE = 3,
+} LoonDeltaLogType;
+
+/**
+ * @brief C structure representing delta logs.
  */
 typedef struct LoonDeltaLogs {
   const char** delta_log_paths;
   uint32_t* delta_log_num_entries;
+  uint32_t* delta_log_types;
   uint32_t num_delta_logs;
 } LoonDeltaLogs;
 
@@ -254,7 +265,7 @@ typedef struct LoonManifest {
   // Embedded ColumnGroups
   LoonColumnGroups column_groups;
 
-  // Delta logs (PRIMARY_KEY type only)
+  // Delta logs
   LoonDeltaLogs delta_logs;
 
   // Stats
@@ -542,6 +553,34 @@ FFI_EXPORT void loon_chunk_reader_destroy(LoonChunkReaderHandle reader);
 /// Opaque handle for Reader
 typedef uintptr_t LoonReaderHandle;
 
+/// Opaque handle for delete-aware alive reader
+typedef uintptr_t LoonMaskedReaderHandle;
+
+typedef struct LoonRowMask {
+  // Arrow-format bitmap view. Bit 1 means the corresponding row is alive.
+  // The caller must keep this view alive with private_data/release and must not
+  // read data after calling release or loon_masked_reader_destroy.
+  const uint8_t* data;
+  int64_t num_bits;
+  int64_t num_bytes;
+  int64_t bit_offset;
+  void (*release)(struct LoonRowMask* self);
+  void* private_data;
+} LoonRowMask;
+
+typedef struct LoonMaskedReadOptions {
+  // 0 means no timestamp upper bound.
+  uint64_t visible_until_ts;
+  // Primary-key field id. milvus-storage has no inherent primary-key concept, so
+  // the caller must supply it when the manifest contains PRIMARY_KEY delta logs.
+  // 0 means unset (no primary key); valid primary-key field ids are > 0.
+  int64_t pk_field_id;
+  // Row-timestamp field id (the per-row timestamp used for row_ts <= delete_ts).
+  // Required when the manifest contains any delete log. 0 means unset; valid
+  // row-timestamp field ids are > 0.
+  int64_t row_timestamp_field_id;
+} LoonMaskedReadOptions;
+
 /**
  * @brief Creates a new Reader for a milvus storage dataset
  *
@@ -559,6 +598,27 @@ FFI_EXPORT LoonFFIResult loon_reader_new(const LoonColumnGroups* column_groups,
                                          size_t num_columns,
                                          const LoonProperties* properties,
                                          LoonReaderHandle* out_handle);
+
+// @param key_retriever Optional decryption-key callback: maps a column's encryption
+//        metadata to its key. Pass NULL when the dataset/delta logs are not encrypted.
+//        Must be installed here because the alive read stream is built eagerly.
+FFI_EXPORT LoonFFIResult loon_masked_reader_new(const LoonManifest* manifest,
+                                                struct ArrowSchema* schema,
+                                                const char* const* needed_columns,
+                                                size_t num_columns,
+                                                const LoonProperties* properties,
+                                                LoonMaskedReaderHandle* out_handle,
+                                                const LoonMaskedReadOptions* options,
+                                                const char* (*key_retriever)(const char* metadata));
+
+FFI_EXPORT LoonFFIResult loon_masked_reader_next(LoonMaskedReaderHandle handle,
+                                                 struct ArrowArray** out_array,
+                                                 struct ArrowSchema** out_schema,
+                                                 LoonRowMask* out_row_mask);
+
+FFI_EXPORT void loon_row_mask_free(LoonRowMask* bitset);
+
+FFI_EXPORT void loon_masked_reader_destroy(LoonMaskedReaderHandle handle);
 
 /**
  * @brief Sets a key retriever callback for dynamic key retrieval
@@ -740,12 +800,13 @@ FFI_EXPORT LoonFFIResult loon_transaction_append_files(LoonTransactionHandle han
  * @param handle Transaction handle
  * @param path Relative path to the delta log file
  * @param num_entries Number of entries in the delta log
+ * @param delta_log_type LoonDeltaLogType value
  * @return result of FFI
- * @note Type is hardcoded to PRIMARY_KEY internally
  */
 FFI_EXPORT LoonFFIResult loon_transaction_add_delta_log(LoonTransactionHandle handle,
                                                         const char* path,
-                                                        int64_t num_entries);
+                                                        int64_t num_entries,
+                                                        uint32_t delta_log_type);
 
 /**
  * @brief Add a stat entry to the transaction updates
--- a/cpp/include/milvus-storage/ffi_c.h
+++ b/cpp/include/milvus-storage/ffi_c.h
@@ -273,6 +273,10 @@ typedef struct LoonManifest {
 
   // LOB files (TEXT column metadata)
   LoonLobFiles lob_files;
+
+  // Declared field semantics (0 = undeclared)
+  int64_t pk_field_id;
+  int64_t row_timestamp_field_id;
 } LoonManifest;
 
 /**
@@ -571,13 +575,14 @@ typedef struct LoonRowMask {
 typedef struct LoonMaskedReadOptions {
   // 0 means no timestamp upper bound.
   uint64_t visible_until_ts;
-  // Primary-key field id. milvus-storage has no inherent primary-key concept, so
-  // the caller must supply it when the manifest contains PRIMARY_KEY delta logs.
-  // 0 means unset (no primary key); valid primary-key field ids are > 0.
+  // Primary-key field id. Needed when the manifest contains PRIMARY_KEY delta
+  // logs; may be left 0 when the manifest itself declares it
+  // (loon_transaction_set_primary_key_field). A non-zero value must match a
+  // non-zero manifest declaration. Valid primary-key field ids are > 0.
   int64_t pk_field_id;
   // Row-timestamp field id (the per-row timestamp used for row_ts <= delete_ts).
-  // Required when the manifest contains any delete log. 0 means unset; valid
-  // row-timestamp field ids are > 0.
+  // Needed when the manifest contains any delete log; same manifest-or-option
+  // sourcing as pk_field_id. 0 means unset; valid field ids are > 0.
   int64_t row_timestamp_field_id;
 } LoonMaskedReadOptions;
 
@@ -619,6 +624,47 @@ FFI_EXPORT LoonFFIResult loon_masked_reader_next(LoonMaskedReaderHandle handle,
 FFI_EXPORT void loon_row_mask_free(LoonRowMask* bitset);
 
 FFI_EXPORT void loon_masked_reader_destroy(LoonMaskedReaderHandle handle);
+
+/// Opaque handle for the query-side delete-info stream
+typedef uintptr_t LoonDeletedTsReaderHandle;
+
+/**
+ * @brief Creates a delete-info stream over a dataset (see loon_deleted_ts_reader_next)
+ *
+ * Data columns are never returned; the columns needed to evaluate deletes
+ * (primary key, row timestamp, predicate fields) are read internally.
+ *
+ * @param manifest Dataset manifest
+ * @param schema Arrow schema with PARQUET:field_id metadata
+ * @param properties Read configuration properties
+ * @param options Same options as the masked reader (visible_until_ts, field ids)
+ * @param key_retriever Optional decryption-key callback, NULL when not encrypted
+ * @param out_handle Output (caller must call loon_deleted_ts_reader_destroy)
+ * @return result of FFI
+ */
+FFI_EXPORT LoonFFIResult loon_deleted_ts_reader_new(const LoonManifest* manifest,
+                                                    struct ArrowSchema* schema,
+                                                    const LoonProperties* properties,
+                                                    const LoonMaskedReadOptions* options,
+                                                    const char* (*key_retriever)(const char* metadata),
+                                                    LoonDeletedTsReaderHandle* out_handle);
+
+/**
+ * @brief Fetches the next chunk of per-row earliest delete timestamps
+ *
+ * Rows follow the dataset's global row order; the running position across
+ * chunks is the global row offset. Each value is the earliest delete timestamp
+ * applying to the row, 0 = alive. Chunk boundaries are unspecified; the
+ * concatenated length equals the dataset row count.
+ *
+ * @param handle Deleted-ts reader handle
+ * @param out_array Output int64 Arrow array; set to NULL at end of stream.
+ *        On success the caller owns it: call (*out_array)->release() then free(*out_array).
+ * @return result of FFI
+ */
+FFI_EXPORT LoonFFIResult loon_deleted_ts_reader_next(LoonDeletedTsReaderHandle handle, struct ArrowArray** out_array);
+
+FFI_EXPORT void loon_deleted_ts_reader_destroy(LoonDeletedTsReaderHandle handle);
 
 /**
  * @brief Sets a key retriever callback for dynamic key retrieval
@@ -807,6 +853,29 @@ FFI_EXPORT LoonFFIResult loon_transaction_add_delta_log(LoonTransactionHandle ha
                                                         const char* path,
                                                         int64_t num_entries,
                                                         uint32_t delta_log_type);
+
+/**
+ * @brief Declare which schema field is the dataset's primary key
+ *
+ * Records a dataset-level fact in the manifest; consumers decide how to use it.
+ * Set-once: committing a value that differs from an already-declared one fails.
+ *
+ * @param handle Transaction handle
+ * @param field_id Primary-key field id (must be > 0)
+ * @return result of FFI
+ */
+FFI_EXPORT LoonFFIResult loon_transaction_set_primary_key_field(LoonTransactionHandle handle, int64_t field_id);
+
+/**
+ * @brief Declare which schema field carries the per-row timestamp
+ *
+ * Same set-once semantics as loon_transaction_set_primary_key_field.
+ *
+ * @param handle Transaction handle
+ * @param field_id Row-timestamp field id (must be > 0)
+ * @return result of FFI
+ */
+FFI_EXPORT LoonFFIResult loon_transaction_set_row_timestamp_field(LoonTransactionHandle handle, int64_t field_id);
 
 /**
  * @brief Add a stat entry to the transaction updates
--- a/cpp/include/milvus-storage/manifest.h
+++ b/cpp/include/milvus-storage/manifest.h
@@ -69,7 +69,8 @@ struct LobFileInfo {
 enum class DeltaLogType {
   PRIMARY_KEY = 0,  // Primary key delete (default)
   POSITIONAL = 1,   // Positional delete
-  EQUALITY = 2,     // Equality delete
+  // Value 2 is retired: equality delete was never shipped; it is expressed as a predicate delete.
+  PREDICATE = 3,  // Predicate delete event log
 };
 
 /**
--- a/cpp/include/milvus-storage/manifest.h
+++ b/cpp/include/milvus-storage/manifest.h
@@ -37,7 +37,8 @@ namespace milvus_storage::api {
 // - Version 3: Changed stats from map<string, vector<string>> to map<string, Statistics>
 // - Version 4: Changed ColumnGroupFile fields (metadata) to properties map
 // - Version 5: Added lob_files field for LOB (Large Object) file metadata
-constexpr int32_t MANIFEST_VERSION = 5;
+// - Version 6: Added pk_field_id / row_timestamp_field_id declared-field-semantics fields
+constexpr int32_t MANIFEST_VERSION = 6;
 
 /**
  * @brief Metadata for a single LOB (Large Object) file
@@ -192,6 +193,23 @@ class Manifest final {
   }
 
   /**
+   * @brief Declared primary-key field id (0 = undeclared)
+   *
+   * Dataset-level fact about which schema field is the primary key. The writer
+   * declares it once; consumers (e.g. delete evaluation) decide how to use it.
+   */
+  [[nodiscard]] int64_t& pkFieldId() { return pk_field_id_; }
+  [[nodiscard]] const int64_t& pkFieldId() const { return pk_field_id_; }
+
+  /**
+   * @brief Declared row-timestamp field id (0 = undeclared)
+   *
+   * Dataset-level fact about which schema field carries the per-row timestamp.
+   */
+  [[nodiscard]] int64_t& rowTimestampFieldId() { return row_timestamp_field_id_; }
+  [[nodiscard]] const int64_t& rowTimestampFieldId() const { return row_timestamp_field_id_; }
+
+  /**
    * @brief Get the manifest format version
    */
   [[nodiscard]] int32_t version() const { return version_; }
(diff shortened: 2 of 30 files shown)


cpp/src/ffi/reader_c.cpp
```diff
@@ -570,6 +571,183 @@ LoonFFIResult loon_take(LoonReaderHandle reader,
...
+      RETURN_ERROR(LOON_ARROW_ERROR, schema_result.status().ToString());
+    }
+
+    std::shared_ptr<Manifest> cpp_manifest;
+    auto import_st = milvus_storage::manifest_import(manifest, &cpp_manifest);
+    if (!import_st.ok()) {
+      RETURN_ERROR(LOON_LOGICAL_ERROR, import_st.ToString());
+    }
+
+    auto reader = Reader::create(cpp_manifest, schema_result.ValueOrDie(),
```
liliu-z, line 650: The FFI alive-reader path (loon_alive_reader_new) creates the Reader and eagerly builds the masked stream without ever supplying a key retriever: the manifest constructor sets key_retriever_callback_ to null (reader.cpp:796), the function has no retriever parameter, and unlike the normal reader (which exposes loon_reader_set_keyretriever) the returned handle is an AliveReaderFFIState rather than a Reader, so a retriever can never be set. Reading an encrypted dataset through the FFI alive reader therefore fails. The C++ API supports encryption via set_keyretriever before get_masked_record_batch_reader; thread the retriever through the FFI alive-reader path the same way.

cpp/src/delete_evaluator.cpp
```diff
@@ -0,0 +1,1073 @@
...
+      }
+      if (delta_log.num_entries == 0) {
+        continue;
+      }
+
+      ARROW_ASSIGN_OR_RAISE(auto format_reader, OpenDeltaLogReader(delta_log, properties_, key_retriever_));
+      auto file_schema = format_reader->get_schema();
+      ARROW_ASSIGN_OR_RAISE(auto batch_reader, format_reader->read_with_range(0, delta_log.num_entries));
+      auto reader = std::make_shared<OwningRecordBatchReader>(std::move(format_reader), std::move(batch_reader));
+      if (ShouldReadAsPredicateDeltaLog(file_schema)) {
```
liliu-z: LoadDeltaLogs dispatches each delta log solely by inspecting the file schema (ShouldReadAsPredicateDeltaLog); the persisted delta_log.type is never consulted, so anything not predicate-shaped — including a POSITIONAL or EQUALITY delta log — silently falls into the PRIMARY_KEY branch, where LoadPrimaryKeyDeltaLog's positional column fallback (:812-814) then consumes its columns as pk/ts. parse_delta_log_type (manifest_c.cpp:242) accepts these types and Avro round-trips them (manifest.cpp:136), so an external producer can create one and have wrong delete semantics applied. No in-tree producer emits POSITIONAL/EQUALITY today, but the evaluator should reject unsupported types instead of coercing them to PRIMARY_KEY.

cpp/include/milvus-storage/ffi_c.h
```diff
@@ -739,12 +781,13 @@ FFI_EXPORT LoonFFIResult loon_transaction_append_files(LoonTransactionHandle han
...
  * @param handle Transaction handle
  * @param path Relative path to the delta log file
+ * @param delta_log_type LoonDeltaLogType value
  * @param num_entries Number of entries in the delta log
  * @return result of FFI
- * @note Type is hardcoded to PRIMARY_KEY internally
  */
 FFI_EXPORT LoonFFIResult loon_transaction_add_delta_log(LoonTransactionHandle handle,
                                                         const char* path,
+                                                        uint32_t delta_log_type,
```
liliu-z: delta_log_type was inserted as the 3rd parameter of loon_transaction_add_delta_log (before num_entries) rather than appended, changing the ABI while the symbol name is unchanged. A caller compiled against the old 3-arg signature and linked without recompiling would pass num_entries where the new function reads delta_log_type, and read num_entries from an unset register. All in-tree consumers (including the Python binding, updated in this PR to pass `path, 0, num_entries`) rebuild in lockstep, so realized risk is low; appending the parameter (or versioning the symbol) would remove the footgun.

cpp/src/ffi/reader_c.cpp
```diff
@@ -570,6 +571,183 @@ LoonFFIResult loon_take(LoonReaderHandle reader,
...
+                                 convert_needed_columns(needed_columns, num_columns), std::move(properties_map));
+    AliveReadOptions options;
```
liliu-z: loon_alive_reader_new default-constructs AliveReadOptions and exposes none of its fields, so FFI consumers cannot set visible_until_ts (nor batch_size/parallelism). With visible_until_ts unset the evaluator applies every delete regardless of timestamp (delete_evaluator.cpp:844 only skips deletes when the option has a value), so snapshot/time-bounded alive reads — supported by the C++ API and covered by C++ tests — cannot be expressed through the FFI. Add the options to the FFI entry point and forward them to get_masked_record_batch_reader.

cpp/src/delete_evaluator.cpp
```diff
@@ -0,0 +1,1073 @@
...
+  }
+  return PredicateBool::kFalse;
+}
+
+static arrow::Result<PredicateBool> EvalPredicateNode(const std::shared_ptr<PredicateNode>& node,
+                                                      const std::shared_ptr<arrow::RecordBatch>& batch,
+                                                      int64_t row) {
+  switch (node->kind) {
+    case PredicateNodeKind::kCompare: {
+      ARROW_ASSIGN_OR_RAISE(auto index, FindFieldIndexByFieldId(batch->schema(), node->field_id));
```
liliu-z: For every row, EvaluatePredicateDeletes (loop at :988) calls EvalPredicateNode, which resolves each referenced field via FindFieldIndexByFieldId at :590 — a linear scan over all schema fields that calls GetFieldId (which std::stoll-parses the field's PARQUET:field_id metadata or name) on each field. The field-id-to-index mapping is constant for the whole batch, so this repeats O(rows × predicate-nodes × fields) string conversions on the hot path and degrades sharply on large batches and wide schemas. Resolve the mapping once before the row loop and index into it.

cpp/src/ffi/reader_c.cpp
MrPresent-Han: The DataNode-side segment delta log path currently does not go through the CMEK encryption path, so delta logs are not encrypted today. I do not think we should implement key retriever support for this path ahead of actual encrypted delta-log support.

cpp/src/delete_evaluator.cpp
MrPresent-Han: Updated the design accordingly: the alive reader now dispatches by the manifest delta_log.type instead of using the delta-log file schema as the routing source. PREDICATE delta logs still assert their file-level predicate metadata/schema during loading, and POSITIONAL/EQUALITY delta logs now fail fast as unsupported instead of falling back to PRIMARY_KEY.

cpp/include/milvus-storage/ffi_c.h
MrPresent-Han: Updated. I moved the new delta_log_type argument to the end of loon_transaction_add_delta_log, so the existing parameter order stays (handle, path, num_entries) and the new type parameter is appended instead of inserted in the middle.

cpp/src/ffi/reader_c.cpp
MrPresent-Han: changed

cpp/src/delete_evaluator.cpp
MrPresent-Han: changed

cpp/include/milvus-storage/ffi_c.h
```diff
@@ -206,11 +206,22 @@ typedef struct LoonColumnGroups {
...
+  LOON_DELTA_LOG_TYPE_PREDICATE = 3,
+} LoonDeltaLogType;
+
+/**
+ * @brief C structure representing delta logs.
  */
 typedef struct LoonDeltaLogs {
   const char** delta_log_paths;
   uint32_t* delta_log_num_entries;
+  uint32_t* delta_log_types;
```
liliu-z, line 224: This PR adds a third pointer field `delta_log_types` to the C `LoonDeltaLogs` struct (and `manifest_export` now populates it), but the hand-written cffi cdef mirror in `python/milvus_storage/_ffi.py` (the `LoonDeltaLogs` typedef near line 174) was not updated to match. In cffi ABI mode the cdef defines the struct layout, so `Manifest._from_c` reads `num_delta_logs` — and the `stats` member that follows `delta_logs` inside `LoonManifest` — at shifted offsets, producing a garbage count and an out-of-bounds read for any manifest that carries a delta log (the primary path for this delete feature). Fix by adding `uint32_t* delta_log_types;` before `num_delta_logs` in the Python cdef.

cpp/src/common/sql_predicate_arrow.cpp
```diff
@@ -0,0 +1,492 @@
...
+      case '(':
+        ++pos_;
+        return Token{TokenKind::kLParen};
+      case ')':
+        ++pos_;
+        return Token{TokenKind::kRParen};
+      case ',':
+        ++pos_;
+        return Token{TokenKind::kComma};
+      default:
```
liliu-z, line 272: The tokenizer rejects the `-` character, so negative numeric literals never parse and predicates such as `col < -5` or `col in (-1, -2)` fail. Instead of filtering, such a predicate aborts the entire segment read and makes the segment unreadable — though this only bites if Milvus's serializer actually emits negative literals, which isn't verified from this repo. Handle a leading `-` as part of a numeric literal during tokenization.

cpp/src/delete_evaluator.cpp
```diff
@@ -0,0 +1,612 @@
...
+        arrow::bit_util::ClearBit(mask, i);
+      }
+    }
+    return arrow::Status::OK();
+  }
+
+  arrow::Status EvaluatePredicateDeletes(const std::shared_ptr<arrow::RecordBatch>& batch, uint8_t* mask) const {
+    ARROW_ASSIGN_OR_RAISE(auto exec_batch, cp::MakeExecBatch(*batch->schema(), arrow::Datum(batch)));
+    for (const auto& expression : predicate_delete_expressions_) {
+      ARROW_ASSIGN_OR_RAISE(auto bound_expr, expression.Bind(*batch->schema()));
```
liliu-z: Every predicate delete expression is re-bound on every batch via `expression.Bind(*batch->schema())`, even though all batches from the reader share the same projected schema, so the bind result is identical each time. Bind once (lazily on the first batch, or against the projected schema at construction) and cache the bound expressions. Note the load-time `Bind(*schema_)` at line 421 binds against the full (unprojected) schema and is deliberately discarded for validation, so it cannot be reused directly for the projected batch schema — the redundant work to eliminate is the per-batch rebind, not simply 'reusing the load-time result'.

cpp/src/reader.cpp
```diff
@@ -758,6 +852,57 @@ class ReaderImpl : public Reader {
...
+    }
+    if (options.batch_size < 0) {
+      return arrow::Status::Invalid("AliveReadOptions.batch_size must be >= 0");
+    }
+    if (options.parallelism == 0) {
+      return arrow::Status::Invalid("AliveReadOptions.parallelism must be > 0");
+    }
+
+    ARROW_ASSIGN_OR_RAISE(auto evaluator,
+                          CreateDeleteEvaluator(manifest_, schema_, properties_, options, key_retriever_callback_));
```
liliu-z, line 914: `loon_alive_reader_new` (reader_c.cpp:644) builds the read stream eagerly while `ReaderImpl::key_retriever_callback_` is still null, and `LoonAliveReaderHandle` exposes no setter to install one, so that null callback is forwarded both to `CreateDeleteEvaluator` here and to the base reader at reader.cpp:900. No code path on the alive reader can therefore supply a decryption key, so any encrypted dataset or encrypted delta log fails to decrypt and reads break. Install the callback before constructing the stream, or add a setter on the handle and defer stream creation until it is provided.

cpp/include/milvus-storage/ffi_c.h
MrPresent-Han: Good catch — confirmed and fixed. Added `uint32_t* delta_log_types;` before `num_delta_logs` in the `LoonDeltaLogs` cdef in `python/milvus_storage/_ffi.py`, so the ABI-mode struct layout now matches the C header and `num_delta_logs` (and the following `stats` member) are read at the correct offsets. (6cb7582)

cpp/src/common/sql_predicate_arrow.cpp
MrPresent-Han: Fixed. The tokenizer now treats a `-` immediately followed by a digit as a negative numeric literal (e.g. `col < -5`, `col in (-1, -2)`); a `-` not followed by a digit (unary minus like `-a`) still falls through to ReadOperator and is rejected, so unary minus stays unsupported. Added `SupportsNegativeNumericLiterals` test coverage. (6cb7582)

cpp/src/delete_evaluator.cpp
MrPresent-Han: Fixed. Predicate expressions are now bound once against the first batch's schema and cached in `bound_predicate_expressions_`, then reused for subsequent batches. As you noted, the load-time `Bind(*schema_)` is validation-only against the full schema and is left untouched. (6cb7582)

cpp/src/reader.cpp
```diff
@@ -100,9 +177,10 @@ class PackedRecordBatchReader final : public arrow::RecordBatchReader {
     // Initialize config from properties
     memory_usage_limit_ =
         milvus_storage::api::GetValueNoError<int64_t>(properties_, PROPERTY_READER_RECORD_BATCH_MAX_SIZE);
-    number_of_row_limit_ =
-        milvus_storage::api::GetValueNoError<int64_t>(properties_, PROPERTY_READER_RECORD_BATCH_MAX_ROWS);
-    parallelism_ = milvus_storage::ThreadPoolHolder::GetParallelism();
+    number_of_row_limit_ = batch_size_ > 0 ? batch_size_
+                                           : milvus_storage::api::GetValueNoError<int64_t>(
+                                                 properties_, PROPERTY_READER_RECORD_BATCH_MAX_ROWS);
+    parallelism_ = parallelism_hint_ > 0 ? parallelism_hint_ : milvus_storage::ThreadPoolHolder::GetParallelism();
```
jiaqizho, line 235: why need pass `batch_size_` and `parallelism` to override old logical?

cpp/include/milvus-storage/ffi_c.h
```diff
@@ -560,6 +599,23 @@ FFI_EXPORT LoonFFIResult loon_reader_new(const LoonColumnGroups* column_groups,
                                          const LoonProperties* properties,
                                          LoonReaderHandle* out_handle);
 
+FFI_EXPORT LoonFFIResult loon_alive_reader_new(const LoonManifest* manifest,
```
jiaqizho: Current reader is streaming reader.

I'm wondering if `get_chunks` and `take` here should also include corresponding handling? There might not be a need for `take` right now( But lance already support), but perhaps there is for `get_chunks`?

jiaqizho: Also there are a problem: If current column is TEXT field, then the LOB would be apply.

cpp/src/reader.cpp
MrPresent-Han: Good point — these two were redundant, so I removed `batch_size` and `parallelism` from both `AliveReadOptions` and the FFI `LoonAliveReadOptions`. The alive read now derives batch size from the existing per-reader property `reader.record_batch_max_rows` and parallelism from the global thread pool (`ThreadPoolHolder`), i.e. the `> 0 ? override : old_source` logic here falls through to the old sources. No behavior lost; one less duplicated knob to keep in sync across the C ABI. (Note: this drops two fields from `LoonAliveReadOptions`, so the Milvus-side caller needs to update the struct.)

cpp/include/milvus-storage/ffi_c.h
MrPresent-Han: Right — the alive / delete-mask path is intentionally streaming-only in this PR: the keep-mask is computed and applied per batch inside `loon_alive_reader_next`, so the streaming alive reader returns each batch together with its alive bitset. Masked `get_chunks`/`take` (applying the delete mask to random-access chunk/row reads) is a reasonable follow-up but out of scope here — `take` isn't needed for the current streaming consumption, and if a chunk-level masked read turns out to be needed we can add it in a separate PR. Keeping this PR focused on the streaming alive read.

cpp/include/milvus-storage/ffi_c.h
```diff
@@ -542,6 +553,34 @@ FFI_EXPORT void loon_chunk_reader_destroy(LoonChunkReaderHandle reader);
 /// Opaque handle for Reader
 typedef uintptr_t LoonReaderHandle;
 
+/// Opaque handle for delete-aware alive reader
+typedef uintptr_t LoonAliveReaderHandle;
```
tedxu: Two problems with the `Alive*` terminology across the API/FFI (`AliveReadOptions`, `LoonAliveReaderHandle`, `LoonAliveReadOptions`, `LoonAliveBitset`, `loon_alive_reader_*`, `AliveReaderFFIState`, `AliveBitsetPrivateData`):

1. "Alive" implies deleted rows are dead, but that's not what this reader models. Visibility is bounded by `visible_until_ts`, so a row deleted at ts=100 is still "alive" for a snapshot at ts=50. The name lies about the semantics.
2. "Stream" is redundant. Every `RecordBatchReader` in the codebase is already streaming; we don't spell that out anywhere else (`get_record_batch_reader` doesn't need to say "stream").

The internal C++ type is already `MaskedRecordBatchReader` and that reads well. Suggest aligning the FFI/options names with it: `LoonMaskedReaderHandle`, `LoonMaskedReadOptions`, `MaskedReadOptions`, `loon_masked_reader_new/next/destroy`, `LoonRowMask` (instead of `LoonAliveBitset`). If `Masked` isn't preferred, `DeleteAware`/`Filtered` are also more accurate than `Alive`.

cpp/include/milvus-storage/manifest.h
```diff
@@ -69,7 +69,8 @@ struct LobFileInfo {
 enum class DeltaLogType {
   PRIMARY_KEY = 0,  // Primary key delete (default)
   POSITIONAL = 1,   // Positional delete
-  EQUALITY = 2,     // Equality delete
+  EQUALITY = 2,     // Deprecated: equality delete is represented as predicate delete in this version.
```
tedxu: `EQUALITY = 2` is marked deprecated, but we still keep it in the enum, parse it in the FFI (`manifest_c.cpp:242`), round-trip it through Avro (`manifest.cpp:136`), and then explicitly reject it in the reader (`delete_evaluator.cpp:260-262`). That's four places carrying an unshipped concept just to say "don't use this."

Since equality delete has never been published, we can just drop the variant. If we want to distinguish "predicate that's just an equality" from "arbitrary SQL predicate" as a downstream optimization hint, rename it to something like `PREDICATE_EQUALITY` and treat it as a subtype — but don't leave a rejected-only enum value on the public API.

cpp/src/delete_evaluator.cpp
```diff
@@ -0,0 +1,650 @@
...
+#include "milvus-storage/common/macro.h"
+#include "milvus-storage/common/metadata.h"
+#include "milvus-storage/filesystem/fs.h"
+#include "milvus-storage/format/format_reader.h"
+
+namespace milvus_storage::api {
+
+namespace cp = arrow::compute;
+
+static constexpr int64_t kMilvusTimestampFieldId = 1;
```
tedxu: The comment at :157-158 already states "milvus-storage has no inherent primary-key concept" — that's why `AliveReadOptions.pk_field_id` exists and is validated at :296-300. But we then hardcode `kMilvusTimestampFieldId = 1` and require every schema to have an int64 field at field-id 1 named as the row timestamp (see also `ResolveTimestampFieldNameByFieldId` and the use at :197, :511). That's the same class of Milvus-specific coupling we said we didn't want.

Two ways out:
- Symmetrical fix: add `AliveReadOptions.row_timestamp_field_id` and require the caller to declare it, matching how PK is handled.
- Or accept that row-timestamp *is* a storage-model primitive (unlike PK): document it explicitly on the reader header, drop the `kMilvusTimestamp` prefix, and stop apologizing for the PK coupling in the surrounding comments. Pick one story.

As written the code says "we're generic" but behaves "we're Milvus-shaped" — the worst spot to be in.

cpp/src/delete_evaluator.cpp
```diff
@@ -0,0 +1,650 @@
...
+static constexpr const char* kDeltaPkColumnName = "pk";
```
tedxu: `LoadPrimaryKeyDeltaLog` at :336-344 reads by these names first, then silently falls back to positional column 0/1 if the names don't match:

```cpp
std::shared_ptr<arrow::Array> pk_array = batch->GetColumnByName(kDeltaPkColumnName);
std::shared_ptr<arrow::Array> ts_column = batch->GetColumnByName(kDeltaTsColumnName);
if (!pk_array || !ts_column) {
  pk_array = batch->column(0);
  ts_column = batch->column(1);
}
```

Two problems:
1. The strings `"pk"`/`"ts"` are the delta writer's implicit contract but they aren't declared anywhere. A producer that uses different names still "works" because of the positional fallback, so the name lookup is decorative.
2. The fallback masks bugs: if a future writer swaps column order (column 0 = ts, column 1 = pk), the delete semantics silently invert — no error surfaces.

`LoadPredicateDeltaLog` at :409-414 already does the honest thing: pure positional (columns 0/1), enforced by type checks in `ValidatePredicateDeltaBatchSchema`, constants documented at :48-52. Do the same for `PRIMARY_KEY`: pick position, enforce with type checks, drop the name lookup, delete these two constants.

cpp/src/delete_evaluator.cpp
```diff
@@ -0,0 +1,650 @@
...
+                                         int pk_index,
+                                         const arrow::Int64Array& ts_array,
+                                         uint8_t* mask,
+                                         int64_t& alive_count) const {
+    auto pk_array = std::static_pointer_cast<arrow::StringArray>(batch->column(pk_index));
+    for (int64_t i = 0; i < batch->num_rows(); ++i) {
+      if (pk_array->IsNull(i) || ts_array.IsNull(i)) {
+        continue;
+      }
+      auto it = string_pk_delete_ts_.find(pk_array->GetString(i));
```
tedxu: `pk_array->GetString(i)` materializes a `std::string` for every row just to use as an `unordered_map` key. For an 8K-row batch with string PKs that's 8K allocations per batch we immediately throw away, on the delete hot path.

A heterogeneous-lookup map (C++20 transparent hash/equal so we can `find(std::string_view)`) or a byte-hash map with a `string_view` overload would let us skip the copy. Not urgent for correctness, but this is the kind of thing that will show up in a flamegraph the first time someone benches predicate + PK delete on string-PK collections.

cpp/src/delete_evaluator.cpp
```diff
@@ -0,0 +1,650 @@
...
+      ARROW_ASSIGN_OR_RAISE(auto delete_ts, ValidateDeleteTimestamp(delete_ts_array.Value(i), path));
+      if (options_.visible_until_ts.has_value() && delete_ts > *options_.visible_until_ts) {
+        continue;
+      }
+      ARROW_ASSIGN_OR_RAISE(auto predicate_sql, GetPredicateSqlValue(predicate_sql_array, i, path));
+      ARROW_ASSIGN_OR_RAISE(auto delete_expr,
+                            BuildPredicateDeleteExpression(predicate_sql, delete_ts, schema_, timestamp_field_name));
+      ARROW_ASSIGN_OR_RAISE(auto bound_expr, delete_expr.Bind(*schema_));
+      (void)bound_expr;
+      predicate_delete_expressions_.push_back(std::move(delete_expr));
```
tedxu, line 614: Two related observations on how PREDICATE deltas accumulate here:

1. **No deduplication at load time.** Each row in a PREDICATE delta log becomes its own `Expression` pushed into `predicate_delete_expressions_`, and `EvaluatePredicateDeletes` (:588) loops through every one on every batch. A segment with N delete events × M batches ⇒ N·M `ExecuteScalarExpression` calls, even when many delete rows share `(predicate_sql, delete_ts)` or share `predicate_sql` with different `delete_ts` (mergeable as `predicate AND ts <= max(delete_ts)`). Not blocking for this PR — worth a comment marking the O(delete_events × batches) scaling as intentional near-term.

2. **Non-atomic state on partial failure.** If parsing/validation of a later row fails, `predicate_delete_expressions_` already holds the predicates from earlier iterations (and earlier delta files in the same load). Currently harmless because `Create` throws the evaluator away on error, but it's a footgun the moment we expose reload or incremental delta loading.

tedxu (commented):
A few naming / API-shape and hot-path comments. Individual threads inline.

cpp/include/milvus-storage/ffi_c.h
MrPresent-Han: Done — renamed the entire API/FFI surface to `Masked`, matching the internal `MaskedRecordBatchReader`: `MaskedReadOptions`, `LoonMaskedReadOptions`, `LoonMaskedReaderHandle`, `loon_masked_reader_new/next/destroy`, and `LoonRowMask` (replacing `LoonAliveBitset`). The internal FFI state / private-data types (`MaskedReaderFFIState`, `RowMaskPrivateData`) were renamed to match. No `Alive*` or redundant `*Stream` names remain on the public API/FFI.

cpp/include/milvus-storage/manifest.h
MrPresent-Han: Dropped the variant entirely, per "never published." Removed `EQUALITY` from `DeltaLogType` and `LoonDeltaLogType`, from `parse_delta_log_type` (`manifest_c.cpp`), and from both reader switches in `delete_evaluator.cpp`. `PREDICATE` keeps its wire value `3` and value `2` is left as a documented retired slot, so there's no on-disk renumbering.

Added regression coverage: `MaskedReaderRejectsUnsupportedDeltaLogTypes` now also feeds the retired value `2` and asserts it is rejected as an unknown type rather than silently coerced into the `PRIMARY_KEY` branch.

cpp/src/delete_evaluator.cpp
MrPresent-Han: Took the symmetrical option. Added `MaskedReadOptions.row_timestamp_field_id` (mirrored as `LoonMaskedReadOptions.row_timestamp_field_id`); the caller must declare it, exactly like `pk_field_id`. Removed the hardcoded `kMilvusTimestampFieldId` — a new `RowTimestampFieldId()` helper returns the caller-declared id and errors when deletes are present but it is unset. The header comment now documents both `pk_field_id` and `row_timestamp_field_id` as caller-declared, so the "storage has no inherent PK/timestamp concept" story is consistent with the behavior end to end.

cpp/src/delete_evaluator.cpp
MrPresent-Han: Done — `LoadPrimaryKeyDeltaLog` is now pure positional (column 0 = primary key, column 1 = delete_timestamp) with a type check on the timestamp column, mirroring `LoadPredicateDeltaLog`. Dropped the `GetColumnByName` lookup and deleted the `kDeltaPkColumnName` / `kDeltaTsColumnName` constants; the position contract is documented on the `kPrimaryKeyDelta*Column` constants. Added `PrimaryKeyDeltaLogReadsByPhysicalColumnOrder` (a delta whose column names don't match) to lock the positional behavior so a future writer swapping order can no longer silently invert delete semantics.

cpp/src/delete_evaluator.cpp
MrPresent-Han: Done — `string_pk_delete_ts_` now uses a transparent-hash map (`TransparentStringHash` with `is_transparent`, plus `std::equal_to<>`), and the per-row lookup uses `pk_array->GetView(i)` (a `std::string_view`). No `std::string` is materialized per row on the delete path anymore.

cpp/src/delete_evaluator.cpp
MrPresent-Han: Addressed both points:

1. **Deduplication at load time.** Predicate deltas are now deduplicated by `predicate_sql`, keeping `max(delete_ts)` — the mergeable `predicate AND row_ts <= max(delete_ts)` form you described. `predicate_max_ts_` accumulates during load, and `BuildPredicateExpressions` emits one bound expression per *unique* predicate, so `EvaluatePredicateDeletes` loops over unique predicates instead of raw delete rows (bounding it by distinct predicates × batches, not delete_events × batches). Covered by `SamePredicateMultipleTimestampsUseMaxTimestamp`.

2. **Atomicity on partial failure.** A predicate that fails to parse/bind is treated as a correctness error and aborts `Create` — the evaluator is discarded and no partial delete set is ever applied. We deliberately do *not* skip an unparseable delete row (skipping would under-delete and surface deleted data), so the state is all-or-nothing at construction.

MrPresent-Han:
1. store pk/ts field directly in the manifest to avoid passing into milvus-storage everytime
2. confirm whether the returned alive bitset can be used in the query-node DeleteRecord, especially in the TieredStrorage case where only a separated part is loaded, may be we need an api to only return alivebitset like as using neededColumns={}
3. if one Record is deleted completely, we should return a total-null record batch

MrPresent-Han:
Closing per cleanup — non-core-milvus PR.

PR #559, opened 2026-06-15
enhance: add optional AWS S3 CRT build support

jiaqizho:
We need a way to build milvus-storage with AWS S3 CRT support without changing the default build path. The reason this is guarded by WITH_CRT is that bringing in AWS CRT still feels more like an experimental feature at this stage, so it should be easy to opt into without making it part of every normal build yet.

This change adds a with_crt Conan option and threads it through the Makefile and CMake as `WITH_CRT`. When it is enabled, Conan turns on `aws-sdk-cpp/*:s3-crt`, CMake defines `WITH_CRT`, and the build links against `AWS::aws-sdk-cpp-s3-crt` so the CRT headers and symbols are available to the storage library build.

### DIFF
--- a/cpp/test/filesystem/s3_crt_build_test.cpp
+++ b/cpp/test/filesystem/s3_crt_build_test.cpp
@@ -33,9 +33,6 @@ TEST(S3CrtBuildSupportTest, HeadersAndStaticClientSymbolsAreAvailable) {
 
   ASSERT_STATUS_OK(EnsureS3Initialized());
 
-  Aws::S3Crt::S3CrtClientConfiguration config;
-  EXPECT_FALSE(config.region.empty());
-
   const char* service_name = Aws::S3Crt::S3CrtClient::GetServiceName();
   ASSERT_NE(service_name, nullptr);
   EXPECT_FALSE(std::string(service_name).empty());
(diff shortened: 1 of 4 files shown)


cpp/test/filesystem/s3_crt_build_test.cpp
```diff
@@ -0,0 +1,46 @@
...
+namespace milvus_storage::test {
+
+TEST(S3CrtBuildSupportTest, HeadersAndStaticClientSymbolsAreAvailable) {
+  static_assert(std::is_class_v<Aws::S3Crt::S3CrtClient>);
+  static_assert(std::is_default_constructible_v<Aws::S3Crt::S3CrtClientConfiguration>);
+
+  ASSERT_STATUS_OK(EnsureS3Initialized());
+
+  Aws::S3Crt::S3CrtClientConfiguration config;
+  EXPECT_FALSE(config.region.empty());
```
liliu-z: The test default-constructs an `Aws::S3Crt::S3CrtClientConfiguration` (line 36), not an `Aws::Client::ClientConfiguration`, but its `region` is resolved the same way: from `AWS_REGION`/`AWS_DEFAULT_REGION` or an AWS profile, falling back to empty when none are set (aws-sdk-cpp 1.11.692). So `EXPECT_FALSE(config.region.empty())` fails on a clean runner with no AWS environment and effectively tests the host's AWS config rather than CRT build support; production code always sets the region explicitly (s3_client.cpp:499-500, s3_filesystem_producer.cpp:167-168). Set `config.region = "us-east-1"` before the assertion, or drop the region check.

PR #560, opened 2026-06-15
enhance: integrate S3 CRT async reads

jiaqizho:
The goal here is to make the optional AWS S3 CRT path actually useful for Vortex reads, not just available at build time. Vortex can issue many range reads concurrently, so when CRT is enabled we want those reads to go through a real async S3 path instead of pushing synchronous C++ read calls onto a blocking pool. At the same time, regular builds and non-CRT readers still need to keep the existing behavior.

This change adds an async read-at capability to the filesystem reader FFI, lets S3 input files expose that capability through a CRT-backed implementation, and teaches the Rust Vortex bridge to detect it at runtime. When async support is available, Vortex submits reads through the new callback-based FFI path and resolves them back into Rust futures; otherwise it falls back to the existing synchronous read flow.

Parquet is covered by the same S3 filesystem path as well. The Parquet reader opens files through `S3FileSystem`, and with `WITH_CRT` enabled that returns `ObjectCrtInputFile`; Arrow's Parquet pre-buffer path then issues range reads against that CRT-backed `RandomAccessFile`, so column chunk prefetching can use CRT async S3 reads. The Parquet footer fast path also uses `NonBlockingReadAtFile::ReadAtAsyncInto` when CRT is available, while keeping the regular `ReadAt` path for non-CRT builds.

On the C++ side, the S3 client construction is split so the normal SDK client and CRT client can share the same option handling, credentials, retry setup, timeouts, proxy settings, and metrics. With `WITH_CRT` enabled, `S3FileSystem` now builds the CRT client for input-file reads and uses CRT `GetObjectAsync` for range reads, while preserving the current sync `ReadAt`, `ReadAsync`, metadata, size, and metric behavior around it.


cpp/src/filesystem/s3/s3_filesystem.cpp
```diff
@@ -2087,19 +2449,23 @@ class S3FileSystem::Impl : public std::enable_shared_from_this<S3FileSystem::Imp
...
-  arrow::Result<std::shared_ptr<ObjectInputFile>> OpenInputFile(const std::string& s, S3FileSystem* fs) {
+  arrow::Result<std::shared_ptr<arrow::io::RandomAccessFile>> OpenInputFile(const std::string& s, S3FileSystem* fs) {
     ARROW_RETURN_NOT_OK(arrow::fs::internal::AssertNoTrailingSlash(s));
     ARROW_ASSIGN_OR_RAISE(auto path, S3Path::FromString(s));
     ARROW_RETURN_NOT_OK(ValidateFilePath(path));
 
     ARROW_RETURN_NOT_OK(CheckS3Initialized());
 
+#ifdef WITH_CRT
+    auto ptr = std::make_shared<ObjectCrtInputFile>(crt_holder_, fs->io_context(), path);
```
liliu-z: The CRT read path is selected purely at compile time (#ifdef WITH_CRT at s3_filesystem.cpp:2459-2463 and 2482-2486) with no runtime check on the provider or auth mechanism. GCP rides on this same S3FileSystem path (GcpFileSystemProducer::Make calls S3FileSystem::Make), and GCP IAM auth works only because the SDK's GoogleHttpClientFactory injects the OAuth2 Bearer header on every request while the SDK itself is fed anonymous credentials. Aws::S3Crt::S3CrtClient uses the CRT native HTTP stack and never consults that HttpClientFactory, so in a WITH_CRT build every GCP IAM read goes out unauthenticated (401/403) while writes still succeed via the regular SDK client (holder_). The same compile-time switch also forces Aliyun/Tencent/Huawei reads onto an untested CRT proxy/CA/path-style surface. Gate CRT read-path selection on the runtime provider/auth mechanism, not the build flag alone.

cpp/include/milvus-storage/filesystem/s3/s3_client_builder.h
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  arrow::Result<std::shared_ptr<HolderType>> BuildClient(std::optional<arrow::io::IOContext> io_context = std::nullopt,
+                                                         std::shared_ptr<FilesystemMetrics> metrics = nullptr);
+
+  private:
+  ConfigType client_config_;
+};
+
+ClientBuilder(S3Options) -> ClientBuilder<S3Client>;
```
liliu-z, line 94: Moving `ClientBuilder` out of `s3_client.h` into the new `s3_client_builder.h` is source-incompatible for any consumer that included only `s3_client.h` to name `ClientBuilder`; it now must also include `s3_client_builder.h`. All in-repo callers were updated and the bare `ClientBuilder builder(options)` form still compiles via the CTAD guide, so the only break is the include path for external users of this internal plumbing. Note the obvious back-include fix is not clean: `s3_client_builder.h` already includes `s3_client.h` (line 30), so re-including it from `s3_client.h` would be circular and would reference `S3ClientHolder` before it is declared.

cpp/src/filesystem/s3/s3_filesystem.cpp
```diff
@@ -1322,24 +1660,48 @@ class CustomOutputStream final : public arrow::io::OutputStream {
...
     }
-    return std::move(result).Value(&holder_);
+    ARROW_RETURN_NOT_OK(std::move(result).Value(&holder_));
+#ifdef WITH_CRT
+    std::shared_ptr<FilesystemMetrics> metrics;
+    {
+      ARROW_ASSIGN_OR_RAISE(auto client_lock, holder_->Lock());
+      metrics = client_lock.Move()->GetMetrics();
+    }
+    auto crt_result = crt_builder_.BuildClient(io_context_, std::move(metrics));
```
liliu-z: Under WITH_CRT, Impl::Init() unconditionally builds the S3 CRT client, but UseCrtReadPath() returns false for GCP so GCP reads always use the synchronous ObjectInputFile. For GCP the CRT client (configured with throughputTargetGbps=50 plus its own CRT event loop and connection manager) is constructed and retained only to back GetMetrics(), never for I/O, and it introduces a new failure mode where a CRT-client construction error now fails S3FileSystem init for GCP users who never touch the CRT path. Consider skipping CRT client construction for GCP (and guarding GetMetrics(), which currently assumes crt_holder_ is non-null).

cpp/src/filesystem/s3/s3_filesystem.cpp
```diff
@@ -1351,6 +1715,10 @@ class S3FileSystem::Impl : public std::enable_shared_from_this<S3FileSystem::Imp
 
   const S3Options& options() const { return builder_.options(); }
 
+#ifdef WITH_CRT
+  bool UseCrtReadPath() const { return options().cloud_provider != kCloudProviderGCP; }
```
liliu-z, line 1719: UseCrtReadPath() sends every provider except GCP through the new ObjectCrtInputFile path, so the producer default (aws) and a directly-constructed S3Options (empty cloud_provider) both take the CRT branch, as do aliyun/tencent/huawei. The added non-cloud tests only assert the GCP->SDK fallback (OpenInputFileFallsBackToSdkFileForGcpProvider); the CRT-backed assertion (OpenInputFileUsesCrtBackedAsyncFileWhenCrtEnabled) is gated behind IsCloudEnv() and is skipped in credential-less CI. Consider adding a non-cloud test asserting a non-GCP S3Options yields a NonBlockingReadAtFile, mirroring the existing GCP test — OpenInputFile's CRT branch does no network I/O in Init(), so it can run locally.

PR #561, opened 2026-06-16
enhance: add configurable CRT async S3 read benchmarks

jiaqizho:
We need a practical way to compare the AWS CRT async S3 read path with the existing blocking S3 reader under real concurrent read patterns, instead of only relying on the default behavior hidden behind the build flag.

This adds a user-facing `fs.s3.crt_async_read` property so CRT builds can switch between async CRT reads and the normal ObjectInputFile path, while keeping async reads enabled by default. The option is wired through filesystem config, S3 options, config hashing/equality, logging, and the C FFI property exports so callers and benchmarks can control it consistently.

The benchmark side adds a dedicated CrtReader suite that prepares synthetic remote data and measures block vs async-crt behavior for concurrent open/read, take, and get_all_chunks workloads. It records open, warmup, read, wall time, peak threads, payload size, file size, and throughput so the results are easier to compare across formats and data shapes. To make those runs representative, the benchmark data helpers now preserve the requested vector dimension and support FixedSizeList<float> vectors, including raw-size accounting and scalar/vector projection handling.


cpp/include/milvus-storage/ffi_c.h
```diff
@@ -102,6 +102,7 @@ FFI_EXPORT extern const char* loon_properties_fs_load_frequency;
 FFI_EXPORT extern const char* loon_properties_fs_tls_min_version;
 FFI_EXPORT extern const char* loon_properties_fs_background_writes;
 FFI_EXPORT extern const char* loon_properties_fs_use_crc32c_checksum;
+FFI_EXPORT extern const char* loon_properties_fs_s3_crt_async_read;
```
sparknack, line 105: bug: This declares a new FFI symbol, but `cpp/ffi_exports.map` and `cpp/ffi_exports_mac.map` still hide it in Python-binding builds. Add `loon_properties_fs_s3_crt_async_read` to both export lists.

cpp/benchmark/benchmark_crt_reader.cpp
```diff
@@ -0,0 +1,1177 @@
...
+
+  bool CheckCommonRequirements(::benchmark::State& st, const std::string& format, CrtIoMode mode) {
+    if (!IsCloudEnv()) {
+      st.SkipWithError("CRT reader benchmark requires STORAGE_TYPE=remote");
+      return false;
+    }
+    if (!CheckFormatAvailable(st, format)) {
+      return false;
+    }
+#ifndef WITH_CRT
```
sparknack, line 372: risk: Async mode is still allowed for GCP, but `UseCrtReadPath()` disables CRT when `cloud_provider == gcp`, so GCP results get labeled `async-crt` while using the blocking path. Skip async mode for GCP or report the actual read path.

cpp/src/ffi/properties_c.cpp
```diff
@@ -60,6 +60,7 @@ const char* loon_properties_fs_load_frequency = PROPERTY_FS_LOAD_FREQUENCY;
 const char* loon_properties_fs_tls_min_version = PROPERTY_FS_TLS_MIN_VERSION;
 const char* loon_properties_fs_background_writes = PROPERTY_FS_BACKGROUND_WRITES;
 const char* loon_properties_fs_use_crc32c_checksum = PROPERTY_FS_USE_CRC32C_CHECKSUM;
+const char* loon_properties_fs_s3_crt_async_read = PROPERTY_FS_S3_CRT_ASYNC_READ;
```
sparknack, line 63: bug: This adds the C symbol, but Python cffi declarations and the `PropertyKeys` mapping are not updated, so Python users cannot access the new knob without hard-coding the string. Add it to `python/milvus_storage/_ffi.py` and `python/milvus_storage/properties.py`.

jiaqizho: ack

cpp/include/milvus-storage/ffi_c.h
jiaqizho: ack

cpp/benchmark/benchmark_crt_reader.cpp
jiaqizho: In this case, GCP has disabled CRT temporarily.It's fine if async-crt not work in GCP

PR #562, opened 2026-06-18
fix: propagate S3 filesystem failures in vortex writer close

jiaqizho:
Vortex writes used to rely on Drop to close the underlying object-store writer. That made close failures easy to lose, because Drop cannot return an error to the caller. In practice, Writer::close could look successful even if the final storage close/upload step failed, which means higher layers might continue as if the file was safely written.

This change moves that close step into the normal finish path. The Rust bridge now keeps an explicit writer handle, closes it after the Vortex writer finishes, and sends any close error back through the C++ writer as an Arrow IO error. So when the final close fails, Writer::close now reports the failure directly instead of hiding it in cleanup.

The Vortex reader open path is also tightened up so storage/open failures are converted into normal Arrow errors instead of escaping as exceptions. Overall, the API now surfaces storage failures at the point where callers can actually handle them.


cpp/src/format/bridge/rust/src/vortex_bridgeimpl.rs
```diff
@@ -1018,10 +1026,7 @@ impl VortexWriter {
                 footer_size,
             });
         }
-        Ok(crate::vortex_ffi::VortexWriteSummary {
-            file_size: 0,
-            footer_size: 0,
-        })
+        Err("Vortex writer has no active inner writer; close may have already completed, failed, or no data was written".into())
```
liliu-z: Replacing the no-active-writer `Ok(VortexWriteSummary{0,0})` with an unconditional `Err(...)` conflates two cases: a re-close/failed-close (which should error) and a writer that legitimately received no data (which the upper layer expects to succeed). ColumnGroupWriter::Close() always calls format_writer_->Close() and only skips appending the result when written_bytes_ == 0, so an empty Vortex column group now makes the entire Writer::close() fail with IOError. This is reachable when a column group's final Flush() rolls exactly to a file boundary (written_bytes_ reset to 0) and no further rows arrive before close, and it diverges from Parquet, which still closes empty files successfully. Consider distinguishing 'no data written' (return a zero-size success or have the caller skip closing empty writers) from 'already closed/failed'.

cpp/src/format/bridge/rust/src/vortex_bridgeimpl.rs
```diff
@@ -1037,6 +1040,11 @@ impl VortexWriter {
             let summary = w
                 .finish()
                 .map_err(|e| Box::new(VortexError::from(e)) as Box<dyn std::error::Error>)?;
+            if let Some(writer_handle) = self.writer_handle.take() {
+                writer_handle
+                    .close()
+                    .map_err(|e| Box::new(e) as Box<dyn std::error::Error>)?;
+            }
```
congqixia, line 1047: NIT: `VortexWriter::close()` still treats the “never initialized” path as a successful close by returning { file_size: 0, footer_size: 0 } when inner_writer is None. That means opening a Vortex writer and calling close() without ever writing a batch will still produce a successful 
ColumnGroupFile with zero-size metadata, even though no real Vortex file was created/finalized. Before this PR that behavior already existed, and this change touches the close path but keeps it unchanged, so callers can still persist invalid file metadata and only fail later when the reader tries to open the recorded path.

congqixia (approved):
/lgtm

PR #565, opened 2026-06-21
enhance: Vortex writer flush semantics

jiaqizho:
Vortex flush now matches the writer semantics we want: it is a lightweight request to push already-written data further down to the filesystem layer, not a request to finalize or reshape Vortex layout data. This keeps flush safe to call between writes without creating synthetic chunks or changing row group behavior just because the caller wanted to drain buffered data.

The C++ writer now forwards each record batch to the Vortex bridge during Write instead of holding batches in memory until Flush. Flush only reaches the shared object-store writer, and if that writer has not been opened yet it simply does nothing. Once data has reached the filesystem path, the existing filesystem and multipart upload logic remains responsible for deciding how much buffered data can be pushed to remote storage.

The Vortex writer boundary also converts bridge-side failures into Arrow statuses for writer creation, write, flush, and close, so the format writer reports errors through the same API shape as the rest of the storage write path.

### DIFF
--- a/cpp/src/format/bridge/rust/src/vortex_layout_strategy_v2.rs
+++ b/cpp/src/format/bridge/rust/src/vortex_layout_strategy_v2.rs
@@ -1481,12 +1481,26 @@ struct StatsBuildState {
     stats: Arc<[Stat]>,
 }
 
+fn row_group_zone_map_stats(stats: &[Stat]) -> Arc<[Stat]> {
+    stats
+        .iter()
+        .copied()
+        .filter(|stat| {
+            matches!(
+                stat,
+                Stat::Min | Stat::Max | Stat::Sum | Stat::NullCount | Stat::NaNCount
+            )
+        })
+        .collect()
+}
+
 impl StatsBuildState {
     fn new(
         dtype: &DType,
         stats: Arc<[Stat]>,
         max_variable_length_statistics_size: usize,
     ) -> VortexResult<Self> {
+        let stats = row_group_zone_map_stats(&stats);
         let struct_fields = dtype
             .as_struct_fields_opt()
             .ok_or_else(|| vortex_err!("{LAYOUT_ID} writer requires struct input"))?;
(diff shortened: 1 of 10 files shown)


cpp/src/format/bridge/rust/src/vortex_layout_strategy_v2.rs
```diff
@@ -1726,13 +1740,56 @@ pub fn build_row_group_strategy(
...
+}
+
 fn build_data_strategy() -> Arc<dyn LayoutStrategy> {
     let flat = FlatLayoutStrategy {
         inline_array_node: true,
         ..Default::default()
     };
-    let compress_flat = CompressingStrategy::new_btrblocks(flat, false);
-    let chunked_inner = ChunkedLayoutStrategy::new(compress_flat.clone());
+    let compress_flat = CompressingStrategy::new_btrblocks(flat.clone(), true);
```
liliu-z: build_data_strategy flips the second new_btrblocks argument from false to true, enabling exclude_int_dict_encoding, so scalar integer columns on the V2 data path are no longer dictionary-encoded — a silent change to on-disk bytes and compression for existing V2 writers. It lands with two other undocumented V2 layout changes (FixedSizeList columns routed to the uncompressed flat layout at line 1769, zone-map stats narrowed at line 1503) inside a PR titled 'flush semantics', and unlike those the int-dict flip has no accompanying test, so confirm it is intentional and document or split it out rather than shipping a format change under a flush PR.

cpp/src/format/vortex/vortex_writer.cpp
```diff
@@ -44,56 +44,52 @@ VortexFileWriter::VortexFileWriter(const std::shared_ptr<arrow::fs::FileSystem>&
...
+    ARROW_ASSIGN_OR_RAISE(auto arrow_struct_array, batch->ToStructArray());
+    assert(arrow_struct_array->num_fields() != 0);
+    ArrowArray exported_array;
+    ArrowSchema exported_schema;
+    ARROW_RETURN_NOT_OK(arrow::ExportArray(*arrow_struct_array, &exported_array, &exported_schema));
+    vx_writer_.Write(exported_schema, exported_array);
+    written_rows_ += batch->num_rows();
+    return arrow::Status::OK();
+  } catch (const VortexException& e) {
+    closed_ = true;
```
liliu-z, line 64: On a bridge Write failure the writer sets closed_ and returns an error but never aborts the in-flight upload. When VortexFileWriter is later destroyed, vx_writer_'s Drop runs ObjectStoreWriterHandle::close -> output_stream->Close(), which completes (rather than aborts) the multipart upload and finalizes a footer-less partial object at file_path_. This only triggers when a row group was already streamed before the failure (otherwise the C++ writer was never opened and close is a no-op), and the partial object is normally orphaned because a failed Close returns no ColumnGroupFile — so standalone impact is wasted storage until GC. Consider an abort path that discards the partial object on the error path instead of completing it.

cpp/src/format/vortex/vortex_writer.cpp
```diff
@@ -44,56 +44,52 @@ VortexFileWriter::VortexFileWriter(const std::shared_ptr<arrow::fs::FileSystem>&
       properties_(properties) {}
 
 arrow::Status VortexFileWriter::Write(const std::shared_ptr<arrow::RecordBatch> batch) {
-  assert(!closed_);
-  assert(batch->schema()->Equals(*schema_, false));
-  written_rows_ += batch->num_rows();
-
-  ARROW_ASSIGN_OR_RAISE(auto arrow_struct_array, batch->ToStructArray());
+  try {
+    assert(!closed_);
```
liliu-z, line 52: The closed_ guard in Write (48), Flush (67), and Close (77) is only an assert, which NDEBUG compiles out, so release builds do not enforce it. That now matters because Write sets closed_=true on a bridge failure (line 60): a later Close() in release skips the assert and, since the bridge inner_writer is already None, returns a bogus ColumnGroupFile with file_size/footer_size 0 instead of an error — which the outer close() records straight into the manifest — and a later Write() would re-lazy-init the bridge and reopen the same path. Match the sibling ParquetFileWriter::Close, which guards with `if (closed_ || !writer_) return arrow::Status::Invalid(...)`.

cpp/src/format/vortex/vortex_writer.cpp
```diff
@@ -44,56 +44,52 @@ VortexFileWriter::VortexFileWriter(const std::shared_ptr<arrow::fs::FileSystem>&
...
-  for (const auto& struct_array : column_arrays_) {
-    ARROW_RETURN_NOT_OK(arrow::ExportArray(*struct_array, &exported_array, &exported_schema));
-    try {
-      vx_writer_.Write(exported_schema, exported_array);
-    } catch (const VortexException& e) {
-      return arrow::Status::IOError(e.what());
-    }
+  try {
+    assert(!closed_);
+    vx_writer_.Flush();
```
liliu-z, line 76: FormatWriter::Flush is documented as leaving no data in memory (format_writer.h:41) and the outer flush() zeroes current_memory_usage_ on that basis (writer.cpp:378), but the new Vortex Flush only flushes the object-store output stream and does not drain the bridge blocking writer's buffered row groups, so data can remain in memory after Flush. This is not an OOM regression — Write streams full row groups so memory stays bounded and the PR even removes the C++ column_arrays_ buffer — but the base-class contract comment is now inaccurate and should be reconciled with the new lightweight-push-to-filesystem semantics.

PR #566, opened 2026-06-22
enhance: derive Vortex footer size from the written file layout

jiaqizho:
Vortex footer size is stored in the manifest and later used by readers to fetch the footer tail in one shot, so it needs to describe the bytes Vortex actually wrote instead of a reserialized estimate. The old writer path rebuilt the footer and used that serialized length, which could drift from the real file layout and make the cached size smaller than the actual footer.

This changes the writer to calculate the footer start from the segment map returned by Vortex after writing. The cached footer_size is now the bytes after the last data segment, minus the fixed EOF marker. Empty segment maps are handled as a four-byte Vortex header, which keeps zero-row files consistent with the actual on-disk layout.

The footer reader now treats a non-zero cached footer size as the fast path: it reads footer_size + EOF once and opens the sparse file directly. If that open fails, or if the opened footer reports a larger real footer body, the reader falls back to the bounded expanding retry path instead of returning an eager error.

The reader also avoids the previous fixed 4KB prefix prefetch. Known footer sizes only materialize the footer tail into the sparse range file, removing an unnecessary extra read against object stores while preserving the existing retry path for unknown, legacy, or stale footer sizes.

This keeps the contract clear: footer_size means footer body size only, the read path adds EOF when fetching the tail, and unknown or legacy footer sizes can still use the speculative retry path.

### DIFF
--- a/cpp/src/format/bridge/rust/src/lib.rs
+++ b/cpp/src/format/bridge/rust/src/lib.rs
@@ -278,7 +278,7 @@ pub mod vortex_ffi {
         fn row_group_zone_map_count(self: &VortexFile) -> Result<u64>;
         fn row_group_zone_map_data_before_zones(self: &VortexFile) -> Result<bool>;
         fn zone_map_segment_ids(self: &VortexFile) -> Result<Vec<u64>>;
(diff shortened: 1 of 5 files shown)


jiaqizho:
@aoiasd PTAL

PR #567, opened 2026-06-23
enhance: default zstd-3 and uncompress vector columns

tedxu:
See milvus-io/milvus#50452.

Two changes that together cut parquet write CPU on vector
workloads without hurting file size:

1. PROPERTY_WRITER_COMPRESSION_LEVEL registry default lowered
   from 5 to 3. At level 5 ZSTD switches from the dfast strategy
   to greedy with row matchfinder; on random float embeddings
   that costs ~8% more CPU for near-zero size benefit. The Milvus
   import benchmark observed +10% wall time on V3 (which used the
   level-5 default) over V2 (which forces level 3 via the legacy
   ParquetFileWriter coercion).

2. ParquetFileWriter emits FIXED_SIZE_BINARY and BINARY columns
   as UNCOMPRESSED with statistics disabled. Vector payloads
   behave like incompressible bytes; ZSTD spends CPU on them
   without shrinking the output.

A new benchmark in cpp/benchmark/benchmark_compression_levels.cpp
exercises the two configurations (zstd-3 vs zstd-5) through
ParquetFileWriter::Make on a pk + text + vector schema.

Tests:
- APIPropertiesTest.basic updated for the new default
- ParquetFileWriterTest.VectorColumnsAreUncompressed
- ParquetFileWriterTest.FileLevelCompressionDoesNotPreventVectorUncompressed

### DIFF
--- a/cpp/src/format/parquet/parquet_writer.cpp
+++ b/cpp/src/format/parquet/parquet_writer.cpp
@@ -18,6 +18,7 @@
 
 #include <arrow/io/buffered.h>
 #include <arrow/io/memory.h>
+#include <arrow/util/compression.h>
 #include <parquet/properties.h>
 #include <parquet/metadata.h>
 #include <boost/variant.hpp>
@@ -129,6 +130,12 @@ ParquetFileWriter::ParquetFileWriter(std::shared_ptr<arrow::Schema> schema,
     auto deep_copied_decryption = writer_props->file_encryption_properties()->DeepClone();
     builder.encryption(std::move(deep_copied_decryption));
   }
+  // Legacy callers (e.g. PackedRecordBatchWriter) often construct an empty
+  // parquet::WriterProperties whose file-level compression defaults to
+  // UNCOMPRESSED. Apply ZSTD level 3 in that case so the legacy path
+  // matches the property-registry default used by the new properties-based
+  // ParquetFileWriter::Make. Explicit non-UNCOMPRESSED settings from the
+  // caller are preserved.
   if (writer_props->default_column_properties().compression() == ::parquet::Compression::UNCOMPRESSED) {
     builder.compression(::parquet::Compression::ZSTD);
     builder.compression_level(3);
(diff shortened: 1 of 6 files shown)


cpp/src/format/parquet/parquet_writer.cpp
```diff
@@ -139,7 +146,13 @@ ParquetFileWriter::ParquetFileWriter(std::shared_ptr<arrow::Schema> schema,
...
       case arrow::Type::FIXED_SIZE_BINARY:
       case arrow::Type::BINARY:
-        // Disable statistics for vector columns
+        // Vector columns (dense fixed-size + sparse binary) are effectively
+        // random bytes; ZSTD wastes CPU on them without shrinking the
+        // output. Always emit them UNCOMPRESSED and skip statistics. The
+        // per-column level is reset to the codec-default sentinel because
+        // parquet rejects "UNCOMPRESSED + explicit level" at build time.
+        builder.compression(field->name(), ::parquet::Compression::UNCOMPRESSED);
+        builder.compression_level(field->name(), ::arrow::util::kUseDefaultCompressionLevel);
```
congqixia: The new “vector column” rule is keyed only on Arrow physical type (BINARY/FIXED_SIZE_BINARY), but we also uses arrow::binary() for ordinary non-vector LOB/binary fields. (A non-vector binary/blob column would now be forced to UNCOMPRESSED)

congqixia (commented):
NIT: it's ok to sync v2/v3 behavior but kind of weird that we could not specify "UNCOMPRESSED" with v2 API (which will be deprecated soon)

tedxu:
On keeping the `if (default == UNCOMPRESSED) -> ZSTD-3` block in the
ParquetFileWriter constructor: this is intentional, for backward
compatibility.

Existing V2 callers (notably milvus-io/milvus's
`internal/core/src/segcore/packed_writer_c.cpp` for both
`NewPackedWriter` and `NewPackedWriterWithStorageConfig`) construct
an empty `parquet::WriterProperties` and pass it through
`PackedRecordBatchWriter::Make` -> `ParquetFileWriter::Make`. The
file-level default in that empty WriterProperties is UNCOMPRESSED.

Dropping the coercion would silently change V2's on-disk output
from ZSTD-3 to UNCOMPRESSED for all columns, including non-vector
ones. Keeping the coercion makes the legacy path agree with the
new properties-based path (registry default zstd / level 3) without
asking Milvus to change its callers.

congqixia (approved):
/lgtm

PR #569, opened 2026-06-26
Fix: parquet projection for nested Arrow fields

jiaqizho:
Parquet reads were still treating projected columns as top-level Arrow field indexes, but ReadRowGroup actually expects Parquet leaf-column indexes. That worked for flat schemas, but nested fields like structs expand into multiple Parquet leaf columns, so projecting a nested top-level field could read the wrong physical columns or return an incomplete batch.

This change makes the projection path explicit: resolve requested top-level Arrow fields by name, compute each field's leaf-column range from the file schema, and pass the expanded leaf-column indexes into Parquet row group reads. Range reads use the same leaf-column projection and rebuild the returned schema from the requested top-level fields, so normal chunk reads and range reads now follow the same projection semantics.

The naming was also cleaned up around this path so the code says leaf column where it really means Parquet leaf columns, instead of mixing that up with Arrow field indexes. Fixed-size-list roundtrip is left disabled for now because that behavior is still separate from the nested projection fix.

### DIFF
--- a/cpp/src/format/parquet/parquet_format_reader.cpp
+++ b/cpp/src/format/parquet/parquet_format_reader.cpp
@@ -25,6 +25,7 @@
 #include <fmt/format.h>
 
 #include <arrow/array/util.h>
+#include <arrow/extension_type.h>
 #include <arrow/io/caching.h>
 #include <arrow/record_batch.h>
 #include <arrow/table.h>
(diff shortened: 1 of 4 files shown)


cpp/src/format/parquet/parquet_format_reader.cpp
```diff
@@ -402,36 +403,97 @@ arrow::Status ParquetFormatReader::open() {
...
-// needed_column_indices_.
+// separate read_schema_/file_schema_ like Lance/Vortex. Projection needs both
+// Arrow field indices and Parquet leaf column indices.
 std::shared_ptr<arrow::Schema> ParquetFormatReader::get_schema() const { return schema_; }
 
-arrow::Status ParquetFormatReader::set_needed_columns(const std::vector<std::string>& needed_columns) {
-  if (!schema_) {
-    return arrow::Status::Invalid(fmt::format("Parquet file schema is not initialized. [path={}]", path_));
+static int get_leaf_column_count(const std::shared_ptr<arrow::DataType>& type) {
+  if (type->num_fields() == 0) {
```
liliu-z, line 417: get_leaf_column_count treats any type with num_fields() == 0 as a single leaf, but Arrow extension types report zero fields (their structure lives in storage_type(), not children_) while the Parquet writer expands them to that storage type before emitting columns. A registered extension over a nested storage type (e.g. struct<a,b>) therefore writes N leaf columns but is counted as 1, shifting every later field's leaf offset and selecting the wrong or incomplete columns on projection. Latent today since this repo registers no such extension types; fix by unwrapping Type::EXTENSION to storage_type() before the num_fields()==0 check, mirroring the writer.

PR #572, opened 2026-06-27
enhance: add packed-specific V2 extend status codes 

jiaqizho:
This change adds a dedicated extend status detail for the packed layer, so packed failures can carry structured error codes instead of relying only on generic Arrow status text.

The packed codes split caller-side invalid input from storage-side and internal failures. Bad schemas, invalid column groups, null output pointers, and null batches now land in PackedInvalidArgs, while file open/read/write/flush/close failures are reported as storage I/O. Corrupted packed metadata, broken file layout, Arrow API failures, and unexpected exceptions each get their own bucket.

The implementation keeps the existing packed behavior intact on the normal path. Reader construction still throws when initialization fails, writer initialization still goes through the existing Result flow, and Write(nullptr) remains a successful no-op. The new catch blocks stay around packed APIs that already return Arrow status, so exceptions are converted only at the public packed boundary where callers can consume them.

### DIFF
--- a/cpp/src/common/extend_status.cpp
+++ b/cpp/src/common/extend_status.cpp
@@ -89,4 +89,24 @@ arrow::Status WrapExtendError(ExtendStatusCode code, std::string message, const 
   return MakeExtendError(wrapped_code, fmt::format("{}: {}", message, cause_message), cause_message);
 }
 
+milvus::ErrorCode ToSegcoreErrorCode(ExtendStatusCode code) {
+  if (code == ExtendStatusCode::PackedInvalidArgs) {
+    return milvus::InvalidParameter;
+  }
+  return milvus::StorageError;
+}
+
+milvus::SegcoreError ToSegcoreError(const arrow::Status& status) {
+  if (status.ok()) {
+    return milvus::SegcoreError::success();
+  }
+
+  auto detail = ExtendStatusDetail::UnwrapStatus(status);
+  if (detail) {
+    return {ToSegcoreErrorCode(detail->code()), status.ToString()};
+  }
+  auto code = status.IsInvalid() ? ExtendStatusCode::PackedInvalidArgs : ExtendStatusCode::PackedArrowError;
+  return {ToSegcoreErrorCode(code), status.ToString()};
+}
+
 }  // namespace milvus_storage
--- a/cpp/src/common/extend_status.cpp
+++ b/cpp/src/common/extend_status.cpp
@@ -20,6 +20,7 @@
 
 #include <arrow/status.h>
 #include <arrow/result.h>
+#include <fmt/format.h>
 
 namespace milvus_storage {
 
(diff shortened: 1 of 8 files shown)


czs007:
Reviewed the change — overall **LGTM**. Nice work giving the packed layer structured, semantic error codes. A few non-blocking notes below.

### What looks good

- **Real bug fixes bundled in**, not just renaming statuses:
  - `init()` now validates `!fs` / `paths.empty()` / `!schema` up front, before `schemaMatching` dereferences `paths[0]`.
  - `schemaMatching` bounds-checks `column_offset.path_index` against `paths.size()` before `paths[path_index]` — closes a potential OOB read on corrupted metadata.
  - `Write()` validates record column indices against `record->num_columns()`, avoiding a splitter OOB on short batches.
- `WrapExtendError` preserving the **innermost** `ExtendStatusCode` when the cause already carries an `ExtendStatusDetail` is a good call — root cause survives the wrap.
- No regression to the FIU fault-injection tests: they assert on the `"Injected fault"` substring, not `IsIOError()`, so switching the injected status to `PackedStorageIO` is safe.
- New test file is auto-picked up via the `GLOB_RECURSE test/*.cpp` rule, and the per-code coverage (incl. the wrap-preserves-root-code case) is solid.

### Non-blocking nits

1. **Duplicated cause text in `WrapExtendError`.** The `message` becomes `"{outer}: {cause.ToString()}"` while `extra_info` also stores the full `cause.ToString()`. When wraps nest (cause is itself an extend error), `ToString()` re-nests `CodeName: extra_info` and the string keeps growing. Consider storing `cause.message()` in `extra_info` instead of the full `ToString()`.

2. **Broad `catch (...) -> PackedUnexpected`** at the `Write`/`ReadNext`/`Close`/`Tell` boundary also swallows programming errors. Note the `assert(...)` calls inside `Write` still abort in debug but fall through to the catch in release — slightly inconsistent behavior between build modes. Acceptable at a public boundary, just flagging.

3. The inner `try/catch` around `SliceChunksByMaxContiguousSlice` overlaps the outer one, but since it yields a more precise `PackedFileCorrupted` it's a justified redundancy.

None of these block merge.

czs007:
Two follow-ups to my earlier review.

**Correction to nit #2 (it's actually a bit worse than I phrased it).** I said the `assert(...)` calls "fall through to the catch in release" — that's misleading. In a release build (`NDEBUG`) the assert compiles away entirely, and the guarded access uses `operator[]`:

```cpp
assert(group.GrpId() < group_writers_.size());
auto& grp_writer = group_writers_[group.GrpId()];   // OOB here is UB, not an exception
```

`std::vector::operator[]` out-of-range is **undefined behavior, not a C++ exception**, so the surrounding `catch (...)` will *not* reliably catch it. So the catch does not make these asserts safe in release. If this invariant can realistically be violated by bad input, prefer an explicit bounds check returning `PackedInvalidArgs` (or `.at()`) over `assert` + `operator[]`.

**Carrier-code reminder for the downstream FFI bridge.** `MakeExtendError` sets the arrow carrier `StatusCode` to `IOError` for every code except `PackedInvalidArgs` — including `PackedMetadataCorrupted` / `PackedFileCorrupted` (which are *permanent*). The fine code only lives in the `ExtendStatusDetail`. So whoever wires the FFI boundary (`reader_c.cpp` / `writer_c.cpp`) must read the code via `ExtendStatusDetail::UnwrapStatus(status)->code()` and **must not** classify off the coarse arrow `StatusCode` — otherwise permanent corruption collapses into `LOON_IO_ERROR` and gets treated as retriable downstream. Not a change request for this PR (the FFI mapping is a separate, in-flight step), just flagging so the carrier-code collapse doesn't get read as a retry signal later.

cpp/src/common/extend_status.cpp
czs007: std::string CodeAsString(ExtendStatusCode code) {
  switch (code) {
    case ExtendStatusCode::PackedInvalidArgs:   return "PackedInvalidArgs";
    case ExtendStatusCode::PackedStorageIO:     return "PackedStorageIO";
    /* ...one case per enumerator, no default... */
    case ExtendStatusCode::TxnResolutionFailed: return "TxnResolutionFailed";
  }
  return "Unknown";   // fallback lives outside the switch
}

Why this shape:
- No default inside the switch → -Werror=switch stays active, so adding a new ExtendStatusCode without a case is a compile error (not a silent fall-through to "Unknown").
- return "Unknown"; after the switch → handles an out-of-enum value (e.g. a bad cast) and satisfies -Wreturn-type ("all paths return"); it does not suppress -Wswitch (only a default: would).

jiaqizho: ack

jiaqizho:
replace by #575

PR #573, opened 2026-06-27
enhance:[2.6] add packed-specific v2 extend status codes

jiaqizho:
This change adds a dedicated v2 extend status detail for the packed layer, so packed failures can carry structured error codes instead of relying only on generic Arrow status text.

The packed codes split caller-side invalid input from storage-side and internal failures. Bad schemas, invalid column groups, null output pointers, and null batches now land in PackedInvalidArgs, while file open/read/write/flush/close failures are reported as storage I/O. Corrupted packed metadata, broken file layout, Arrow API failures, and unexpected exceptions each get their own bucket.

The implementation keeps the existing packed behavior intact on the normal path. Reader construction still throws when initialization fails, writer initialization still goes through the existing Result flow, and Write(nullptr) remains a successful no-op. The new catch blocks stay around packed APIs that already return Arrow status, so exceptions are converted only at the public packed boundary where callers can consume them.

### DIFF
--- a/cpp/src/common/extend_status.cpp
+++ b/cpp/src/common/extend_status.cpp
@@ -17,6 +17,7 @@
 #include <optional>
 #include <string>
 #include <string_view>
+#include <utility>
 
 #include <arrow/status.h>
 #include <arrow/result.h>
@@ -39,6 +40,18 @@ std::string ExtendStatusDetail::extra_info() const { return extra_info_; }
 
 std::string ExtendStatusDetail::CodeAsString() const {
   switch (code()) {
+    case ExtendStatusCode::PackedInvalidArgs:
+      return "PackedInvalidArgs";
+    case ExtendStatusCode::PackedStorageIO:
+      return "PackedStorageIO";
+    case ExtendStatusCode::PackedMetadataCorrupted:
+      return "PackedMetadataCorrupted";
+    case ExtendStatusCode::PackedFileCorrupted:
+      return "PackedFileCorrupted";
+    case ExtendStatusCode::PackedArrowError:
+      return "PackedArrowError";
+    case ExtendStatusCode::PackedUnexpected:
+      return "PackedUnexpected";
     case ExtendStatusCode::NoSuchUpload:
       return "NoSuchUpload";
     default:
@@ -56,9 +69,16 @@ std::shared_ptr<ExtendStatusDetail> ExtendStatusDetail::UnwrapStatus(const arrow
 }
 
 arrow::Status MakeExtendError(ExtendStatusCode code, std::string message, std::string extra_info) {
-  arrow::StatusCode arrow_code = arrow::StatusCode::IOError;
-  return arrow::Status(arrow_code, std::move(message),
-                       std::make_shared<ExtendStatusDetail>(code, std::move(extra_info)));
+  auto arrow_code =
+      code == ExtendStatusCode::PackedInvalidArgs ? arrow::StatusCode::Invalid : arrow::StatusCode::IOError;
+  return {arrow_code, std::move(message), std::make_shared<ExtendStatusDetail>(code, std::move(extra_info))};
+}
+
+arrow::Status WrapExtendError(ExtendStatusCode code, std::string message, const arrow::Status& cause) {
+  auto detail = ExtendStatusDetail::UnwrapStatus(cause);
+  auto wrapped_code = detail ? detail->code() : code;
+  auto cause_message = cause.ToString();
+  return MakeExtendError(wrapped_code, std::move(message) + ": " + cause_message, cause_message);
 }
 
 }  // namespace milvus_storage
--- a/cpp/src/packed/reader.cpp
+++ b/cpp/src/packed/reader.cpp
@@ -14,6 +14,11 @@
 
 #include <memory>
 #include <algorithm>
+#include <exception>
+#include <iterator>
+#include <stdexcept>
+#include <string>
+#include <utility>
 
 #include <arrow/array/data.h>
 #include <arrow/array/util.h>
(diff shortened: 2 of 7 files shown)


cpp/src/common/extend_status.cpp
```diff
@@ -20,6 +20,7 @@
 
 #include <arrow/status.h>
 #include <arrow/result.h>
+#include <fmt/format.h>
```
liliu-z: This PR adds `fmt::format` and `#include <fmt/format.h>` in extend_status.cpp, reader.cpp, and writer.cpp, but fmt is never declared as a conan requirement (conanfile.py) nor linked in cpp/CMakeLists.txt, and no existing dependency provides it transitively (a tree-wide grep finds fmt only in these three files). The milvus-storage shared library will fail to compile or link once CI builds it; add fmt to the conan requirements and link `fmt::fmt`, or fall back to the existing std::string concatenation style.

cpp/src/packed/reader.cpp
```diff
@@ -66,11 +113,12 @@ arrow::Status PackedRecordBatchReader::init(std::shared_ptr<arrow::fs::FileSyste
   for (const auto& path : needed_paths_) {
     auto result = MakeArrowFileReader(*fs, path, reader_props, arrow_reader_props);
     if (!result.ok()) {
-      return arrow::Status::Invalid("Error making file reader with path " + path + ":" + result.status().ToString());
+      return WrapExtendError(ExtendStatusCode::PackedStorageIO,
```
liliu-z: The reader's init paths build structured errors such as PackedStorageIO here, but the packed reader's only public entry is the throwing constructor, which at reader.cpp:89 does `throw std::runtime_error(status.ToString())`, and reader.h exposes no Result-returning factory. The ExtendStatusDetail and its ExtendStatusCode are destroyed, leaving callers only the flattened text, so construction failures can't be classified by code the way the writer's Make() factory allows. Consider a Result-returning factory for the reader.

cpp/src/packed/reader.cpp
```diff
@@ -240,58 +312,93 @@ arrow::Status PackedRecordBatchReader::advanceBuffer() {
...
-      // RVO here, no need std::move
-      arrays.emplace_back(batch->column(batch_index++));
-    } else {
-      auto null_array = arrow::MakeArrayOfNull(schema_->field(i)->type(), chunk_size).ValueOrDie();
-      arrays.emplace_back(std::move(null_array));
+    // Determine the maximum contiguous slice across all tables
+    std::vector<std::shared_ptr<arrow::ArrayData>> batch_data;
+    try {
+      batch_data = chunk_manager_->SliceChunksByMaxContiguousSlice(row_limit_ - absolute_row_position_, tables_);
+    } catch (const std::exception& e) {
```
liliu-z, line 332: The inner catch (const std::exception&) at reader.cpp:332 wraps every exception from SliceChunksByMaxContiguousSlice (chunk_manager.cpp:32) as PackedFileCorrupted. That function signals genuine layout corruption with std::runtime_error (e.g. "Table is empty for path_index...", "Column index ... out of bounds"), but it also allocates four containers and calls chunk->Slice(...), any of which can throw std::bad_alloc under memory pressure or very wide schemas. An OOM on a healthy file is then misreported as data corruption, sending callers down the wrong recovery path. Catch only the known layout exceptions as PackedFileCorrupted and let std::bad_alloc / unexpected exceptions fall through to the outer PackedUnexpected handler.

cpp/src/common/extend_status.cpp
```diff
@@ -56,9 +69,16 @@ std::shared_ptr<ExtendStatusDetail> ExtendStatusDetail::UnwrapStatus(const arrow
...
+  auto arrow_code =
+      code == ExtendStatusCode::PackedInvalidArgs ? arrow::StatusCode::Invalid : arrow::StatusCode::IOError;
+  return {arrow_code, std::move(message), std::make_shared<ExtendStatusDetail>(code, std::move(extra_info))};
+}
+
+arrow::Status WrapExtendError(ExtendStatusCode code, std::string message, const arrow::Status& cause) {
+  auto detail = ExtendStatusDetail::UnwrapStatus(cause);
+  auto wrapped_code = detail ? detail->code() : code;
+  auto cause_message = cause.ToString();
+  return MakeExtendError(wrapped_code, fmt::format("{}: {}", message, cause_message), cause_message);
```
liliu-z: WrapExtendError writes the full cause into both the message and extra_info. Because arrow::Status::ToString() already appends the inner detail's ToString(), the cause text is emitted 2-3x per wrap and compounds on nested wraps. Combined with full schema_->ToString(true) dumps spliced into init errors (writer.cpp:91,102; reader.cpp:166), a wide Milvus schema produces very large error/log lines. Error-path only, so it is log bloat rather than a hot-path cost. Put the cause in extra_info only and avoid re-dumping the entire schema.

cpp/src/packed/reader.cpp
```diff
@@ -24,17 +28,50 @@
...
+  }
+
+  try {
+    auto result = PackedFileMetadata::Make(metadata);
+    if (!result.ok()) {
+      return WrapExtendError(ExtendStatusCode::PackedMetadataCorrupted,
+                             fmt::format("Failed to parse packed file metadata. [path={}]", path), result.status());
+    }
+    return result;
+  } catch (const std::exception& e) {
```
liliu-z, line 64: MakePackedMetadata's `catch (const std::exception&)` maps every exception from PackedFileMetadata::Make to PackedMetadataCorrupted, so a std::bad_alloc or other unexpected failure during metadata parsing is misreported as a corrupt file — the same over-broad-catch pattern flagged for the SliceChunks catch. Catch the known parse-failure type and let std::bad_alloc and other unexpected exceptions surface as PackedUnexpected.

jiaqizho:
replace by #575

PR #575, opened 2026-07-01
enhance: classify packed extend status codes into segcore error codes

czs007:
## What

Add the producer-owned mapping `ToSegcoreErrorCode` / `ToSegcoreError` so milvus-storage classifies its own packed failures once, at the source, and milvus can consume a typed segcore `ErrorCode` instead of a generic Arrow status text.

> Note: this **builds on #572** (jiaqizho — packed `ExtendStatusCode` + `WrapExtendError`), which is not yet merged, so this branch currently includes those commits. Once #572 lands this PR should be rebased onto it so it only carries the mapping delta. Happy to coordinate.

## Mapping (no-`default` switch + `-Werror=switch`)

| ExtendStatusCode | segcore ErrorCode | retry |
|---|---|---|
| PackedInvalidArgs | InvalidParameter (2042) | non-retriable (input) |
| PackedStorageIO | **StorageTransientError (2045)** | **retriable** |
| PackedMetadataCorrupted / PackedFileCorrupted | DataFormatBroken (2024) | permanent |
| PackedArrowError / PackedUnexpected / AWS / Txn | StorageError (2044) | permanent |

The retriable verdict is load-bearing: a transient storage IO failure maps to the retriable `StorageTransientError(2045)` and must never collapse into the non-retriable `StorageError(2044)`. The switch has no `default` + a post-switch fallback under `-Werror=switch`, so a new `ExtendStatusCode` breaks the build until classified. The no-detail path keeps a coarse arrow mapping (IO→transient, OOM→mem-allocate, Invalid/Type/Key→corruption).

## Dependencies

Depends on `milvus-common` `StorageTransientError(2045)`: zilliztech/milvus-common#102. Bump the `milvus-common` conan pin to the published 2045 revision before merge (the pin here points at a local build for verification).

Design + tracking: milvus-io/milvus#50903. Consumed by milvus-io/milvus#50768.

### DIFF
--- a/cpp/conanfile.py
+++ b/cpp/conanfile.py
@@ -167,11 +167,7 @@ class StorageConan(ConanFile):
         self.requires("libavrocpp/1.12.1.1@milvus/dev#cde7bb587a29f6f233bae7e18b71815d")
         self.requires("google-cloud-cpp/2.28.0@milvus/dev#468918b43cec43624531a0340398cf43")
         self.requires("opentelemetry-cpp/1.23.0@milvus/dev#11bc565ec6e82910ae8f7471da756720")
-        # NOTE: ToSegcoreErrorCode maps PackedStorageIO to milvus::StorageTransientError(2045),
-        # which is added by the milvus-common change "add retriable StorageTransientError(2045)".
-        # Bump this pin to the official milvus-common revision that ships 2045 before merge;
-        # the value below points at a local build used for verification.
-        self.requires("milvus-common/1.0.0-6fd4b2d@milvus/dev#9139f6b9bb6bbac7b9d85992e1bbad07")
+        self.requires("milvus-common/1.0.0-60a563c@milvus/dev#a7448f82ed17d10934eacb6d1b152fd8")
         # azure-sdk-for-cpp is a transitive dep of Arrow, but must be declared
         # as a direct dep so CMakeDeps generates standalone cmake config files.
         # Without this, find_package(Azure) can't find include directories.
--- a/cpp/conanfile.py
+++ b/cpp/conanfile.py
@@ -167,7 +167,11 @@ class StorageConan(ConanFile):
         self.requires("libavrocpp/1.12.1.1@milvus/dev#cde7bb587a29f6f233bae7e18b71815d")
         self.requires("google-cloud-cpp/2.28.0@milvus/dev#468918b43cec43624531a0340398cf43")
         self.requires("opentelemetry-cpp/1.23.0@milvus/dev#11bc565ec6e82910ae8f7471da756720")
-        self.requires("milvus-common/1.0.0-9ca5ea6@milvus/dev#274d428d85f1d3d996e1092f0c9c7144")
+        # NOTE: ToSegcoreErrorCode maps PackedStorageIO to milvus::StorageTransientError(2045),
+        # which is added by the milvus-common change "add retriable StorageTransientError(2045)".
+        # Bump this pin to the official milvus-common revision that ships 2045 before merge;
+        # the value below points at a local build used for verification.
+        self.requires("milvus-common/1.0.0-6fd4b2d@milvus/dev#9139f6b9bb6bbac7b9d85992e1bbad07")
         # azure-sdk-for-cpp is a transitive dep of Arrow, but must be declared
         # as a direct dep so CMakeDeps generates standalone cmake config files.
         # Without this, find_package(Azure) can't find include directories.
--- a/cpp/src/common/extend_status.cpp
+++ b/cpp/src/common/extend_status.cpp
@@ -58,6 +58,12 @@ std::string ExtendStatusDetail::CodeAsString() const {
       return "AwsErrorConflict";
     case ExtendStatusCode::AwsErrorPreConditionFailed:
       return "AwsErrorPreConditionFailed";
+    case ExtendStatusCode::AwsErrorNotFound:
+      return "AwsErrorNotFound";
+    case ExtendStatusCode::AwsErrorAccessDenied:
+      return "AwsErrorAccessDenied";
+    case ExtendStatusCode::AwsErrorNonRetryable:
+      return "AwsErrorNonRetryable";
     case ExtendStatusCode::TxnExhaustedRetry:
       return "TxnExhaustedRetry";
     case ExtendStatusCode::TxnResolutionFailed:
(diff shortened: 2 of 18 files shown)


cpp/src/common/extend_status.cpp
```diff
@@ -64,8 +77,90 @@ std::shared_ptr<ExtendStatusDetail> ExtendStatusDetail::UnwrapStatus(const arrow
...
+  }
+
+  // No structured ExtendStatusDetail attached: fall back to arrow's coarse
+  // status code. Transient IO / OOM stay retriable, malformed data is permanent
+  // corruption, and anything else is an internal storage error.
+  milvus::ErrorCode code;
+  if (status.IsOutOfMemory()) {
+    code = milvus::MemAllocateFailed;  // 2034, retriable
+  } else if (status.IsIOError()) {
+    code = milvus::StorageTransientError;  // 2045, retriable
```
jiaqizho: no retriable. notice that: all of v2 error is not retriable.

cpp/src/common/extend_status.cpp
```diff
@@ -64,8 +77,90 @@ std::shared_ptr<ExtendStatusDetail> ExtendStatusDetail::UnwrapStatus(const arrow
...
+// StorageError(2044), which would turn a recoverable IO blip into a hard
+// failure.
+#pragma GCC diagnostic push
+#pragma GCC diagnostic error "-Wswitch"
+milvus::ErrorCode ToSegcoreErrorCode(ExtendStatusCode code) {
+  switch (code) {
+    case ExtendStatusCode::PackedInvalidArgs:
+      return milvus::InvalidParameter;  // 2042, caller's fault (non-retriable input)
+    case ExtendStatusCode::PackedStorageIO:
+      return milvus::StorageTransientError;  // 2045, retriable storage IO
```
jiaqizho: ditto

cpp/conanfile.py
```diff
@@ -167,7 +167,11 @@ def requirements(self):
         self.requires("libavrocpp/1.12.1.1@milvus/dev#cde7bb587a29f6f233bae7e18b71815d")
         self.requires("google-cloud-cpp/2.28.0@milvus/dev#468918b43cec43624531a0340398cf43")
         self.requires("opentelemetry-cpp/1.23.0@milvus/dev#11bc565ec6e82910ae8f7471da756720")
-        self.requires("milvus-common/1.0.0-9ca5ea6@milvus/dev#274d428d85f1d3d996e1092f0c9c7144")
+        # NOTE: ToSegcoreErrorCode maps PackedStorageIO to milvus::StorageTransientError(2045),
+        # which is added by the milvus-common change "add retriable StorageTransientError(2045)".
+        # Bump this pin to the official milvus-common revision that ships 2045 before merge;
+        # the value below points at a local build used for verification.
+        self.requires("milvus-common/1.0.0-6fd4b2d@milvus/dev#9139f6b9bb6bbac7b9d85992e1bbad07")
```
jiaqizho: change to `milvus-common/1.0.0-60a563c@milvus/dev#a7448f82ed17d10934eacb6d1b152fd8`

cpp/src/common/extend_status.cpp
```diff
@@ -64,8 +83,136 @@ std::shared_ptr<ExtendStatusDetail> ExtendStatusDetail::UnwrapStatus(const arrow
...
+  // ErrorToStatus -- so a *plain* IOError that reaches this branch is a genuine
+  // transient (throttle / 5xx / timeout / connection). querynode can still
+  // reroute it to another replica/node, so it MUST stay retriable
+  // (StorageTransientError/2045); collapsing it into StorageError/2044 would
+  // turn a recoverable blip into a hard failure. OOM is retriable; malformed
+  // data is permanent corruption; anything else internal.
+  milvus::ErrorCode code;
+  if (status.IsOutOfMemory()) {
+    code = milvus::MemAllocateFailed;  // 2034, retriable
+  } else if (status.IsIOError()) {
```
jiaqizho, line 203: noretry

cpp/src/common/extend_status.cpp
czs007: Resolved in 5a034db: `PackedStorageIO` (and every `Packed*` code) now maps to the non-retriable `StorageError(2044)`. Also worth noting: no live consumer routes a `Packed*` status through `ToSegcoreErrorCode` today (the packed C-APIs hardcode `FileReadFailed`/`FileWriteFailed`), so this switch is a reserved classification for future direct-link callers.

cpp/src/common/extend_status.cpp
czs007: Same — resolved in 5a034db.

cpp/conanfile.py
czs007: Done in your 0dfa18b — thanks for shipping 2045 and bumping the pin.

cpp/src/common/extend_status.cpp
czs007: Agreed, and thanks for 0dfa18b — after reading #574 this classification makes sense: with your source-side transient tagging (Network/Timeout/Throttling/Service + the `retryable` flag on the detail), everything that legitimately deserves a retry carries a detail, so the *plain*-IOError fallback should be conservative. Verified locally on top of 0dfa18b: 772/773 tests pass (the one failure is the pre-existing `ColumnGroupTest.MemoryUsageCalculation`, unrelated — happy to file it separately).

Proposed division of labor so #574 and this PR compose instead of colliding:
- **#574 (yours): the transient half** — source tagging + `retryable` metadata + FFI plumbing.
- **#575 (this one, rebased onto #574 once it merges): the permanent half** — NoSuchKey/NoSuchBucket → `ObjectNotExist(2017)`, the AccessDenied family and SDK-judged non-retryable → 2044, plus the fine-grained segcore mapping on top of your retryable axis (keeping `DataFormatBroken(2024)`/OOM/`InvalidParameter` instead of collapsing everything into 2044/2045) and the no-default-switch drift guard.

One gap I'd like to fold into the shared translation during that rebase: MinIO-style rate limiting arrives as `S3Errors::UNKNOWN` with exception name "SlowDown" and the non-retryable flag set (this is why `IsConnectError()` matches by exception name — see `TestIsConnectError`). The type/response-code switches in #574 won't tag it, so with the conservative fallback a genuine throttle would now classify as permanent. A name-based guard on top of your translation closes it.

PR #576, opened 2026-07-02
enhance: use a global LRU cache for format metadata

jiaqizho:
This change moves format reader metadata reuse from reader-local state to a process-wide cache, so repeated readers can share expensive immutable file metadata across calls instead of warming it up again for every Reader instance. The cache is still gated by the existing reader metadata cache property, and encrypted Parquet reads with a key retriever continue to bypass metadata reuse so the decryption path stays correct.

The implementation adds a byte-bounded LRU cache for completed metadata entries and keeps the per-format singleflight behavior so concurrent opens for the same file only load metadata once. Cache keys now include the resolved filesystem identity, which keeps entries separated across different storage configurations while still allowing immutable files to be reused safely within the same process.

Each format reader now separates reusable file metadata from per-read state more clearly. Parquet, Vortex, Lance, and Iceberg can rebuild independent readers from cached metadata while applying the current projection, predicate, filesystem properties, and logical chunk settings at reader creation time. Vortex also records a real metadata cache size when manifest file size fields are not available.

### DIFF
--- a/cpp/src/format/lance/lance_table_reader.cpp
+++ b/cpp/src/format/lance/lance_table_reader.cpp
@@ -111,6 +111,10 @@ std::string LanceTableReader::MetaTrait::cache_key(const milvus_storage::api::Co
   auto cache_key = fmt::format("lance-table|path:{}|file_size:{}|footer_size:{}", file.path,
                                file.Get<uint64_t>(milvus_storage::api::kPropertyFileSize),
                                file.Get<uint64_t>(milvus_storage::api::kPropertyFooterSize));
+  auto table_version_it = file.properties.find(kLanceTableVersionProperty);
+  if (table_version_it != file.properties.end()) {
+    cache_key += fmt::format("|{}:{}", kLanceTableVersionProperty, table_version_it->second);
+  }
   auto metadata_it = file.properties.find(milvus_storage::api::kPropertyMetadata);
   if (metadata_it != file.properties.end()) {
     cache_key += "|metadata:" + metadata_it->second;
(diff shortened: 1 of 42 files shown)


cpp/src/format/vortex/vortex_format_reader.cpp
```diff
@@ -355,19 +367,22 @@ arrow::Result<VortexFormatReader::MetaTrait::MetadataPtr> VortexFormatReader::Me
...
+  auto row_count = vxfile->RowCount();
+  ARROW_ASSIGN_OR_RAISE(auto cache_size, get_metadata_cache_size(footer_size, *vxfile));
 
   auto metadata = std::make_shared<Metadata>(Metadata{
       .cache_key = std::move(key),
       .path = uri.key,
       .file_schema = std::move(file_schema),
       .row_group_infos = std::move(row_group_infos),
-      .cache_size = footer_size,
+      .cache_size = cache_size,
```
liliu-z, line 381: The global LRU charges each entry only its on-disk footer size, which undercounts the handles Vortex and Lance actually retain. Vortex sets cache_size from get_metadata_cache_size (footer size) while the payload keeps the opened VortexFile handle, fs_holder, and row_ranges; Lance sets cache_size to the manifest footer size, or a small sizeof(Metadata)+N*sizeof(RowGroupInfo) estimate when that is 0, while the payload keeps an opened BlockingDataset. Since the cache is process-wide, enabled by default, and defaults to a 1 GiB byte cap, real retained memory can substantially exceed the configured capacity and outlive the readers that created the entries. (Parquet and Iceberg, which charge footer bytes plus EstimateDeleteChargeBytes, are not affected.)

cpp/src/format/lance/lance_table_reader.cpp
```diff
@@ -204,13 +204,16 @@ arrow::Result<std::shared_ptr<LanceTableReader>> LanceTableReader::MetaTrait::cr
     return arrow::Status::Invalid("Cannot open Lance reader from metadata with null dataset");
   }
 
+  ARROW_ASSIGN_OR_RAISE(auto logical_chunk_rows,
+                        milvus_storage::api::GetValue<uint64_t>(properties, PROPERTY_READER_LOGICAL_CHUNK_ROWS));
+
   auto reader = std::make_shared<LanceTableReader>(metadata->payload.dataset, metadata->payload.fragment_id,
```
liliu-z, line 220: load_metadata opens the Lance dataset once and create_from_metadata reuses that exact captured snapshot (metadata->payload.dataset) instead of re-opening, whereas the uncached path opens the dataset fresh in open() and therefore sees the current Lance version. Now that completed metadata lives in a process-wide cache that outlives the reader which populated it, a later Reader hitting the same path/file_size/footer_size cache key reuses the stale snapshot. For Lance-native in-place version or deletion changes within a single process, reads can silently miss the new version or applied deletions until the entry is evicted or the cache is cleared.

cpp/src/format/vortex/vortex_format_reader.cpp
jiaqizho: vortex is fine, the biggest part of cache is the footer size of vortex.

cpp/src/format/vortex/vortex_format_reader.cpp
```diff
@@ -355,19 +367,22 @@ arrow::Result<VortexFormatReader::MetaTrait::MetadataPtr> VortexFormatReader::Me
   const auto memory_usage = compute_total_mem_usage(*vxfile);
   auto row_group_infos =
       create_row_group_infos(memory_usage, vxfile->RowCount(), recalc_row_ranges(row_ranges, logical_chunk_rows));
+  auto row_count = vxfile->RowCount();
+  ARROW_ASSIGN_OR_RAISE(auto cache_size, get_metadata_cache_size(footer_size, *vxfile));
```
liliu-z, line 374: `load_metadata` propagates the `IOError` from `get_metadata_cache_size` via `ARROW_ASSIGN_OR_RAISE`, so when `footer_size == 0` and `vxfile.FooterSize()` throws, an otherwise-complete metadata load is aborted. The vxfile, schema, row ranges, and row_group_infos are all already built before this line; cache_size is used only for LRU accounting and does not affect read correctness, so fall back to `cache_size = 0` on that error instead of failing the read.

PR #579, opened 2026-07-07
enhance: Configure storage runtimes and simplify Vortex writer ownership

jiaqizho:
This change cleans up the Vortex object-store writer path so the Rust side has a clearer ownership model. Instead of cloning writer-like objects around flush and close calls, the opened C++ writer now lives behind one shared inner state with explicit unopened, open, and closed states. Writes, flushes, and close all go through the same state machine, so a flush no longer accidentally drops or destroys the underlying C++ writer while the file is still being written.

The Vortex writer is also moved onto the async Vortex writer API backed by the shared bridge runtime. That keeps blocking C++ filesystem calls offloaded through the runtime, while still preserving the existing close semantics where finishing the Vortex file and closing the object-store writer are controlled separately.

This also adds a first-use runtime configuration path for the Rust bridge. C++ can now configure the shared Tokio runtime before it is initialized, and the same storage-level API wires that together with Arrow CPU and IO thread pool limits. If the Rust runtime has already been initialized or configured, the storage configuration path reports that as a warning instead of turning runtime setup into a hard failure.

Overall, the intent is to make storage runtime sizing explicit from C++, keep Lance, Iceberg, and Vortex on one shared Rust runtime, and remove the lifecycle ambiguity around Vortex object-store writes without changing the external write/flush/close behavior.

### DIFF
--- a/cpp/src/format/bridge/rust/src/rust_runtime.rs
+++ b/cpp/src/format/bridge/rust/src/rust_runtime.rs
@@ -1,0 +1,307 @@
+use std::sync::{LazyLock, Mutex};
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+struct RustRuntimeConfig {
+    worker_threads: usize,
+    max_blocking_threads: usize,
+}
+
+#[derive(Default)]
+struct RustRuntimeState {
+    config: Option<RustRuntimeConfig>,
+    initialized: bool,
+}
+
+struct RustRuntime {
+    state: Mutex<RustRuntimeState>,
+}
+
+impl RustRuntime {
+    const fn new() -> Self {
+        Self {
+            state: Mutex::new(RustRuntimeState {
+                config: None,
+                initialized: false,
+            }),
+        }
+    }
+
+    fn configure(&self, worker_threads: usize, max_blocking_threads: usize) -> anyhow::Result<()> {
+        if worker_threads == 0 {
+            anyhow::bail!("worker_threads must be greater than 0");
+        }
+        if max_blocking_threads == 0 {
+            anyhow::bail!("max_blocking_threads must be greater than 0");
+        }
+
+        let mut state = self.state.lock().unwrap();
+        if state.initialized {
+            anyhow::bail!(
+                "Rust runtime has already been initialized; call ConfigureRustRuntime before first Rust bridge use"
+            );
+        }
+        if state.config.is_some() {
+            anyhow::bail!("Rust runtime has already been configured");
+        }
+
+        state.config = Some(RustRuntimeConfig {
+            worker_threads,
+            max_blocking_threads,
+        });
+        Ok(())
+    }
+}
+
+pub(crate) fn configure_rust_runtime(
+    worker_threads: u32,
+    max_blocking_threads: u32,
+) -> anyhow::Result<()> {
+    RUST_RUNTIME.configure(worker_threads as usize, max_blocking_threads as usize)
+}
+
+static RUST_RUNTIME: RustRuntime = RustRuntime::new();
+
+fn default_runtime_thread_count() -> usize {
+    std::thread::available_parallelism()
+        .map(usize::from)
+        .unwrap_or(32)
+}
+
+/// Shared Tokio runtime for async work in the Rust bridge.
+///
+/// Lance, Iceberg, and Vortex all run through this runtime so the bridge does
+/// not create separate Tokio worker and blocking thread pools per format.
+pub(crate) static TOKIO_RT: LazyLock<tokio::runtime::Runtime> = LazyLock::new(|| {
+    let config = {
+        let mut state = RUST_RUNTIME.state.lock().unwrap();
+        state.initialized = true;
+        state.config
+    };
+
+    let mut builder = tokio::runtime::Builder::new_multi_thread();
+    builder.enable_all();
+    if let Some(config) = config {
+        builder.worker_threads(config.worker_threads);
+        builder.max_blocking_threads(config.max_blocking_threads);
+    } else {
+        let thread_count = default_runtime_thread_count();
+        builder.worker_threads(thread_count);
+        builder.max_blocking_threads(thread_count);
+    }
+    builder.build().expect("Failed to create tokio runtime")
+});
+
+#[cfg(test)]
+mod runtime_config_tests {
+    use super::*;
+    use std::sync::{
+        Arc, Condvar,
+        atomic::{AtomicUsize, Ordering},
+    };
+    use std::time::Duration;
+
+    #[derive(Default)]
+    struct RuntimeProbe {
+        running: AtomicUsize,
+        peak_running: AtomicUsize,
+        started: AtomicUsize,
+        release: Mutex<bool>,
+        cv: Condvar,
+    }
+
+    impl RuntimeProbe {
+        fn run(&self) {
+            let running = self.running.fetch_add(1, Ordering::SeqCst) + 1;
+            self.update_peak(running);
+            self.started.fetch_add(1, Ordering::SeqCst);
+            self.cv.notify_all();
+
+            let mut release = self.release.lock().unwrap();
+            while !*release {
+                release = self.cv.wait(release).unwrap();
+            }
+
+            self.running.fetch_sub(1, Ordering::SeqCst);
+        }
+
+        fn update_peak(&self, value: usize) {
+            let mut current = self.peak_running.load(Ordering::SeqCst);
+            while value > current {
+                match self.peak_running.compare_exchange_weak(
+                    current,
+                    value,
+                    Ordering::SeqCst,
+                    Ordering::SeqCst,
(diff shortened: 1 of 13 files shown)


cpp/src/format/bridge/rust/src/rust_runtime.rs
```diff
@@ -0,0 +1,307 @@
...
+
+    let mut builder = tokio::runtime::Builder::new_multi_thread();
+    builder.enable_all();
+    if let Some(config) = config {
+        builder.worker_threads(config.worker_threads);
+        builder.max_blocking_threads(config.max_blocking_threads);
+    } else {
+        let thread_count = default_runtime_thread_count();
+        builder.worker_threads(thread_count);
+        builder.max_blocking_threads(thread_count);
```
liliu-z, line 89: When ConfigureStorageRuntime is not called (or is called with a small io-thread count), the shared runtime's default branch sets max_blocking_threads to available_parallelism() (fallback 32), down from the previous 256 for Vortex (old VORTEX_TOKIO_RT) and Tokio's 512 default for Lance/Iceberg. The Vortex sync read path fans out CONCURRENCY=256 reads via buffer_unordered over spawn_blocking (filesystem_c.rs:732, :912, :941) on this same pool, so on typical hardware those 256 concurrent blocking reads now queue behind ~CPU-count blocking threads, serializing reads that were previously fully parallel. Consider a larger blocking-pool default, or decouple max_blocking_threads from worker_threads.

cpp/src/thread_pool.cpp
```diff
@@ -0,0 +1,37 @@
...
+namespace milvus_storage {
+
+void ConfigureStorageRuntime(uint32_t num_of_cpu_threads, uint32_t num_of_io_threads) {
+  if (auto error = ConfigureRustRuntime(num_of_cpu_threads, num_of_io_threads); error.has_value()) {
+    LOG_STORAGE_WARNING_ << "Failed to configure Rust runtime: " << *error;
+  }
+  if (auto status = arrow::SetCpuThreadPoolCapacity(num_of_cpu_threads); !status.ok()) {
+    LOG_STORAGE_WARNING_ << "Failed to configure Arrow CPU thread pool: " << status.ToString();
+  }
+  if (auto status = arrow::io::SetIOThreadPoolCapacity(num_of_io_threads); !status.ok()) {
```
liliu-z, line 31: num_of_io_threads is passed as both the Tokio blocking-pool size (ConfigureRustRuntime's max_blocking_threads at line 26) and the Arrow IO pool capacity (SetIOThreadPoolCapacity at line 32). These pools have different sizing needs — the Vortex read path wants a large blocking pool (~256) while the Arrow IO pool is a separate concern — so one knob cannot size both, and raising it to fix the blocking-pool regression would also inflate the Arrow IO pool. Split into distinct Tokio-blocking and Arrow-IO thread-count parameters.

cpp/src/format/bridge/rust/src/rust_runtime.rs
jiaqizho: expert

cpp/src/thread_pool.cpp
jiaqizho: expect

xiaofan-luan:
Two things we should nail down about the consumer-side calling convention for `ConfigureStorageRuntime` before this lands.

### 1. It must be called at cgo init, before any Rust bridge use

`TOKIO_RT` is a `LazyLock` and `configure_rust_runtime` bails out once the runtime is already initialized. Unlike the Arrow CPU/IO pools — which `SetCpuThreadPoolCapacity` / `SetIOThreadPoolCapacity` can resize at any time — the tokio worker/blocking thread counts are locked in at the runtime's first touch and cannot be changed afterwards. So if the first Lance/Iceberg/Vortex call happens before we configure, the runtime silently comes up with defaults (`available_parallelism()` for both worker and blocking) and the intended config is dropped.

Concretely, on the milvus side this should be invoked from the C core init path (right next to `InitCpuNum`), fed the cgroup-aware CPU count that Go already computes via `pkg/util/hardware`, and it must run before any storage read/write. Note this also needs a `loon_`-prefixed C FFI export — right now `ConfigureStorageRuntime` isn't reachable from Go and has no caller, so the sizing is plumbing-only until that's wired.

### 2. We need to prove it actually took effect — not just log a warning

Today a late / out-of-order call degrades to a single `LOG_STORAGE_WARNING_`, which nobody notices in production, and the layer quietly runs on defaults. Two requests:

- Make a late/duplicate configure a **loud, startup-visible failure** surfaced back to Go, so a mis-ordered init fails fast instead of silently falling back.
- Add an **end-to-end check through the real cgo init path** (in the spirit of `arrow_thread_limit_test`) that asserts the tokio worker/blocking parallelism actually matches what was configured — i.e. verify it worked end to end, not just that the setter returned Ok.

(Minor, and already noted above by @liliu-z: `num_of_io_threads` currently drives both tokio `max_blocking_threads` and the Arrow IO pool capacity; worth deciding whether those two really want the same number, since blocking/IO fan-out usually wants more than the core count.)

xiaofan-luan:
One more integration point, following up on the wiring above.

milvus **already** configures the process-global Arrow IO pool on its own: `SetArrowIOThreadPoolCapacity` in `internal/core/src/common/init_c.cpp` calls `arrow::io::SetIOThreadPoolCapacity(threads)` directly. Since `ConfigureStorageRuntime` also calls `arrow::io::SetIOThreadPoolCapacity` (plus the CPU pool), once this is wired we'd have **two independent code paths writing the same global Arrow pools** — last-writer-wins, order-dependent, and easy to get subtly out of sync.

Let's not run two mechanisms. Proposed direction:

- Route everything through `ConfigureStorageRuntime` as the single entry point for storage runtime sizing (tokio worker/blocking + Arrow CPU + Arrow IO), called once at cgo init with the cgroup-aware CPU count Go already computes.
- **Remove** the standalone `SetArrowIOThreadPoolCapacity` path in `init_c.cpp` so the Arrow IO pool has exactly one owner.

That keeps a single source of truth for all storage-side thread pools and avoids the two paths fighting over the same global. The genuinely new capability this PR adds over what milvus already does is the **Rust tokio** sizing — the Arrow part should consolidate onto this entry point rather than duplicate it.

jiaqizho:
> Two things we should nail down about the consumer-side calling convention for `ConfigureStorageRuntime` before this lands.
> 
> ### 1. It must be called at cgo init, before any Rust bridge use
> `TOKIO_RT` is a `LazyLock` and `configure_rust_runtime` bails out once the runtime is already initialized. Unlike the Arrow CPU/IO pools — which `SetCpuThreadPoolCapacity` / `SetIOThreadPoolCapacity` can resize at any time — the tokio worker/blocking thread counts are locked in at the runtime's first touch and cannot be changed afterwards. So if the first Lance/Iceberg/Vortex call happens before we configure, the runtime silently comes up with defaults (`available_parallelism()` for both worker and blocking) and the intended config is dropped.
> 
> Concretely, on the milvus side this should be invoked from the C core init path (right next to `InitCpuNum`), fed the cgroup-aware CPU count that Go already computes via `pkg/util/hardware`, and it must run before any storage read/write. Note this also needs a `loon_`-prefixed C FFI export — right now `ConfigureStorageRuntime` isn't reachable from Go and has no caller, so the sizing is plumbing-only until that's wired.
> 
> ### 2. We need to prove it actually took effect — not just log a warning
> Today a late / out-of-order call degrades to a single `LOG_STORAGE_WARNING_`, which nobody notices in production, and the layer quietly runs on defaults. Two requests:
> 
> * Make a late/duplicate configure a **loud, startup-visible failure** surfaced back to Go, so a mis-ordered init fails fast instead of silently falling back.
> * Add an **end-to-end check through the real cgo init path** (in the spirit of `arrow_thread_limit_test`) that asserts the tokio worker/blocking parallelism actually matches what was configured — i.e. verify it worked end to end, not just that the setter returned Ok.
> 
> (Minor, and already noted above by @liliu-z: `num_of_io_threads` currently drives both tokio `max_blocking_threads` and the Arrow IO pool capacity; worth deciding whether those two really want the same number, since blocking/IO fan-out usually wants more than the core count.)

@xiaofan-luan   

I’ve updated this PR on the milvus-storage side:
  - `ConfigureStorageRuntime` now returns `arrow::Status` and surfaces late/duplicate runtime configuration as a  failure.
  - Added the `loon_configure_storage_runtime` FFI entrypoint.
  - Added `ffi_runtime_test.c` to verify the FFI path is usable.
  - Updated the comments to clarify the one-shot initialization requirement and the CPU/IO thread mapping.

  For the Milvus-side integration:
  - The existing `SetArrowIOThreadPoolCapacity` path will be merged into `ConfigureStorageRuntime`, so storage runtime sizing has a single entrypoint.
  - The actual cgo init-path wiring will be handled in a follow-up Milvus PR.

So the remaining storage-side changes requested here have been added; the downstream Milvus wiring will be tracked separately.

PR #581, opened 2026-07-10
enhance: Add async read support across public and format readers

jiaqizho:
The read stack previously exposed mostly blocking APIs, while the parallel work that did exist was hidden inside individual reader layers. That made it difficult for callers to compose operations, overlap reads across files and column groups, or control where backend work should run. This change adds async chunk-reader creation, multi-chunk reads, and take operations from the public reader APIs down to the format layer, while keeping synchronous APIs and ready-future fallbacks available for existing callers and formats.

Instead of letting each layer create its own scheduling model, the async path now plans the complete workload once and passes those tasks through the rest of the stack. At a high level, the architecture looks like this:

```
                           Public async API
            ChunkReader::get_chunks_async / Reader::take_async / ...
                                    |
                                    v
                     Validate input and projection
                                    |
                                    v
                        Build natural tasks
                   +----------------+----------------+
                   |                                 |
                   v                                 v
             ChunkTask                          TakeTask
       file + contiguous range          column group + file
                                        + original positions
                   |                                 |
                   +----------------+----------------+
                                    |
                                    v
                    Split tasks and fan out futures
                                    |
                                    v
          ColumnGroupReader / ColumnGroupLazyReader
                                    |
                                    v
                Independent FormatReader for each task
                                    ^
                                    |
                           immutable metadata
                                    |
                              MetadataCache
                                    |
                                    v
                   +----------------+----------------+
                   |                |                |
                   v                v                v
                Parquet           Vortex          Fallback
            Arrow generator    Rust/Tokio scan     Sync call
            caller executor     FFI callback      ready future
                   |                |                |
                   +----------------+----------------+
                                    |
                                    v
                           folly::collectAll
                                    |
                   +----------------+----------------+
                   |                                 |
                   v                                 v
            Restore chunk order              Reorder take rows
            Preserve duplicates              Combine projected columns
                   |                                 |
                   +----------------+----------------+
                                    |
                                    v
                           Ordered Arrow result
```

ChunkTask groups sorted chunk indices into physically contiguous ranges within one file, while TakeTask groups logical row indices by column group and file and keeps their original positions for reconstruction. The split strategy can preserve natural tasks, split toward the requested parallelism, or fully split the work. All planned futures are created before fan-in, allowing native async backends to overlap work across files and column groups. Chunk results are mapped back by chunk index to preserve request order and duplicates, while take results are reordered within each column group before the requested columns are combined and missing schema columns are filled with nulls.

Each task uses an independent stateful format reader, so concurrent operations do not share mutable projection, predicate, or scan state. Immutable file metadata is still reused through the metadata cache, which now coordinates synchronous and asynchronous same-key loads through shared singleflight state. Async followers reuse the leader result without blocking a thread, synchronous callers avoid waiting behind an async leader when that could stall its executor, and failed loads remain uncached so later calls can retry.

Parquet runs range and take reads through Arrow's async record-batch generator. Generator startup is deferred until the Folly future is consumed, decode and materialization work inherit the caller-selected executor, and the executor adapter and file reader stay alive until Arrow completes. Vortex keeps open and scan execution on the shared Rust Tokio runtime, transfers owned handles and callback contexts across FFI, and returns results through Arrow C streams and Folly promises. Projection, output schema, predicates, ranges, and sparse row selections are configured before the native scan is submitted.

The synchronous and asynchronous paths remain separate so existing behavior stays predictable. The storage layer does not silently create a Folly executor, and both paths keep the same validation, projection, ordering, and error semantics while making reader, metadata, callback, and executor ownership explicit.


cpp/src/reader.cpp
```diff
@@ -649,6 +685,83 @@ arrow::Result<std::vector<std::shared_ptr<arrow::RecordBatch>>> ChunkReaderImpl:
...
+folly::SemiFuture<arrow::Result<std::vector<std::shared_ptr<arrow::RecordBatch>>>> ChunkReaderImpl::get_chunks_async(
+    const std::vector<int64_t>& chunk_indices, size_t parallelism) {
+  // Plan on sorted unique chunks; fan-in below restores caller order and duplicates.
+  std::vector<int64_t> unique_chunk_indices(chunk_indices.begin(), chunk_indices.end());
+  std::sort(unique_chunk_indices.begin(), unique_chunk_indices.end());
+  unique_chunk_indices.erase(std::unique(unique_chunk_indices.begin(), unique_chunk_indices.end()),
+                             unique_chunk_indices.end());
+
+  auto total_chunks = chunk_reader_->total_number_of_chunks();
+  for (auto idx : unique_chunk_indices) {
```
sparknack: Since `unique_chunk_indices` is sorted, why not just check the first and last elements?

cpp/src/reader.cpp
```diff
@@ -649,6 +685,83 @@ arrow::Result<std::vector<std::shared_ptr<arrow::RecordBatch>>> ChunkReaderImpl:
...
+  for (auto& task : all_tasks) {
+    task_chunk_lists.push_back(task.chunk_indices);
+    futures.push_back(chunk_reader_->get_chunks_async(task));
+  }
+
+  // Associate results with task chunk ids, then rebuild the original request;
+  // asynchronous completion order is intentionally irrelevant.
+  return folly::collectAll(std::move(futures))
+      .deferValue([chunk_indices, task_chunk_lists = std::move(task_chunk_lists)](
+                      auto&& all_results) mutable -> arrow::Result<std::vector<std::shared_ptr<arrow::RecordBatch>>> {
```
sparknack: `mutable` can be removed?

cpp/src/reader.cpp
```diff
@@ -1045,6 +1183,167 @@ class ReaderImpl : public Reader {
...
+  task_cg_indices.reserve(all_tasks.size());
+  task_positions.reserve(all_tasks.size());
+  futures.reserve(all_tasks.size());
+
+  // Submit the complete cross-column-group task set before waiting, allowing
+  // native async backends to overlap work across both files and column groups.
+  for (auto& task : all_tasks) {
+    task_cg_indices.push_back(task.reader_index);
+    task_positions.push_back(std::move(task.original_positions));
+    futures.push_back(readers[task.reader_index]->take_async(task));
```
liliu-z, line 1367: take_tables_async (and get_chunks_async) create one SemiFuture per task before folly::collectAll with no in-flight cap, and each task opens an independent format reader, so the number of concurrently open files and read buffers grows with the task count. Natural tasks are never merged, so this is uncapped even at the default parallelism=1 — a take whose rows scatter across N files fans out N open readers at once — and the 'all' split strategy degrades to one task per row. This is already listed under the design doc's Current Limitations (#3/#4) as an accepted tradeoff, so treat it as a bounded-scalability risk that should gate very large takes/scans, not silent breakage; a bounded in-flight window would remove the exposure.

cpp/src/format/async_tasks.cpp
```diff
@@ -0,0 +1,328 @@
...
+    }
+
+    const auto& files = column_groups[reader_index]->files;
+    std::map<uint32_t, std::vector<std::pair<int64_t, size_t>>> file_groups;
+    // Resolve each column-group logical row against manifest-visible file lengths.
+    // Keep the logical value in the task; execution converts it to file-local.
+    for (size_t pos = 0; pos < row_indices.size(); ++pos) {
+      int64_t row_index_remain = row_indices[pos];
+      bool found_file = false;
+      for (uint32_t file_index = 0; file_index < files.size(); ++file_index) {
```
liliu-z, line 143: TakeTask::Build resolves each requested row with an inner loop that rescans the file table from index 0, giving O(rows × files) planning even though row_indices are validated sorted-and-unique and file ranges are monotonic — a forward cursor over the files would make it O(rows + files). It does not actually call get_index_and_offset_of_file as originally claimed; it inlines an equivalent scan, and that same per-row scan is then repeated at execution time in the new take_async path (column_group_lazy_reader.cpp:298-306). Mirrors the pre-existing sync scan shape, so it is a missed optimization in new code rather than a regression.

cpp/src/reader.cpp
```diff
@@ -808,78 +952,72 @@ class ReaderImpl : public Reader {
...
+  [[nodiscard]] folly::SemiFuture<arrow::Result<std::shared_ptr<arrow::Table>>> take_async(
+      const std::vector<int64_t>& row_indices,
+      size_t parallelism = 1,
+      const std::shared_ptr<std::vector<std::string>>& needed_columns = nullptr) override {
+    if (row_indices.empty()) {
+      if (!schema_) {
+        return folly::makeSemiFuture(arrow::Result<std::shared_ptr<arrow::Table>>(
+            arrow::Status::Invalid("Cannot create empty table without a schema")));
       }
+      FOLLY_ARROW_ASSIGN_OR_RAISE(auto empty_table, arrow::Table::MakeEmpty(schema_));
```
liliu-z: On an empty index set both take (reader.cpp:943) and the new take_async (reader.cpp:974) return arrow::Table::MakeEmpty(schema_) — the full schema — without resolving the requested projection, so a caller passing a projection with an empty index list gets an empty table carrying every column instead of the projected subset. The non-empty path projects via resolved_columns, so the shapes diverge only when a projection is requested. Build the empty table from the projected schema in both branches.

cpp/src/reader.cpp
jiaqizho: ack, thanks

cpp/src/reader.cpp
jiaqizho: ack, and checked all.

cpp/src/reader.cpp
jiaqizho: ack

cpp/src/reader.cpp
jiaqizho: This behavior is intentional: parallelism controls task-splitting granularity in the storage layer; it is not an in-flight I/O limit. In particular, parallelism=1 means that natural tasks are not split further, not that they are serialized. Admission control, request batching, concurrency caps, and rate limiting belong to the caller, which has the workload-level context needed to choose those policies. Adding a bounded window inside storage would introduce hidden scheduling semantics and could conflict with caller-side control. We therefore keep the full fan-out behavior and document very large scattered takes/scans as an accepted limitation; callers that need a bound should batch or gate these operations externally.

cpp/src/format/async_tasks.cpp
jiaqizho: Agreed. added a FIXME to track the O(rows + files) optimization.

cpp/src/common/arrow_util.cpp
```diff
@@ -152,26 +154,55 @@ arrow::Result<std::vector<std::shared_ptr<arrow::RecordBatch>>> ConvertTableToRe
...
   std::vector<std::shared_ptr<arrow::ChunkedArray>> new_columns;
+  new_columns.reserve(table->num_columns());
   for (int i = 0; i < table->num_columns(); ++i) {
-    ARROW_ASSIGN_OR_RAISE(auto datum, arrow::compute::Take(table->column(i), index_array, options, &context));
+    const auto& column = table->column(i);
+    arrow::internal::ChunkResolver resolver(column->chunks());
+
+    // Extension arrays are intentionally unsupported because Arrow cannot create
+    // a builder for ExtensionType with MakeBuilderExactIndex().
+    ARROW_ASSIGN_OR_RAISE(auto builder, arrow::MakeBuilderExactIndex(column->type()));
```
liliu-z, line 175: CopySelectedRows was reimplemented from arrow::compute::Take to a per-column builder using arrow::MakeBuilderExactIndex, which cannot construct a builder for ExtensionType (as its own inline comment states), so any take over a column group containing an extension-typed column now returns an error where the previous compute::Take-based implementation succeeded (Arrow 17's Take supports extension arrays). This regresses every take path: parquet_format_reader.cpp:659 (sync), parquet_format_reader.cpp:994 (async), and reader.cpp:1365 (async reorder). Handle ExtensionType by building on its storage type and re-wrapping.

cpp/src/format/vortex/vortex_format_reader.cpp
```diff
@@ -737,6 +854,154 @@ arrow::Result<std::shared_ptr<arrow::RecordBatchReader>> VortexFormatReader::rea
...
 }
 
+template <typename T>
+struct VortexAsyncContext {
+  folly::Promise<arrow::Result<T>> promise;
+  // Rust writes the owned C stream here before invoking a success callback.
+  ArrowArrayStream stream;
+};
+
+static void vortex_take_async_callback(void* ctx_raw, ArrowArrayStream* /*out_stream*/, const char* error_msg) {
```
liliu-z: vortex_take_async_callback (864) and vortex_read_range_async_callback (906) are invoked by Rust across the extern "C" VortexAsyncCallback boundary but are not noexcept and run throwing operations (std::vector reserve/emplace_back, Arrow imports, RecordBatch/Table assembly). A C++ exception unwinding into the Rust extern "C" frame is undefined behavior / process abort. The sibling vortex_open_async_callback already wraps its body in try/catch and converts failures to a Status; the take and read_range callbacks should do the same.

cpp/src/format/parquet/parquet_format_reader.cpp
```diff
@@ -409,6 +434,12 @@ arrow::Status ParquetFormatReader::open() {
   return set_needed_columns(needed_columns_);
 }
 
+folly::SemiFuture<arrow::Status> ParquetFormatReader::open_async() {
+  // Keep the reader alive until the deferred open() finishes.
+  auto self = shared_from_this();
```
liliu-z, line 500: open_async() calls shared_from_this() (parquet_format_reader.cpp:439; mirrored at vortex_format_reader.cpp:534), but both readers' primary constructors are public (parquet_format_reader.h:69 and vortex_format_reader.h:80, under the class public: section — the intervening `private:` at header line 62/73 is inside the nested MetaTrait struct), so a reader created outside a shared_ptr that then calls open_async() throws std::bad_weak_ptr. Every in-repo construction goes through make_shared or the shared_ptr-returning factory, so this is latent; making the constructor private behind a shared_ptr factory removes the footgun.

cpp/src/format/parquet/parquet_format_reader.cpp
```diff
@@ -776,4 +807,194 @@ ParquetFormatReader::ParquetFormatReader(const ParquetFormatReader& other,
...
+        auto arrow_future = arrow::CollectAsyncGenerator(std::move(gen));
+        return bridge_arrow_future(
+            arrow_future.Then(
+                [row_indices = std::move(row_indices), chunk_indices = std::move(chunk_indices),
+                 unique_chunk_indices = std::move(unique_chunk_indices),
+                 row_group_infos = std::move(row_group_infos)](std::vector<std::shared_ptr<arrow::RecordBatch>> batches)
+                    -> arrow::Result<std::shared_ptr<arrow::Table>> {
+                  // Generator output is ordered by unique_chunk_indices. Record
+                  // each row group's base offset in the concatenated table.
+                  ARROW_ASSIGN_OR_RAISE(auto table, arrow::Table::FromRecordBatches(batches));
```
liliu-z, line 1040: ParquetFormatReader::take_async calls arrow::Table::FromRecordBatches(batches) with no empty-batch guard, so an empty selected row-group set makes the future fail with Arrow's 'Must pass at least one record batch or an explicit schema' error instead of yielding an empty projected result. The sibling read_with_range_async explicitly handles the empty case by returning RecordBatchReader::Make({}, projected_schema). Current callers pre-filter empty inputs so this is latent, but the two async paths should handle emptiness consistently.

cpp/src/common/arrow_util.cpp
jiaqizho: expected. I do not wish to handle `ExtensionType` specially; the `ExtensionType` type is not used in the current version.

cpp/src/format/vortex/vortex_format_reader.cpp
jiaqizho: ack

cpp/src/format/parquet/parquet_format_reader.cpp
jiaqizho: expected. There is no strict constraint enforced here regarding the format reader; changing its constructor to `private` would entail significant modifications, so a `FIXME` has been added. Note that this is not a correctness issue, as the top-level reader remains unaffected.

cpp/src/format/parquet/parquet_format_reader.cpp
jiaqizho: expected. already handle it in `reader.cpp`

cpp/include/milvus-storage/reader.h
```diff
@@ -251,6 +269,20 @@ class Reader {
...
+  /**
+   * @brief Asynchronously get a chunk reader for a specific column group.
+   *
+   * The returned future includes the format-reader open/footer work needed to
+   * initialize the chunk reader.
+   *
+   * @note The default implementation invokes get_chunk_reader() before returning
+   *       a ready future and may therefore block.
+   */
+  [[nodiscard]] virtual folly::SemiFuture<arrow::Result<std::unique_ptr<ChunkReader>>> get_chunk_reader_async(
```
liliu-z, line 333: The new async virtuals are inserted mid-class — get_chunk_reader_async between get_chunk_reader and take (line 281), get_chunks_async before get_chunks_metadata in ChunkReader, take_async before set_keyretriever — which shifts every subsequent vtable slot. They are NOT pure virtuals as the original claim states; each has a default body, so source-level compatibility with existing subclasses is preserved. The layout shift only matters for binary compatibility, and only because the default build exports C++ symbols at full visibility (-fvisibility=hidden and the version scripts are gated behind WITH_PYTHON_BINDING, OFF) and the target sets no SOVERSION, so a consumer that mixed old headers with a newer shared object would dispatch through shifted slots. Since Conan ships headers and lib as one version-pinned bundle this is unlikely in practice, but appending the new virtuals at the end of each interface preserves the vtable layout at zero cost.

jiaqizho: With the current release model, headers and shared libraries are distributed together, so vtable compatibility is not an issue. We do not support using separately versioned headers and shared libraries.

cpp/src/format/column_group_reader.cpp
```diff
@@ -659,26 +777,82 @@ arrow::Result<uint64_t> ColumnGroupReaderImpl<ReaderT>::get_chunk_rows(int64_t c
...
+
+              // A format may coalesce the range into different batch boundaries;
+              // slice it back into one result per logical chunk.
+              std::vector<std::shared_ptr<arrow::RecordBatch>> result;
+              result.reserve(chunk_infos.size());
+              size_t rbs_idx = 0;
+              size_t rbs_offset = 0;
+              for (const auto& chunk_info : chunk_infos) {
+                if (UNLIKELY(rbs_idx >= rbs.size() ||
+                             (rbs[rbs_idx]->num_rows() - rbs_offset) < chunk_info.number_of_rows)) {
```
buqian-zilliz, line 835: **[P2] Do not slice filtered Vortex output by unfiltered chunk sizes**

`create_async(..., predicate)` propagates the predicate into Vortex, so `read_with_range_async()` filters the range before this loop runs. This loop still partitions the result with each chunk's pre-filter `number_of_rows`; once any row is removed, including rows whose predicate evaluates to UNKNOWN/null, the task eventually returns `Invalid slice of record batches` instead of valid chunks. Mirror the predicate-aware synchronous path, preserve post-filter chunk boundaries or row IDs, or reject predicates for this API.

cpp/src/format/async_tasks.cpp
```diff
@@ -0,0 +1,328 @@
...
+  }
+  return tasks;
+}
+
+size_t TakeTask::SplitTraits::size(const TakeTask& task) const { return task.row_indices.size(); }
+
+bool TakeTask::SplitTraits::can_split(const TakeTask& task) const { return size(task) > 1; }
+
+TakeTask TakeTask::SplitTraits::split(TakeTask& left) const {
+  size_t mid = left.row_indices.size() / 2;
```
buqian-zilliz, line 102: **[P2] Avoid splitting take tasks inside Parquet row groups**

This midpoint split is based only on requested-row count, so rows from one Parquet row group can land in several tasks. Every task opens an independent reader, and `ParquetFormatReader::take_async()` fully decodes each touched row group before selecting rows; `parallelism=8` can therefore turn one decode into eight, while the `all` strategy can decode the same group once per requested row. Split Parquet tasks on row-group boundaries or share each row-group decode before fan-out.

cpp/src/format/vortex/vortex_format_reader.cpp
```diff
@@ -818,4 +949,177 @@ arrow::Result<uint64_t> VortexFormatReader::total_mem_usage() {
...
+  vortex_scan_collect_async(handle, &raw_ctx->stream, vortex_take_async_callback, static_cast<void*>(raw_ctx));
+  return semi_future;
+}
+
+folly::SemiFuture<arrow::Result<std::shared_ptr<arrow::RecordBatchReader>>> VortexFormatReader::read_with_range_async(
+    uint64_t start_offset, uint64_t end_offset) {
+  assert(vxfile_);
+
+  // Match the default window so builder creation reuses the open file instead of synchronously reopening it.
+  FOLLY_ARROW_ASSIGN_OR_RAISE(auto scan_builder, vxfile_->CreateScanBuilder(kSmallCoalescingWindow));
```
buqian-zilliz: **[P2] Preserve range-read coalescing in the async Vortex path**

`read_with_range()` uses the 8 MiB `kLargeCoalescingWindow`, but the new async path always builds the scan with the 1 MiB window. Reusing the default view avoids the current synchronous alternate-view open, but large contiguous chunk tasks can now issue substantially more object-store reads solely because the async API was selected. Open or cache the large-window view asynchronously, or otherwise preserve the synchronous path's I/O granularity.

docs/async-read-design.md
```diff
@@ -0,0 +1,637 @@
...
+The default implementations call the synchronous operation immediately and wrap
+its result with `folly::makeSemiFuture(...)`. The wrapper itself does not select
+`ThreadPoolHolder`, call `folly::via()`, or choose another executor. Any runtime
+used inside the synchronous operation belongs to that backend's sync path.
+
+Current format behavior is:
+
+| Format | Open/create | Range/take |
+|---|---|---|
+| Parquet | Synchronous ready-future wrapper | Native async Arrow generator path |
```
buqian-zilliz: **[P3] Refresh the documented open and metadata model**

The format table still says Parquet and Vortex open/create are synchronous ready-future wrappers, and the later Open and Metadata Loading section repeats that no opens are deferred or fanned out. Current head defers Parquet open, runs Vortex open on Tokio, routes `create_reader_async()`, and composes async metadata opens, so this gives callers the wrong blocking and executor model. Update both sections to match the implementation.

cpp/src/format/column_group_reader.cpp
jiaqizho: top level reader already avoid this case.

cpp/src/format/async_tasks.cpp
jiaqizho: The design here is based on the split strategy, rather than mandating that every strategy take row groups into account; otherwise, the changes made here would violate the design of the strategy itself.

cpp/src/format/vortex/vortex_format_reader.cpp
jiaqizho: ack

docs/async-read-design.md
jiaqizho: ack

PR #584, opened 2026-07-16
feat: add MergeResolver and idempotent append for concurrency-safe manifest commits

xiaofan-luan:
## What / why

StorageV3 manifest commits default to `OverwriteResolver`, which rebuilds the committed
manifest from the transaction's **stale pinned `read_version`** and silently drops any
commit another task made to the same segment's manifest in between — a lost update. The
version-newer acceptance never catches it, because an overwrite commit's version is always
`latest + 1`.

This is the root mechanism behind milvus-io/milvus#51376 (schema-bump / in-place commits
dropping a concurrently-built index). A sealed segment's manifest has many concurrent,
independent writers by design (bump / function materialization, bloom / bm25, text / json
index, L0 deletes), so this is a structural race, not an isolated bug.

## What this PR does (storage-layer foundation)

- **`MergeResolver`** (new default write mode): applies updates onto the **latest** committed
  manifest, so a concurrent commit made after `read_version` was pinned is preserved (layered)
  instead of overwritten. `requireLatest() == true`.
- **`InsertResolver`**: explicit create / first-write semantics for **birth** commits
  (a brand-new / not-yet-announced manifest), so every remaining non-merge write site is an
  auditable "this is a create". `OVERWRITE` is retained but deprecated.
- **Idempotent `dedup`** in `applyUpdates` (used only by `MergeResolver`): skips appended
  files / delta logs / lob files / column groups already present in the base, so a commit
  retry that re-applies the same updates does not duplicate them. Identity is the
  **filename**, not the full path — `Manifest::toRelativePaths` / `ToAbsolutePaths` rewrite
  the directory prefix on the relative⇄absolute round-trip, so full-path equality is not
  stable across a commit.
- **FFI**: `LOON_TRANSACTION_RESOLVE_MERGE = 1`, `LOON_TRANSACTION_RESOLVE_INSERT = 3`; the
  resolver switch now defaults to `MERGE`.

## Tests

`MergeResolver` preserves concurrent file / delta-log / column-group / stats-by-key commits;
Commit-level dedup for duplicate file / delta-log. Full transaction suite green (34 passed,
1 skipped).

**Not covered here:** the threaded `AlreadyExists → retry` path requires remote storage with
conditional write (CAS) and is only exercised by the existing remote-only concurrency test,
not the local-fs suite.

## Follow-up (milvus side)

This is only the storage-layer mechanism. The actual fix for #51376 needs the milvus repo to
(a) flip non-birth manifest write sites (bump / stats / L0) to `MERGE` and mark birth sites
`INSERT`, and (b) rework the datacoord acceptance to adopt-if-version-newer and drop the
`errStatsResultStale` / `BaseManifest` machinery (which would otherwise discard
correctly-merged results). Those land in a separate milvus PR.

Relates to milvus-io/milvus#51376

### DIFF
--- a/cpp/include/milvus-storage/transaction/transaction.h
+++ b/cpp/include/milvus-storage/transaction/transaction.h
@@ -144,10 +144,18 @@ class Resolver {
  *
  * @param manifest Base manifest to apply updates to
  * @param updates Updates to apply
+ * @param dedup When true, appended files / delta logs / lob files / column groups whose
+ *        identity (storage path, or column set for column groups) already exists in the
+ *        base manifest are skipped instead of appended. This makes re-application of the
+ *        same updates onto a base that already contains them idempotent — required when a
+ *        resolver applies updates onto the latest manifest (MergeResolver), which under
+ *        commit retries may already carry a prior attempt's writes. Defaults to false so
+ *        the create/overwrite paths keep their existing append-always semantics.
  * @return New manifest with updates applied, or error status
  */
 arrow::Result<std::shared_ptr<Manifest>> applyUpdates(const std::shared_ptr<Manifest>& manifest,
-                                                      const Updates& updates);
+                                                      const Updates& updates,
+                                                      bool dedup = false);
 
 // ==================== Helper Resolver Functions ====================
 
@@ -171,6 +179,30 @@ extern const Resolver& OverwriteResolver;
  * version drift, since the resolver only inspects manifests when versions match.
  */
 extern const Resolver& FailResolver;
+
+/**
+ * @brief Resolver for create / first-write (birth) commits.
+ *
+ * Applies updates onto the read manifest without reconciling against concurrent changes.
+ * Use ONLY when the target manifest is brand-new / not yet announced (read_version == 0 or
+ * a freshly-created, task-local manifest), so no concurrent writer's commit can be lost.
+ * Same behavior as the deprecated OverwriteResolver, but named to make birth intent explicit.
+ *
+ * requireLatest() == false.
+ */
+extern const Resolver& InsertResolver;
+
+/**
+ * @brief Default resolver for in-place writes onto a live (announced) segment's manifest.
+ *
+ * Applies updates onto the LATEST committed manifest, so a concurrent commit made after this
+ * transaction pinned its read_version is preserved (layered) instead of being silently
+ * reverted. Re-application is idempotent: appended files / delta logs / lob files / column
+ * groups already present in the latest manifest are skipped, making commit retries safe.
+ *
+ * requireLatest() == true — Commit() loads the latest manifest so it can be merged into.
+ */
+extern const Resolver& MergeResolver;
 
 class Transaction {
   public:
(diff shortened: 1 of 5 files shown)


jiaqizho:
u can just revert this pr: https://github.com/milvus-io/milvus-storage/pull/499

cpp/include/milvus-storage/transaction/transaction.h
```diff
@@ -172,6 +180,30 @@ extern const Resolver& OverwriteResolver;
...
+ * @brief Resolver for create / first-write (birth) commits.
+ *
+ * Applies updates onto the read manifest without reconciling against concurrent changes.
+ * Use ONLY when the target manifest is brand-new / not yet announced (read_version == 0 or
+ * a freshly-created, task-local manifest), so no concurrent writer's commit can be lost.
+ * Same behavior as the deprecated OverwriteResolver, but named to make birth intent explicit.
+ *
+ * requireLatest() == false.
+ */
+extern const Resolver& InsertResolver;
```
jiaqizho, line 193: I don’t see a behavioral difference between InsertResolver and OverwriteResolver. Both apply updates to read_manifest and ignore latest_manifest.

InsertResolver also does not enforce create-only semantics. It does not check that both read_version and latest_version are 0.

We still need OverwriteResolver for intentional overwrite cases. For example, if we want to create version 12 based on version 10 and deliberately discard version 11, that operation should remain OVERWRITE, not INSERT.

Without create-only validation, INSERT is just an alias for OVERWRITE and may be misleading.

cpp/src/transaction/transaction.cpp
```diff
@@ -80,8 +80,37 @@ const std::vector<LobFileInfo>& Updates::GetAddedLobFiles() const { return added
 
 // ==================== Helper Functions ====================
 
+namespace {
```
jiaqizho, line 83: I do not think these helper functions are the right abstraction. pathBaseName and fileWithPathPresent hide the fact that entries are compared only by filename. If this comparison is really needed, it should be explicit and type-constrained, for example:
```
  template <typename T>
    requires requires(const T& value) {
      { value.path } -> std::convertible_to<std::string_view>;
    }
  static inline bool has_same_filename(const T& left, const T& right) {
    return std::filesystem::path(left.path).filename() ==
           std::filesystem::path(right.path).filename();
  }
```

More importantly, **the same filename should not be treated as the identity of a manifest entry.**

For external tables, one physical file may be represented by multiple ColumnGroupFile entries with the same path but different start_index and end_index ranges. Deduplicating them by filename could silently drop valid entries.

columnGroupPresent has a similar problem. It checks only columns.front(), so different column groups with one overlapping column may incorrectly be treated as identical and skipped.

If retry idempotency is required, the caller or update type should provide a stable logical identity. The storage layer should not infer identity from a filename or the first column. The manifest model should remain flexible and allow entries with the same filename when they represent different logical data.

jiaqizho (changes requested):

xiaofan-luan:
Closing for now — the MergeResolver/dedup approach needs rework based on @jiaqizho's review (INSERT needs real create-only semantics vs. keeping OVERWRITE; dedup identity must be a stable logical id, not filename, to avoid dropping external-table ColumnGroupFile ranges). Will re-open a revised PR.

PR #585, opened 2026-07-16
enhance: estimate Lance column and chunk memory from footer metadata

jiaqizho:
Lance row groups currently report memory_size as a placeholder value of 1, which gives cache and scheduling callers no useful idea of how much decoded memory a chunk may occupy. We need a best-effort estimate that also works for externally referenced Lance files, without relying on extra information being recorded when the file is written.

This change derives the estimate from the existing schema and footer metadata. Fixed-width leaf columns use the logical row count multiplied by their value width. Variable-width and nested columns use Lance page metadata: completed pages follow the 8 MiB writer page target, the final page is prorated from its row count, and encoded page bytes are only used as a lower bound or single-page fallback instead of being mistaken for decoded memory. Deletions are handled by applying the live-to-physical row ratio once, and physical columns belonging to nested fields are grouped back into their top-level schema columns.

The Rust bridge now provides estimates for each top-level column and keeps the fragment-level API as the saturating sum of those values. LanceTableReader distributes the fragment estimate across logical chunks according to their row ranges, so the chunk values add back up to the fragment total and remain consistent across projections and cached reader creation. Estimation stays best effort on the C++ side, with a zero fallback when metadata cannot be estimated, including unsupported legacy files.

The estimator resolves both normal dataset files and files referenced through external base paths. It only reads schema, column, and page metadata from the footer and never reads or decompresses data pages. It also relies on Lance's normalized ColumnInfo and PageInfo representation for storage formats 2.1 through 2.3, avoiding encoding-specific logic that would need to be maintained whenever the file format changes. The current Lance version does not expose get_file_metadata publicly, so the implementation temporarily uses the public read_all_metadata path without sharing Lance's metadata cache; after upgrading to a Lance version that exposes cached file metadata, this can switch to cached metadata and remove the duplicate footer I/O.


cpp/src/format/bridge/rust/src/lance_bridge.cpp
```diff
@@ -104,6 +104,14 @@ uint64_t BlockingDataset::GetFragmentRowCount(uint64_t fragment_id) const {
   }
 }
 
+uint64_t BlockingDataset::EstimateFragmentMemory(uint64_t fragment_id) const {
+  try {
+    return ffi::estimate_fragment_memory(*impl_, fragment_id);
+  } catch (const rust::cxxbridge1::Error&) {
+    return 0;
```
sparknack, line 111: bug: Returning 0 for every estimator error makes non-empty chunks zero-cost. Reader::load_internal then keeps preloading because memory_used_ does not increase, so legacy or transient footer failures can schedule the whole fragment and exceed the memory limit. Propagate an unknown/error state or use a conservative non-zero fallback, and test this through the top-level Reader.

cpp/src/format/bridge/rust/src/lance_memory_estimator.rs
```diff
@@ -0,0 +1,349 @@
...
+}
+
+/// Return the value-buffer estimate for a directly fixed-width leaf.
+///
+/// `None` means the field must use the page heuristic. Structurally nested,
+/// variable-width, and dictionary fields take that path. Nullable validity
+/// buffers are deliberately omitted because this estimator only targets a
+/// coarse value-memory estimate.
+fn fixed_column_memory(field: &Field, rows: u64) -> Option<u64> {
+    if !field.is_leaf() {
```
sparknack, line 91: bug: FixedSizeList never reaches fixed_data_type_memory because Lance Field::is_leaf() is false for lists. The current test passes through the single-page disk fallback with compression disabled. Handle FixedSizeList before this guard and add a compressed or multi-page case.

cpp/src/format/bridge/rust/src/lance_memory_estimator.rs
```diff
@@ -0,0 +1,349 @@
...
+/// Current Lance manifests persist this mapping in `column_indices`. Older
+/// files derive the same mapping from the file schema's pre-order field list.
+/// A negative persisted index means the field has no physical file column.
+fn file_field_columns(data_file: &DataFile, metadata: &CachedFileMetadata) -> Vec<(i32, u32)> {
+    if !data_file.column_indices.is_empty() {
+        return data_file
+            .fields
+            .iter()
+            .copied()
+            .zip(data_file.column_indices.iter().copied())
```
sparknack, line 199: bug: zip drops extra column_indices. Lance permits one field to span multiple physical columns; unmapped indices are continuations of the previous field. Include those continuation columns or these fields are underestimated.

sparknack (changes requested):
Findings:
- Zero or unknown estimates can bypass the reader memory budget.
- FixedSizeList does not use the exact fixed-width path.
- Continuation physical columns are omitted from the estimate.

Please address these behavioral gaps before merge.

cpp/src/format/bridge/rust/src/lance_memory_estimator.rs
jiaqizho: I don't think this applies to Lance 7.

  `Field::is_leaf()` is based on `children.is_empty()`, and the Arrow-to-Lance field conversion only creates a
  child for `FixedSizeList` when its item type is `Struct`. Therefore, the `FixedSizeList<Float32>` used by this
  test is a leaf and does reach `fixed_data_type_memory`.

  Also, `rle-threshold=1.0` does not disable compression. For the constant Float32 data in this test, it allows
  RLE to be selected; `bss=off` only disables BSS. If the estimator were taking the single-page disk fallback,
  the compressed size would be much smaller than the Arrow logical memory size, so the exact equality assertion
  would fail. The test passes in CI.

  `FixedSizeList<Struct>` is non-leaf, but its Struct element cannot use the direct fixed-width calculation
  anyway, so using the page heuristic for that case is appropriate.

cpp/src/format/bridge/rust/src/lance_bridge.cpp
jiaqizho: Returning 0 is expected. Caller should handle it.

cpp/src/format/bridge/rust/src/lance_memory_estimator.rs
jiaqizho: ack.

cpp/src/format/bridge/rust/src/lance_memory_estimator.rs
```diff
@@ -0,0 +1,391 @@
...
+        .open_file_with_priority(&path, 0, &data_file.file_size_bytes)
+        .await?;
+
+    // FIXME: `FileFragment::get_file_metadata` is not currently a public API.
+    // `FileReader::read_all_metadata` is public, but calling it directly cannot
+    // reuse Lance's metadata cache and may cause duplicate footer I/O. After
+    // https://github.com/lance-format/lance/pull/7820 is available in the Lance
+    // version used here, use `get_file_metadata` to reuse cached metadata and
+    // remove this extra I/O.
+    FileReader::read_all_metadata(&file_scheduler).await
```
liliu-z, line 149: The estimator calls `FileReader::read_all_metadata` (which reads the file footer) while `estimate_fragment_columns` holds the per-dataset `fragment_open_mutex`. This footer read is duplicative — the in-code FIXME notes it bypasses Lance's metadata cache — and issuing it inside the shared critical section extends lock-hold time, serializing concurrent fragment opens/estimates that share the same dataset instance. Move the metadata read out of the mutex, or switch to the cached `get_file_metadata` path the FIXME already plans.

jiaqizho: So i added a FIXME...Can't reslove this problem in lance 7.0.0 with clear plan.

PR #586, opened 2026-07-17
enhance: add per-column memory size estimates for chunks

jiaqizho:
ChunkReader currently reports only the total estimated decoded-memory size for each chunk, which makes it difficult for callers to understand how much each top-level column contributes when planning reads and managing memory. Add metadata-only APIs that expose per-column estimates without reading data pages, and keep the results based on the complete file schema instead of the active projection.

Carry column memory sizes through RowGroupInfo, format reader metadata, ColumnGroupReader, and the public ChunkReader interface. Provide both field-name lookup and an all-column view, preserve full file-schema ordering, reject missing or ambiguous field names, and return NotImplemented when the underlying format does not provide column-level metadata. Keep every supported chunk’s column estimates consistent with its existing total memory estimate.

For Parquet, aggregate physical leaf-column sizes under each top-level Arrow field and use those values as weights against the existing row-group memory estimate, including nested fields and files that rely on standard footer metadata. For Vortex, retain the top-level uncompressed-size statistics and distribute them across logical row groups while preserving zero-size behavior when statistics are unavailable.

For Iceberg, reuse the underlying Parquet estimates and scale both total and per-column sizes according to the rows that remain after positional deletes. Also handle partially selected row groups by scaling their total size using the overlapping row count and redistributing the column estimates so partial chunks preserve the same size invariant.

### DIFF
--- a/cpp/include/milvus-storage/reader.h
+++ b/cpp/include/milvus-storage/reader.h
@@ -149,11 +149,10 @@ class ChunkReader {
  * including column groups, data layout, and metadata, providing optimized access
  * patterns for analytical workloads.
  *
- * Some formats validate column-level memory-estimation metadata while opening
- * their underlying format readers. If that metadata is unavailable,
- * get_record_batch_reader(), get_chunk_reader(), or take() may return
- * arrow::Status::NotImplemented. Callers must handle this status; no usable
- * RecordBatchReader, ChunkReader, or Table is returned in that case.
+ * Memory-estimation metadata is optional and is not required for normal data
+ * reads. When it is unavailable, reader creation and data access remain usable;
+ * only the chunk-size estimation APIs return arrow::Status::NotImplemented.
+ * The underlying statistics failure is not exposed through those APIs.
  */
 class Reader {
   public:
--- a/cpp/include/milvus-storage/reader.h
+++ b/cpp/include/milvus-storage/reader.h
@@ -90,10 +90,13 @@ class ChunkReader {
   /**
    * @brief Returns the estimated decoded-memory size of every chunk.
    *
-   * The result is indexed by chunk. Estimates are derived from format metadata
-   * and may be zero when a chunk has no live rows or its known decoded-memory
-   * size is zero. Unavailable estimates are returned as an error, typically
-   * arrow::Status::NotImplemented, rather than being represented by zero.
+   * The result is indexed by chunk. When per-column metadata is available, each
+   * estimate is the sum of the logical column group's fields, independently of
+   * the active projection; fields present only in a physical file are excluded.
+   * Formats without per-column metadata retain their aggregate estimate.
+   * Estimates may be zero when a chunk has no live rows or its known
+   * decoded-memory size is zero. Unavailable estimates are returned as an
+   * error, typically arrow::Status::NotImplemented, rather than as zero.
    *
    * @return One estimated size in bytes per chunk, or an error status.
    */
@@ -102,11 +105,12 @@ class ChunkReader {
   /**
    * @brief Returns the estimated decoded-memory size of a top-level field in every chunk.
    *
-   * The field is resolved against the complete file schema, independently of
-   * the active projection. A successful zero value is a known zero estimate;
-   * unavailable estimates are returned as an error rather than as zero.
+   * The field is resolved against the logical column group, independently of
+   * the active projection. A field absent from a particular physical file has
+   * a known zero estimate for that chunk. Unavailable statistics are returned
+   * as an error rather than as zero.
    *
-   * @param field_name Name of a unique top-level field in the complete file schema.
+   * @param field_name Name of a unique top-level field in the column group.
    * @return One estimated size in bytes per chunk, or an error status.
    */
   [[nodiscard]] virtual arrow::Result<std::vector<uint64_t>> get_chunk_column_estimated_size(
@@ -115,7 +119,7 @@ class ChunkReader {
   /**
    * @brief Returns estimated decoded-memory sizes for every top-level field and chunk.
    *
-   * The outer dimension follows complete file-schema order and the inner
+   * The outer dimension follows logical column-group order and the inner
    * dimension is indexed by chunk: result[column_index][chunk_index]. Results
    * are independent of the active projection. Individual estimates may be
    * zero only when the corresponding estimate is known to be zero. Unavailable
--- a/cpp/include/milvus-storage/reader.h
+++ b/cpp/include/milvus-storage/reader.h
@@ -88,9 +88,52 @@ class ChunkReader {
       const std::vector<int64_t>& chunk_indices, size_t parallelism = 1) = 0;
 
   /**
-   * @brief Retrieves the metadata of chunks
+   * @brief Returns the estimated decoded-memory size of every chunk.
+   *
+   * The result is indexed by chunk. Estimates are derived from format metadata
+   * and may be zero when a chunk has no live rows or its known decoded-memory
+   * size is zero. Unavailable estimates are returned as an error, typically
+   * arrow::Status::NotImplemented, rather than being represented by zero.
+   *
+   * @return One estimated size in bytes per chunk, or an error status.
    */
-  [[nodiscard]] virtual arrow::Result<std::vector<uint64_t>> get_chunk_size() = 0;
+  [[nodiscard]] virtual arrow::Result<std::vector<uint64_t>> get_chunk_estimated_size() = 0;
+
+  /**
+   * @brief Returns the estimated decoded-memory size of a top-level field in every chunk.
+   *
+   * The field is resolved against the complete file schema, independently of
+   * the active projection. A successful zero value is a known zero estimate;
+   * unavailable estimates are returned as an error rather than as zero.
+   *
+   * @param field_name Name of a unique top-level field in the complete file schema.
+   * @return One estimated size in bytes per chunk, or an error status.
+   */
+  [[nodiscard]] virtual arrow::Result<std::vector<uint64_t>> get_chunk_column_estimated_size(
+      const std::string& field_name) = 0;
+
+  /**
+   * @brief Returns estimated decoded-memory sizes for every top-level field and chunk.
+   *
+   * The outer dimension follows complete file-schema order and the inner
+   * dimension is indexed by chunk: result[column_index][chunk_index]. Results
+   * are independent of the active projection. Individual estimates may be
+   * zero only when the corresponding estimate is known to be zero. Unavailable
+   * estimates are returned as an error rather than as zero.
+   *
+   * @return Estimated sizes in bytes arranged as [column][chunk], or an error status.
+   */
+  [[nodiscard]] virtual arrow::Result<std::vector<std::vector<uint64_t>>> get_chunk_column_estimated_size() = 0;
+
+  /**
+   * @brief Returns the exact logical row count of every chunk.
+   *
+   * Unlike the estimated-size APIs, failures are returned as an error status
+   * and are never converted to a zero fallback. A returned zero is an exact
+   * row count for an empty logical chunk.
+   *
+   * @return One exact row count per chunk, or an error status.
+   */
   [[nodiscard]] virtual arrow::Result<std::vector<uint64_t>> get_chunk_rows() = 0;
 };
 
@@ -105,6 +148,12 @@ class ChunkReader {
  * This reader leverages the manifest system to understand the dataset structure,
  * including column groups, data layout, and metadata, providing optimized access
  * patterns for analytical workloads.
+ *
+ * Some formats validate column-level memory-estimation metadata while opening
+ * their underlying format readers. If that metadata is unavailable,
+ * get_record_batch_reader(), get_chunk_reader(), or take() may return
+ * arrow::Status::NotImplemented. Callers must handle this status; no usable
+ * RecordBatchReader, ChunkReader, or Table is returned in that case.
  */
 class Reader {
   public:
--- a/cpp/src/format/bridge/rust/src/lance_bridge.cpp
+++ b/cpp/src/format/bridge/rust/src/lance_bridge.cpp
@@ -104,6 +104,20 @@ uint64_t BlockingDataset::GetFragmentRowCount(uint64_t fragment_id) const {
   }
 }
 
+arrow::Result<std::vector<uint64_t>> BlockingDataset::EstimateFragmentColumnMemory(uint64_t fragment_id) const {
+  try {
+    auto estimates = ffi::estimate_fragment_column_memory(*impl_, fragment_id);
+    std::vector<uint64_t> memory_sizes;
+    memory_sizes.reserve(estimates.size());
+    for (const auto& estimate : estimates) {
+      memory_sizes.push_back(estimate.memory_size);
+    }
+    return memory_sizes;
+  } catch (const rust::cxxbridge1::Error& e) {
+    return arrow::Status::NotImplemented("Lance column memory size estimation is not available: ", e.what());
+  }
+}
+
 uint64_t BlockingDataset::EstimateFragmentMemory(uint64_t fragment_id) const {
   try {
     return ffi::estimate_fragment_memory(*impl_, fragment_id);
--- a/cpp/src/format/column_group_reader.cpp
+++ b/cpp/src/format/column_group_reader.cpp
@@ -15,8 +15,11 @@
 #include "milvus-storage/format/column_group_reader.h"
 
 #include <numeric>
+#include <limits>
 #include <unordered_map>
+#include <unordered_set>
 #include <future>
+#include <string_view>
 #include <vector>
 #include <algorithm>
 
@@ -38,6 +41,7 @@
 #include "milvus-storage/common/metadata.h"
 #include "milvus-storage/common/arrow_util.h"
 #include "milvus-storage/common/constants.h"
+#include "milvus-storage/common/log.h"
 #include "milvus-storage/common/macro.h"  // for UNLIKELY
 #include "milvus-storage/common/fiu_local.h"
 #include "milvus-storage/format/iceberg/iceberg_format_reader.h"
@@ -49,6 +53,33 @@ namespace milvus_storage::api {
 
 using milvus_storage::RowGroupInfo;
 using ChunkRBMapResult = arrow::Result<std::unordered_map<int64_t, std::shared_ptr<arrow::RecordBatch>>>;
+using ColumnMemorySizes = std::unordered_map<std::string, uint64_t>;
+using ColumnMemorySizesPtr = std::shared_ptr<const ColumnMemorySizes>;
+
+namespace {
+
+arrow::Result<ColumnMemorySizesPtr> BuildColumnMemorySizes(const arrow::Schema& file_schema,
+                                                           const std::vector<uint64_t>& sizes) {
+  if (sizes.empty()) {
+    return ColumnMemorySizesPtr{};
+  }
+  if (sizes.size() != static_cast<size_t>(file_schema.num_fields())) {
+    return arrow::Status::Invalid("Column memory size count does not match the file schema");
+  }
+
+  auto column_memory_sizes = std::make_shared<ColumnMemorySizes>();
+  column_memory_sizes->reserve(sizes.size());
+  for (int field_index = 0; field_index < file_schema.num_fields(); ++field_index) {
+    const auto& field_name = file_schema.field(field_index)->name();
+    if (!column_memory_sizes->emplace(field_name, sizes[field_index]).second) {
+      return arrow::Status::Invalid("Duplicate field name in file schema: ", field_name);
+    }
+  }
+  return std::static_pointer_cast<const ColumnMemorySizes>(column_memory_sizes);
+}
+
+}  // namespace
+
 struct ChunkInfo {
   public:
   size_t file_index;               // current chunk belong which file
@@ -58,9 +89,9 @@ struct ChunkInfo {
   size_t row_group_index_in_file;  // the index of this row group in its file
   size_t global_row_end;           // the ending row offset of this row group in the whole chunk reader
   uint64_t avg_memory_size;        // average memory usage of this row group
-  // FIXME(jiaqizho): Use shared_ptr for immutable column metadata so cached readers and full chunks can share it;
-  // only partial row groups should materialize scaled copies.
-  std::vector<uint64_t> column_memory_sizes;
+  // Keyed by the current file's field names so chunks from files with different
+  // physical column orders can share immutable metadata without normalization.
+  ColumnMemorySizesPtr column_memory_sizes;
   // False means avg_memory_size is only a placeholder and column_memory_sizes is unavailable.
   bool memory_size_available;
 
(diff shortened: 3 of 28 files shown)


cpp/src/format/bridge/rust/src/lance_bridge.cpp
```diff
@@ -104,6 +104,20 @@ uint64_t BlockingDataset::GetFragmentRowCount(uint64_t fragment_id) const {
...
+  try {
+    auto estimates = ffi::estimate_fragment_column_memory(*impl_, fragment_id);
+    std::vector<uint64_t> memory_sizes;
+    memory_sizes.reserve(estimates.size());
+    for (const auto& estimate : estimates) {
+      memory_sizes.push_back(estimate.memory_size);
+    }
+    return memory_sizes;
+  } catch (const rust::cxxbridge1::Error& e) {
+    return arrow::Status::NotImplemented("Lance column memory size estimation is not available: ", e.what());
```
liliu-z, line 117: BlockingDataset::EstimateFragmentColumnMemory catches every rust::cxxbridge1::Error and returns arrow::Status::NotImplemented. The underlying Rust estimator does real object-store I/O while reading footer/column metadata (load_file_metadata awaits object_store(), open_file_with_priority(), and read_all_metadata()), so a transient I/O fault (e.g. an S3 hiccup) is collapsed into NotImplemented. Because LanceTableReader::open()/load_metadata propagate this status and callers treat NotImplemented as an unsupported, non-retryable result, a retryable storage error becomes a permanent read failure. The sibling EstimateFragmentMemory swallows the same errors to 0, so only this new path hard-fails. Distinguish 'metadata genuinely unavailable' from an I/O error and map the latter to the I/O/retryable status.

xiaofan-luan:
Thanks for adding per-column memory size estimates — the backend coverage (Parquet / Vortex / Lance, plus Iceberg positional-delete scaling) is nice.

One thing I'd suggest addressing before merge: the **missing-stats path currently fails instead of degrading**. When per-column statistics are unavailable the code returns `arrow::Status::NotImplemented` — e.g. `get_column_uncompressed_sizes` ("Vortex column memory size statistics are not available"), the generic "Column memory size metadata is not available for this format", and the Lance estimator ("Lance column memory size estimation is not available").

Because these are evaluated on the reader open / metadata path (`ARROW_ASSIGN_OR_RAISE`), a file that has no per-column stats — e.g. an empty Vortex file — now **fails to open**, whereas today it opens fine: `get_chunk_size()` already degrades gracefully (Parquet falls back to the row-group `total_byte_size()`, Vortex returns 0). I believe this is exactly what the two failing unit tests are hitting, and it would be a loading regression for existing readers that don't need column-level sizing.

Since this is a best-effort sizing hint, could we **degrade rather than fail**? Concretely:

- When per-column stats are missing, don't return `NotImplemented`. Instead derive a safe/conservative value from the row-group size, which is always available — e.g. split the row-group `total_byte_size()` / `get_chunk_size()` evenly across the columns (`row_group_size / num_columns`), or attribute it as a conservative per-column bound.
- Keep the accurate per-column stats as the primary path; the even-split is only the fallback.
- That way the reader always opens, callers always get a usable (if coarse) estimate, and the behavior matches the existing graceful degradation of `get_chunk_size()`.

This keeps the new API non-blocking for consumers that don't require column-level sizing, and it lets the size estimate stay best-effort (a coarse over-estimate is fine for e.g. cache reservation; a hard failure is not). Happy to help wire the fallback if useful.

xiaofan-luan (changes requested):
**Requesting changes.** The core algorithm here is solid — the Parquet nested-leaf → top-level aggregation, the uint128 + overflow-checked `DistributeMemorySizes`, and the Iceberg positional-delete / partial-row-group scaling all look right, and the full-schema, projection-independent result avoids the single-column-projection normalization issue in the alternative approach. My concern is that per-column estimation is wired as a hard prerequisite of reading, which turns a best-effort hint into correctness/availability regressions. Three blocking issues plus one compatibility issue below.

### [P1] Vortex: missing column stats fails normal reads
Vortex metadata load unconditionally requires `UncompressedSizeInBytes`; when stats are disabled, missing, or the file is empty, `open()` returns `NotImplemented`. CI already confirms this:
```
VortexBasicTest.TestEmptyWriteRead/V1  failed
VortexBasicTest.TestEmptyWriteRead/V2  failed
NotImplemented: Vortex column memory size statistics are not available
```
Column-level estimation is a best-effort capability and should not affect existing read paths (`get_chunk()`, `take()`, …). Please open the reader normally and only surface `NotImplemented` (or a fallback) when the column-estimation API is actually called.

### [P1] Lance: schema evolution makes old fragments unreadable
The Rust estimator returns results in dataset-schema order and carries `field_id`, but the C++ bridge (`lance_bridge.cpp`) drops `field_id` and keeps only the size, then `lance_table_reader.cpp` matches by vector length against the fragment schema. After a column is added:
```
dataset schema:   [a, b, c]
old fragment:     [a, b]
estimator result: [a_size, b_size, 0]
```
`3 != 2` → `Invalid` → the old fragment can't open. Even with an equal count but a changed order, sizes get attributed to the wrong columns. Please keep `field_id` and map by field ID onto the fragment/file schema instead of matching by position/length.

### [P1] Lance: transient I/O errors mis-mapped to NotImplemented
`estimate_fragment_column_memory` reads object-storage footer/page metadata, but the bridge catches every `cxxbridge1::Error` and returns `NotImplemented`. So retryable I/O failures — S3 timeout, network jitter, credential refresh — all become "format not supported", and because the estimator runs inside `open()`, they block normal reads entirely. Suggestions:
- Return `NotImplemented` only for a confirmed-unsupported legacy format.
- Map object-storage failures to `IOError` so they stay retryable.
- Ideally move the estimator off the reader-open path into the on-demand column API.

### [P2] Removing `get_chunk_size()` breaks existing consumers
The public interface drops `get_chunk_size()` and replaces it with new pure-virtual methods. Milvus still calls the old method (`internal/core/src/segcore/storagev2translator/ManifestGroupTranslator.cpp:133`), so bumping milvus-storage would fail to compile, and every external `ChunkReader` implementation would have to implement the three new methods at once. Please keep the old method as a compatibility alias, and make the new column-level methods non-pure-virtual with a default that returns `NotImplemented` / empty.

### What looks good
- Parquet nested-leaf → top-level aggregation.
- `DistributeMemorySizes` uint128 math + overflow checks.
- Iceberg positional-delete / partial-row-group scaling invariants.
- Full-schema, projection-independent semantics (no single-column normalization issue).

### Net suggestion
Keep this PR's full-schema / Iceberg algorithm, but wrap it in a compatibility-preserving, on-demand, best-effort framework: keep the old `get_chunk_size()` API, make the new column-level methods non-pure with defaults, and never fail `open()` when stats are absent — fall back to a conservative value derived from the row-group size (e.g. `total_byte_size()` split evenly across columns). Happy to help wire the fallback / compat shims.

jiaqizho:
> **Requesting changes.** The core algorithm here is solid — the Parquet nested-leaf → top-level aggregation, the uint128 + overflow-checked `DistributeMemorySizes`, and the Iceberg positional-delete / partial-row-group scaling all look right, and the full-schema, projection-independent result avoids the single-column-projection normalization issue in the alternative approach. My concern is that per-column estimation is wired as a hard prerequisite of reading, which turns a best-effort hint into correctness/availability regressions. Three blocking issues plus one compatibility issue below.
> 
> ### [P1] Vortex: missing column stats fails normal reads
> Vortex metadata load unconditionally requires `UncompressedSizeInBytes`; when stats are disabled, missing, or the file is empty, `open()` returns `NotImplemented`. CI already confirms this:
> 
> ```
> VortexBasicTest.TestEmptyWriteRead/V1  failed
> VortexBasicTest.TestEmptyWriteRead/V2  failed
> NotImplemented: Vortex column memory size statistics are not available
> ```
> 
> Column-level estimation is a best-effort capability and should not affect existing read paths (`get_chunk()`, `take()`, …). Please open the reader normally and only surface `NotImplemented` (or a fallback) when the column-estimation API is actually called.
> 
> ### [P1] Lance: schema evolution makes old fragments unreadable
> The Rust estimator returns results in dataset-schema order and carries `field_id`, but the C++ bridge (`lance_bridge.cpp`) drops `field_id` and keeps only the size, then `lance_table_reader.cpp` matches by vector length against the fragment schema. After a column is added:
> 
> ```
> dataset schema:   [a, b, c]
> old fragment:     [a, b]
> estimator result: [a_size, b_size, 0]
> ```
> 
> `3 != 2` → `Invalid` → the old fragment can't open. Even with an equal count but a changed order, sizes get attributed to the wrong columns. Please keep `field_id` and map by field ID onto the fragment/file schema instead of matching by position/length.
> 
> ### [P1] Lance: transient I/O errors mis-mapped to NotImplemented
> `estimate_fragment_column_memory` reads object-storage footer/page metadata, but the bridge catches every `cxxbridge1::Error` and returns `NotImplemented`. So retryable I/O failures — S3 timeout, network jitter, credential refresh — all become "format not supported", and because the estimator runs inside `open()`, they block normal reads entirely. Suggestions:
> 
> * Return `NotImplemented` only for a confirmed-unsupported legacy format.
> * Map object-storage failures to `IOError` so they stay retryable.
> * Ideally move the estimator off the reader-open path into the on-demand column API.
> 
> ### [P2] Removing `get_chunk_size()` breaks existing consumers
> The public interface drops `get_chunk_size()` and replaces it with new pure-virtual methods. Milvus still calls the old method (`internal/core/src/segcore/storagev2translator/ManifestGroupTranslator.cpp:133`), so bumping milvus-storage would fail to compile, and every external `ChunkReader` implementation would have to implement the three new methods at once. Please keep the old method as a compatibility alias, and make the new column-level methods non-pure-virtual with a default that returns `NotImplemented` / empty.
> 
> ### What looks good
> * Parquet nested-leaf → top-level aggregation.
> * `DistributeMemorySizes` uint128 math + overflow checks.
> * Iceberg positional-delete / partial-row-group scaling invariants.
> * Full-schema, projection-independent semantics (no single-column normalization issue).
> 
> ### Net suggestion
> Keep this PR's full-schema / Iceberg algorithm, but wrap it in a compatibility-preserving, on-demand, best-effort framework: keep the old `get_chunk_size()` API, make the new column-level methods non-pure with defaults, and never fail `open()` when stats are absent — fall back to a conservative value derived from the row-group size (e.g. `total_byte_size()` split evenly across columns). Happy to help wire the fallback / compat shims.

Thanks for the detailed review. I checked the three P1 concerns against the
current Vortex implementation and Lance 7.0.0 behavior.

## 1. Vortex missing column statistics

The CI regression was specific to empty files. An empty file has no footer
statistics entries, although its decoded size is known exactly: every column is
zero bytes. This is now handled before reading the footer statistics by returning
one zero estimate per schema field, so an empty reader opens and reads normally.
`VortexBasicTest.TestEmptyWriteRead` now passes for both V1 and V2.

For non-empty files, disabling writer statistics does not remove
`UncompressedSizeInBytes`. The writer uses `VORTEX_NON_STATS`, which still
contains `Stat::UncompressedSizeInBytes`; it only omits the other optional
statistics. `VortexV2Test.TestV2StatsDisabledUsesPlainRowGroupLayout` already
opens and reads such a file successfully.

I therefore kept the accurate per-column path for non-empty files and used the
exact zero result for empty files. I did not evenly split the row-group size,
because that would manufacture a column distribution that is not present in the
file metadata.

## 2. Lance schema evolution

I checked this against Lance 7.0.0. In this version,
`FileFragment::schema()` returns the current dataset schema, not a physical
per-fragment schema. Consequently, both `GetFragmentSchema()` and
`estimate_fragment_column_memory()` use the same dataset-schema order. A field
that is not physically stored in a fragment receives a zero estimate, but it is
still present in both vectors, so the proposed `3 != 2` mismatch does not occur.

Physical column order is also normalized inside the Rust estimator by Lance
field ID before the final result is emitted in dataset-schema order. The C++
bridge dropping the field ID does not introduce a positional mismatch at that
point.

I added `LanceBasicTest.TestReaderHandlesFragmentMissingNullableDatasetColumn`
to lock this behavior down. It creates a three-column dataset with a fragment
that physically omits the nullable third column and verifies that:

- the reader opens successfully;
- the estimate contains all three dataset columns;
- the missing physical column has a zero estimate; and
- reads materialize the missing column as nulls.

The full `LanceBasicTest.*` suite passes (10/10).

## 3. Lance I/O error status

I agree that estimation failures must be returned as a non-OK result and handled
by the caller. However, I do not think this PR should infer a detailed Arrow
status category from `cxxbridge1::Error`. At the current CXX boundary, the Rust
error is exposed only as an error string, without a stable structured error kind.
Classifying it by message text would be brittle.

The current call sites do not branch on `NotImplemented` versus `IOError`; every
non-OK Arrow status follows the same control flow and is handled by the caller.
Changing only the status category would therefore not change the behavior or
retry policy. The important contract here is whether the estimation operation
succeeds or returns a non-OK status, rather than the precise status subtype.

If callers later need status-code-based retry semantics, I think the robust
solution is to add a structured Rust error enum across the FFI boundary in a
separate change. For this PR, I would prefer to keep the single non-OK mapping
instead of adding heuristic error classification.

cpp/src/format/column_group_reader.cpp
```diff
@@ -211,6 +226,7 @@ arrow::Status ColumnGroupReaderImpl<ReaderT>::open() {
             .row_group_index_in_file = j,
             .global_row_end = rows_in_all_files + rows_in_file,
             .avg_memory_size = row_group_in_file[j].memory_size,
+            .column_memory_sizes = row_group_in_file[j].column_memory_sizes,
```
weiliu1031: [P1] Normalize per-file estimates to a canonical schema order

column_memory_sizes remains in each file's physical schema order, while file_schema_ is initialized only from the first file. The public lookup then resolves a field index against that first schema and applies the same index to every chunk.

Transaction only compares column-name sets, so appending a file with [b, a] after one with [a, b] is accepted. The second file's estimates will then be silently attributed to the wrong fields.

Please remap each file's estimates by field name or stable field ID before storing them, and reject missing or ambiguous fields. A regression test with two files using opposite field orders would cover this case.

cpp/src/format/lance/lance_table_reader.cpp
```diff
@@ -269,11 +308,15 @@ arrow::Status LanceTableReader::open() {
   // Build the read schema for fragment reader:
   // use user-provided schema if available, otherwise project file schema by needed_columns
   ARROW_ASSIGN_OR_RAISE(auto read_schema, build_read_schema(file_schema_, read_schema_, needed_columns_));
-  ARROW_RETURN_NOT_OK(arrow::ExportSchema(*read_schema, &c_arrow_schema));
 
   ARROW_ASSIGN_OR_RAISE(logical_chunk_rows_, api::GetValue<uint64_t>(properties_, PROPERTY_READER_LOGICAL_CHUNK_ROWS));
 
-  auto fragment_memory_size = dataset_->EstimateFragmentMemory(fragment_id_);
+  ARROW_ASSIGN_OR_RAISE(
```
weiliu1031: [P1] Do not make column estimation a prerequisite for opening a Lance reader

This estimation runs before BlockingFragmentReader::Open, so an unsupported estimate prevents all normal reads. Legacy Lance fragments were previously readable with a zero fallback, but they now fail open() with NotImplemented, even though the data reader itself still supports them.

Column estimation is an optional metadata capability and should not determine whether the data can be read. Please allow open() to succeed, preserve an estimate-unavailable state, and return NotImplemented only when the new column-estimation API is called.

cpp/src/format/vortex/vortex_format_reader.cpp
```diff
@@ -459,8 +494,12 @@ arrow::Status VortexFormatReader::open() {
   ARROW_ASSIGN_OR_RAISE(file_schema_, import_vortex_file_schema(*vxfile_));
 
   ARROW_ASSIGN_OR_RAISE(auto row_ranges, get_vortex_splits(*vxfile_));
-  row_group_infos_ = create_row_group_infos(total_mem_usage(), vxfile_->RowCount(),
-                                            recalc_row_ranges(row_ranges, logical_chunk_rows_));
+  ARROW_ASSIGN_OR_RAISE(auto file_column_uncompressed_sizes,
```
weiliu1031: [P1] Missing optional Vortex statistics should not make the file unreadable

UncompressedSizeInBytes is needed for the new estimate, but it is not required to read the data. get_column_uncompressed_sizes() returns NotImplemented when this statistic is absent, and open() now propagates that error unconditionally.

Valid historical and externally generated Vortex files can lack this statistic. Milvus writers before [milvus-storage PR #295](https://github.com/milvus-io/milvus-storage/pull/295) used VortexWriteOptions::default(), whose Vortex 0.52.1 default statistics did not include UncompressedSizeInBytes.

Please decouple estimate availability from reader initialization. Missing statistics should only make the estimate API unavailable, not block get_chunk(), take(), or normal scans. Please also add a fixture that genuinely omits this statistic; the current writer always includes it.

cpp/src/format/bridge/rust/src/lance_bridge.cpp
```diff
@@ -104,6 +104,20 @@ uint64_t BlockingDataset::GetFragmentRowCount(uint64_t fragment_id) const {
...
+arrow::Result<std::vector<uint64_t>> BlockingDataset::EstimateFragmentColumnMemory(uint64_t fragment_id) const {
+  try {
+    auto estimates = ffi::estimate_fragment_column_memory(*impl_, fragment_id);
+    std::vector<uint64_t> memory_sizes;
+    memory_sizes.reserve(estimates.size());
+    for (const auto& estimate : estimates) {
+      memory_sizes.push_back(estimate.memory_size);
+    }
+    return memory_sizes;
+  } catch (const rust::cxxbridge1::Error& e) {
```
weiliu1031, line 116: [P2] Preserve the underlying Rust error category

This catch maps every rust::cxxbridge1::Error to NotImplemented, but the estimator can also fail with an invalid fragment, an internal error, a damaged footer, or an object-store/scheduler I/O error.

Collapsing these cases loses the information needed for retry decisions and diagnostics. For example, a transient S3 failure becomes indistinguishable from a genuinely unsupported legacy format.

Please propagate a structured error category through the FFI and map failures to Invalid, IOError, Internal, or NotImplemented as appropriate. Only a confirmed capability or format limitation should return NotImplemented.

cpp/include/milvus-storage/reader.h
```diff
@@ -88,9 +88,52 @@ class ChunkReader {
...
+   * @brief Returns the estimated decoded-memory size of every chunk.
+   *
+   * The result is indexed by chunk. Estimates are derived from format metadata
+   * and may be zero when a chunk has no live rows or its known decoded-memory
+   * size is zero. Unavailable estimates are returned as an error, typically
+   * arrow::Status::NotImplemented, rather than being represented by zero.
+   *
+   * @return One estimated size in bytes per chunk, or an error status.
+   */
+  [[nodiscard]] virtual arrow::Result<std::vector<uint64_t>> get_chunk_estimated_size() = 0;
```
weiliu1031, line 103: [P1] Preserve the existing public C++ source and ABI contract

Removing get_chunk_size() breaks existing source consumers; [Milvus master still calls this method](https://github.com/milvus-io/milvus/blob/bf674af67b835bd031898fe18d6dfb601dff9a10/internal/core/src/segcore/storagev2translator/ManifestGroupTranslator.cpp#L133).

This also changes the public virtual-function layout. Adding new pure virtual methods before get_chunk_rows() shifts its vtable slot, so a binary compiled against the old header may dispatch to a method with a different signature, resulting in undefined behavior.

Please retain get_chunk_size() as a deprecated compatibility alias, preserve the order of existing virtual methods, and append new methods with default NotImplemented implementations. If an ABI break is intentional, the library needs a corresponding ABI or SONAME version bump and coordinated downstream migration.

cpp/src/format/column_group_reader.cpp
weiliu1031: [P2] Avoid multiplying per-column metadata copies

Adding `column_memory_sizes` turns the existing by-value `RowGroupInfo` flow into O(row_groups × columns) deep copies. On the cached path, `metadata->row_group_infos` is copied into `row_group_in_file`, again into `ColumnGroupReaderImpl::row_group_infos_`, into the format reader, and here into every `ChunkInfo`. `ColumnGroupReaderImpl::row_group_infos_` is only cleared, resized, and assigned; it is never read, so at least one complete copy is dead. For wide schemas or many chunks, these copies can substantially inflate metadata memory.

Please remove the unused `row_group_infos_` copy and share immutable row-group metadata between the cache and readers. Full chunks can reference the canonical per-column vector; only partial row groups need to materialize a scaled vector.

jiaqizho: Won't have this case in storage. In same column group, we can't append  [b, a] after one with [a, b].

cpp/include/milvus-storage/reader.h
jiaqizho: nope, won't direct use the ABI

cpp/src/format/bridge/rust/src/lance_bridge.cpp
jiaqizho: that's fine, The estimated size does not affect normal read/write logic.

cpp/src/format/lance/lance_table_reader.cpp
jiaqizho: ack.

cpp/src/format/vortex/vortex_format_reader.cpp
jiaqizho: ack.

cpp/src/format/column_group_reader.cpp
jiaqizho: I do that part of vector copy is fine. added FIXME.

cpp/src/reader.cpp
```diff
@@ -409,15 +417,17 @@ class PackedRecordBatchReader final : public arrow::RecordBatchReader {
       return true;
     };
 
-    while (!sorted_offsets.empty() && (memory_used_ < memory_usage_limit_ || !all_cgs_have_data())) {
+    bool memory_size_unavailable = false;
```
buqian-zilliz, line 420: **[P1] Keep the unknown-size fallback active while data remains queued**

The unavailable-size fallback lasts for only one `load_internal()` call, while `current_rbs_` survives across calls. If an unknown-size column group leaves a partially consumed batch queued and another group empties first, the refill round resets this flag and may preload a full `memory_usage_limit_` of other chunks on top of the unaccounted batch. Persist unavailable-size state per reader or column group, and clear it only after the corresponding queued data is consumed.

cpp/src/reader.cpp
```diff
@@ -360,12 +360,20 @@ class PackedRecordBatchReader final : public arrow::RecordBatchReader {
     // must have one more chunk to read, should checked in caller
     assert(current_cg_chunk_index < number_of_chunks_per_cg_[column_group_index]);
 
-    ARROW_ASSIGN_OR_RAISE(auto chunk_size, cg_reader->get_chunk_size(current_cg_chunk_index));
+    auto chunk_size_result = cg_reader->get_chunk_estimated_size(current_cg_chunk_index);
+    if (chunk_size_result.ok()) {
+      memory_used_ += chunk_size_result.ValueOrDie();
```
buqian-zilliz, line 365: **[P1] Cap preloading when an estimate is exactly zero**

The new logical-column aggregation can legitimately return zero when a chunk's logical fields are absent or known to have zero decoded size. Adding zero leaves `memory_used_` below the limit, so `load_internal()` can enqueue every remaining zero-estimate chunk in one `get_chunks()` call. RecordBatch objects, queue entries, and other unaccounted allocations still consume memory, allowing long files to bypass `reader.record_batch_max_size`. Apply a conservative per-chunk charge or stop batching after a zero estimate.

cpp/src/format/format_reader.cpp
```diff
@@ -14,12 +14,44 @@
...
+    return arrow::Status::Invalid("Cannot distribute a non-zero memory size with zero column weights");
+  }
+
+  uint64_t allocated = 0;
+  for (size_t i = 0; i + 1 < weights.size(); ++i) {
+    // weights[i] <= total_weight, so the quotient is at most total_size and is safe to cast back to uint64_t.
+    result[i] = static_cast<uint64_t>(static_cast<unsigned __int128>(total_size) * weights[i] / total_weight);
+    allocated += result[i];
+  }
+  result.back() = total_size - allocated;
```
buqian-zilliz, line 51: **[P2] Do not assign rounding bytes to a zero-weight column**

The complete rounding remainder is assigned to the last column even when that column has zero weight. For example, `total_size=1` with weights `[1, 1, 0]` produces `[0, 0, 1]`. Lance uses zero weights for fields absent from a fragment, so an absent trailing field can be reported as non-zero, violating the API contract. Preserve zero-weight entries and distribute remainders only among positive-weight columns.

cpp/src/format/column_group_reader.cpp
```diff
@@ -503,13 +588,64 @@ arrow::Result<std::vector<std::shared_ptr<arrow::RecordBatch>>> ColumnGroupReade
...
+  if (!chunk_info.column_memory_sizes) {
+    return chunk_info.avg_memory_size;
+  }
+
+  // Files in one column group may have evolved physical schemas and can retain
+  // columns removed from the logical group. Sum the logical column estimates
+  // so physical-only columns are excluded while the result remains independent
+  // of the active projection.
+  uint64_t total_size = 0;
+  std::unordered_set<std::string_view> seen_columns;
```
buqian-zilliz, line 610: **[P2] Precompute the logical chunk total**

`PackedRecordBatchReader` calls this method for every chunk considered by the preload heap. Each call allocates a fresh `unordered_set`, walks every logical column, and performs another hash lookup per column, changing the former O(1) estimate lookup into O(chunks × columns) allocations and work on normal scans. Validate duplicate names and compute the logical total once while constructing `ChunkInfo`, then return the cached value here.

cpp/src/reader.cpp
jiaqizho: expected. A zero estimate is a valid result rather than an unavailable estimate. We intentionally do not introduce an arbitrary per-chunk charge, because it would change scheduling semantics without providing a real memory bound.

cpp/src/reader.cpp
jiaqizho: same.

PR #589, opened 2026-07-23
enhance: fix bundled JNI library resolution

congqixia:
The fat JAR only set $ORIGIN on libmilvus-storage and its JNI wrapper. DT_RUNPATH is not inherited by transitive dependencies, so libraries such as Arrow could not find bundled AWS SDK libraries after extraction on a clean host.

Patch every bundled ELF with a RUNPATH relative to the native resource root. Add a dependency check that clears LD_LIBRARY_PATH and LD_PRELOAD before running ldd, and run it in Java CI and the release workflow. Install patchelf in those environments and the Java build image.

### DIFF
--- a/.github/workflows/java-ci.yml
+++ b/.github/workflows/java-ci.yml
@@ -93,7 +93,7 @@ jobs:
             exit 1
           fi
           cp "$JNI_LIB" "$MAIN_LIB" java/native/linux-x86_64/
-          cp -P cpp/build/Release/libs/*.so* java/native/linux-x86_64/ || true
+          cp -a cpp/build/Release/libs/. java/native/linux-x86_64/
           java/patch_native_runpath.sh java/native/linux-x86_64
           java/verify_native_dependencies.sh \
             java/native/linux-x86_64/libmilvus-storage-jni.so
--- a/.github/workflows/java-ci.yml
+++ b/.github/workflows/java-ci.yml
@@ -27,7 +27,7 @@ jobs:
       - name: Install C++ dependencies
         run: |
           sudo apt-get update
-          sudo apt-get install -y cmake libaio-dev
+          sudo apt-get install -y cmake patchelf libaio-dev
           pip install conan==${{ env.CONAN_VERSION }}
 
       - name: Setup JDK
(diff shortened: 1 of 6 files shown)


.github/workflows/java-ci.yml
```diff
@@ -103,6 +83,21 @@ jobs:
...
+        run: |
+          mkdir -p java/native/linux-x86_64
+          JNI_LIB=$(find cpp/build -name "libmilvus-storage-jni.so" -type f | head -1)
+          MAIN_LIB=$(find cpp/build -name "libmilvus-storage.so" -type f | head -1)
+          if [ -z "$JNI_LIB" ] || [ -z "$MAIN_LIB" ]; then
+            echo "ERROR: milvus-storage JNI libraries not found"
+            exit 1
+          fi
+          cp "$JNI_LIB" "$MAIN_LIB" java/native/linux-x86_64/
+          cp -P cpp/build/Release/libs/*.so* java/native/linux-x86_64/ || true
```
jiaqizho: Some `.so` files are missing here: `cpp/build/Release/libs/*.so*`.

```
➜  cpp git:(storage-async-api) ls -d build/Release/libs/*/
build/Release/libs/engines-3/    build/Release/libs/ossl-modules/
```

java/verify_native_dependencies.sh
```diff
@@ -0,0 +1,32 @@
...
+    exit 1
+fi
+
+if ! command -v ldd >/dev/null 2>&1; then
+    echo "ldd is required to verify bundled native dependencies" >&2
+    exit 1
+fi
+
+# Do not allow a developer shell or CI job to hide missing RUNPATH entries.
+ldd_output="$(env -u LD_LIBRARY_PATH -u LD_PRELOAD ldd "$JNI_LIBRARY")"
```
jiaqizho, line 23: `ldd -r`? which also check the `undefined symbol`

congqixia: Thanks for the suggestion. ldd -r would provide stronger validation by also checking relocations and undefined symbols.

I tried it against libmilvus-storage-jni.so, but it currently reports several unresolved Aws::Crt symbols from libarrow.so.1700. The Conan metadata marks aws-crt-cpp as a static dependency, while its generated target does not contribute an actual library to the final ELF dependency closure.

This appears to be a separate, pre-existing Arrow/AWS CRT linkage issue rather than a RUNPATH or packaging issue introduced by this PR. I suggest keeping the regular ldd check in this PR and addressing the AWS CRT linkage in a follow-up.
Once that is fixed, we can safely strengthen this verification to use ldd -r without allowlisting unresolved symbols.

.github/workflows/java-ci.yml
congqixia: addressed

PR #590, opened 2026-07-24
enhance: use adaptive rate-limited retry strategy for S3 client

bigsheeper:
## What & Why

The S3 client's default `ConnectRetryStrategy` used a fixed 200ms retry interval (up to ~30 attempts) and had **no client-side rate limiting**. Under sustained S3 request-rate throttling (`503 SlowDown`) — e.g. a large bulk import writing many small objects to a cold bucket prefix — this hammers the throttled endpoint at a constant rate and amplifies the overload.

This replaces it with a strategy built on the AWS SDK's `AdaptiveRetryStrategy`, which adds:

- **client-side rate limiting** via a send token bucket that auto-enables on throttling responses (CUBIC/AIMD), pacing outgoing requests toward the endpoint's dynamic capacity;
- **jittered exponential backoff** (inherited from the standard strategy), replacing the fixed 200ms interval.

The `IsConnectError` gate is preserved, so both AWS-standard retryable errors and MinIO-specific ones (`SlowDown`, `XMinioServerNotInitialized`) remain retriable.

## Notes

- The rate limiter is dormant under normal load and engages only once throttling is observed, so steady-state behavior is unchanged.
- All S3 requests issued through this client (including `CompleteMultipartUpload`, which goes through the SDK's retry machinery) are paced by the limiter.
- Verified against the aws-sdk-cpp 1.11.692 retry API. Runtime rate-limiting behavior under throttling is to be validated in integration.

### DIFF
--- a/cpp/include/milvus-storage/filesystem/s3/s3_internal.h
+++ b/cpp/include/milvus-storage/filesystem/s3/s3_internal.h
@@ -23,6 +23,7 @@
 #include <utility>
 
 #include <aws/core/Aws.h>
+#include <aws/core/client/AdaptiveRetryStrategy.h>
 #include <aws/core/client/RetryStrategy.h>
 #include <aws/core/http/HttpTypes.h>
 #include <aws/core/http/HttpResponse.h>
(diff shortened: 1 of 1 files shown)


xiaofan-luan:
Thanks for working on this. I suggest treating retry behavior as an object-storage-wide policy rather than an S3-only implementation detail:

1. Please cover all supported object-storage backends, including non-S3-native paths such as Azure Blob/ADLS. The scope should also explicitly state whether credential/STS requests, custom retry strategies, CRT reads, and multipart-upload retry paths are included.
2. Please audit each cloud SDK and align the retry semantics across providers: retryable error classification (429/503, timeout, connection reset, service-not-ready), exponential backoff and jitter, `Retry-After`, maximum attempts and total elapsed-time budget, client-side rate-limiter scope, and idempotency requirements.

“Consistent” should mean a common behavioral contract, not necessarily identical SDK classes or constants. Where a provider SDK cannot provide equivalent adaptive throttling, we should add a provider adapter or a common limiter above the SDK.

Please also add provider-specific fault-injection tests for AWS S3, MinIO, GCS, Aliyun OSS, Tencent COS, Huawei OBS, and Azure to verify the full path from the real SDK error mapping to retry, backoff, and rate limiting.

If this is too large for this PR, please clearly scope this PR as S3-compatible-only and track the Azure/native-SDK and credential-path work as explicit follow-ups.

PR #594, opened 2026-07-27
feat: support Apache Paimon external tables

### LINKED ISSUE #593: [Feature]: Add Apache Paimon table format support
## Motivation

Milvus external collections need to read Apache Paimon tables through Milvus Storage. Safely raw-convertible files can use the existing file readers, while merge-on-read and other split-level semantics require native Paimon `DataSplit` execution.

## Scope

- Add a paimon-rust bridge for table discovery, split planning, and native split reads.
- Add a Paimon format reader with explicit `auto`, `direct-file`, and `data-split` modes:
  - use direct-file readers for safely raw-convertible Parquet and Vortex files, including supported file-level deletion vectors;
  - use atomic native `DataSplit` reads when Paimon-level processing is required, such as merge-on-read, row ranges, partition columns, schema or data evolution, sidecars, or unsupported direct-file formats.
- Keep merge-on-read and deletion-vector handling inside the Paimon format layer.
- Stream native splits in bounded record batches instead of materializing an entire split.
- Forward supported object-storage options through the bridge.
- Persist native split descriptors with explicit codec/version and binding validation.
- Expose Paimon through the C and Python APIs and the `loon` inspection path.
- Add unit and integration coverage for append-only, deletion-vector, merge-on-read, Parquet, Vortex, descriptor validation, and storage options.

## Non-goals

- Paimon table writes.
- ORC direct-file support in the first iteration; `auto` may use the DataSplit path instead.

## Implementation tracking

- [x] Direct-file reads: #602
- [x] Native DataSplit reads and `auto` fallback: #616
- [ ] Delegated cloud credentials and ARN support: follow-up PR

The original combined implementation in #594 was closed in favor of these smaller PRs.

## Cross-repository tracking

Design and external-collection tracking: milvus-io/milvus#45881

yanbinyang:
Closes #593

## What changed

Add Apache Paimon as a read-only external table format using paimon-rust through the existing Rust/CXX bridge.

- Plan Paimon snapshots and persist versioned native `DataSplit` descriptors.
- Route raw-compatible append-only files to the existing direct-file readers.
- Use atomic native `DataSplit` scans when merge-on-read or deletion-vector semantics are required.
- Keep deletion-vector application inside the Paimon reader and stream native splits in bounded record batches.
- Forward supported S3, OSS, Azure, and GCS storage options.
- Expose Paimon through the C and Python APIs and the `loon` inspection path.
- Propagate direct-file physical sizes for downstream planning.

The paimon-rust dependency is pinned to an exact commit from its `release-0.3.0` line, and `Cargo.lock` is committed for reproducible builds.

## Compatibility and scope

- Paimon tables are read-only.
- Parquet and compatible Vortex append-only files can use direct-file reads.
- Merge-on-read, deletion vectors, and other non-raw-compatible splits use the native Paimon reader.
- ORC direct-file reads are outside the initial scope.
- Existing Parquet, Vortex, Lance, and Iceberg behavior is unchanged.

## Test plan

- Rust split codec, planning, deletion-vector, and storage-option tests.
- C++ Paimon format reader and integration tests.
- `loon` manifest/read round-trip tests for append-only, deletion-vector, merge-on-read, Parquet, and Vortex tables.
- Downstream Milvus C++/Go tests and client/v3 Paimon E2E on Linux.


cpp/src/format/bridge/rust/src/paimon_bridgeimpl.rs
```diff
@@ -0,0 +1,1606 @@
...
+        "Paimon deletion vector short read: expected {total_length} bytes, got {}",
+        bytes.len()
+    );
+
+    let declared_length = u32::from_be_bytes(bytes[0..4].try_into()?) as u64;
+    ensure!(
+        declared_length == length,
+        "Paimon deletion vector length mismatch: descriptor {length}, payload {declared_length}"
+    );
+    let magic = u32::from_be_bytes(bytes[4..8].try_into()?);
```
xiaofan-luan: [P2] Detect the real bitmap64 envelope

This reads the magic as big-endian and assumes bitmap64 has the same length contract as bitmap32. Paimon Bitmap64DeletionVector actually uses a big-endian outer data length, a little-endian magic and payload, and DeletionFile.length describes the complete serialized size. A real bitmap64 file therefore will not reach this NotImplemented branch and may fail as a range or invalid-magic error. The current test writes a synthetic big-endian magic and masks the issue. Please parse the real envelope and use a Java-generated fixture. The payload is compatible with roaring::RoaringTreemap, so it can be decoded directly if bitmap64 is in scope; otherwise reject it accurately and do not suggest data-split as a workaround unless the pinned paimon-rust reader supports it.

cpp/src/format/bridge/rust/src/paimon_bridgeimpl.rs
```diff
@@ -0,0 +1,1606 @@
...
+    schema: SchemaRef,
+}
+
+impl Iterator for PaimonStreamReader {
+    type Item = std::result::Result<RecordBatch, ArrowError>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        TOKIO_RT
+            .block_on(self.stream.next())
+            .map(|result| result.map_err(|error| ArrowError::ExternalError(Box::new(error))))
```
xiaofan-luan: [P1] Preserve transient versus permanent stream errors

Wrapping every Paimon stream error as ArrowError::ExternalError makes Arrow C Stream export EINVAL, which C++ imports as Status::Invalid. Transient storage failures such as S3 throttling, timeout, and connection reset therefore become non-retryable. The reverse setup path also defaults every unmarked error to IOError, so permanent descriptor or data failures can become retryable. Please classify errors at the Paimon/OpenDAL source: transient failures should cross as ArrowError::IoError, while corrupt or invalid data and unsupported operations should remain Invalid or NotImplemented. Tests should inject actual stream-time failures instead of validating marker strings only.

cpp/test/format/paimon/paimon_integration_test.cpp
```diff
@@ -0,0 +1,627 @@
...
+  column_group->columns = {"id"};
+  column_group->format = LOON_FORMAT_PAIMON_TABLE;
+  column_group->files = {files.front()};
+  auto schema = arrow::schema({arrow::field("id", arrow::int32())});
+  ASSERT_AND_ASSIGN(auto reader, api::ColumnGroupReader::create(schema, column_group, {"id"}, properties_, nullptr));
+  ASSERT_EQ(reader->total_rows(), static_cast<int64_t>(kRows));
+
+  std::vector<int64_t> chunk_indices(reader->total_number_of_chunks());
+  std::iota(chunk_indices.begin(), chunk_indices.end(), 0);
+  ASSERT_AND_ASSIGN(auto batches, reader->get_chunks(chunk_indices, 2));
```
xiaofan-luan, line 363: [P1] This test does not actually span merge-read batches, and async remains uncovered

The fixture SetUp overrides reader.logical_chunk_rows to 4, so this test creates 4-row logical chunks rather than the 8192-row chunks described in the comment. Since Paimon emits roughly 1024-row batches and 4 divides 1024, no logical chunk crosses a RecordBatch boundary; the test therefore does not exercise the synchronous reassembly added by this PR. Please set logical_chunk_rows to 8192 explicitly in this test, verify the returned chunk sizes and values, and run the same fixture through get_chunks_async. The generic async slicer still requires a whole logical chunk to fit in the current RecordBatch and will otherwise return Invalid slice of record batches.

cpp/src/format/bridge/rust/src/paimon_bridgeimpl.rs
yanbinyang: Fixed in [5dda5bf](https://github.com/milvus-io/milvus-storage/commit/5dda5bfbb97ee89dedba4ee3c03c831015353ad4). Temporary OpenDAL errors are now exported as `ArrowError::IoError` (`EIO`). Permanent storage and invalid-data errors remain non-retryable (`EINVAL`), and unsupported errors map to `NotYetImplemented`.

The test injects errors during `get_next` and checks the Arrow C Stream error codes directly.

cpp/test/format/paimon/paimon_integration_test.cpp
yanbinyang: Fixed in [916a75a](https://github.com/milvus-io/milvus-storage/commit/916a75ab639efe1cfb8f75904401876d4657e6c6). Both synchronous and asynchronous reads now use `ReassembleChunks`, so a logical chunk can span multiple record batches.

The test sets `reader.logical_chunk_rows=8192` with 1024-row Paimon batches and checks row values through both `get_chunks` and `get_chunks_async`.

cpp/src/format/bridge/rust/src/paimon_bridgeimpl.rs
yanbinyang: Fixed in [72d80c0](https://github.com/milvus-io/milvus-storage/commit/72d80c0441fdce425bf996f516685cff653345bd) by correcting bitmap64 detection. The reader now recognizes the Java envelope (big-endian outer length, little-endian bitmap64 magic, and `DeletionFile.length` covering the complete serialized value), validates the referenced range, and returns a non-retryable `NotImplemented` error instead of treating it as bitmap32 or a retryable I/O failure.

The data-split workaround was removed because the pinned paimon-rust reader does not decode bitmap64 either, so changing the scan mode would not make the table readable. The test uses a fixture generated with Paimon Java 1.2.0 and read back with the Java decoder. This keeps the behavior fail-closed and consistent across scan modes.

cpp/test/format/paimon/paimon_integration_test.cpp
yanbinyang: Added explicit per-chunk row-count assertions in [581e3d7](https://github.com/milvus-io/milvus-storage/commit/581e3d7379ae87e0ba64c68803107f70d6e41062). With 1024-row Paimon output batches, an 8192-row logical chunk crosses eight batches.

yanbinyang:
Updated the dependency in [3a5965fa](https://github.com/milvus-io/milvus-storage/commit/3a5965fa7b534b228116b1f309f306e1809c661e) to use the published `paimon = "=0.3.0"` crate instead of the temporary Git revision. The `v0.3.0` tag resolves to the same `7b54d44c4875` revision previously pinned here, so this only changes dependency provenance.

cpp/include/milvus-storage/format/paimon/paimon_format_reader.h
```diff
@@ -0,0 +1,110 @@
...
+#pragma once
+
+#include <memory>
+#include <string>
+#include <vector>
+
+#include "milvus-storage/format/format_reader.h"
+#include "milvus-storage/format/parquet/parquet_format_reader.h"
+#include "milvus-storage/format/vortex/vortex_format_reader.h"
+#include "paimon_bridge.h"
```
jiaqizho, line 22: `bridge.h` should not declare in header. Just move it into cpp

cpp/src/ffi/column_groups_c.cpp
```diff
@@ -15,13 +15,25 @@
...
 
 #include "milvus-storage/ffi_internal/result.h"
 #include "milvus-storage/ffi_internal/bridge.h"
 #include "milvus-storage/manifest.h"
 
 using namespace milvus_storage::api;
 
+namespace {
+
+char* dup_cstr(const char* value) {
```
jiaqizho, line 28: why not use the `strdup`?

cpp/src/format/bridge/rust/testdata/paimon-java-bitmap64.dv
jiaqizho, line 1: What is this file for? notice that: current CI won't run any rust test.

cpp/src/format/bridge/rust/Cargo.toml
```diff
@@ -101,6 +101,20 @@ url = "2"
...
+# Paimon pins this release because 1.13.3 regressed its Linux Vortex tests.
+# Make the shared resolution explicit so the existing Lance/DataFusion edge is
+# updated from the older lockfile entry instead of producing two incompatible
+# solver choices.
+unicode-segmentation = "=1.13.2"
+
+# Apache Paimon reader/planner. Keep the exact release and Cargo.lock together.
+# Paimon uses Arrow 58, which is already present for Lance and crosses the C++
+# boundary only through Arrow's C Data interface.
+paimon = { version = "=0.3.0", default-features = false, features = ["storage-memory", "storage-fs", "storage-s3", "storage-oss", "storage-azdls", "storage-gcs", "vortex"] }
```
jiaqizho, line 117: support `storage-azdls` without `storage-azblob`?

cpp/src/format/paimon/paimon_common.cpp
```diff
@@ -0,0 +1,129 @@
...
+  if (message.find("[paimon:error=invalid]") != std::string_view::npos) {
+    return arrow::Status::Invalid(context, ": ", error.what());
+  }
+  if (message.find("[paimon:error=not-implemented]") != std::string_view::npos) {
+    return arrow::Status::NotImplemented(context, ": ", error.what());
+  }
+  return arrow::Status::IOError(context, ": ", error.what());
+}
+
+std::unordered_map<std::string, std::string> ToStorageOptions(const ArrowFileSystemConfig& config) {
```
jiaqizho, line 35: Need also update the `cpp/test/format/external_table_arn_test.cpp`. Current external table won't direct use ak/sk or IAM access the data.

And current storage support 
- aws assume role.
- aliyun OIDC ARN.
- gcp impersonation.

So just add the the PAIMON-TABLE into the TESTP, also u need verify it.(i can help verify the aliyun case..)

jiaqizho: Please provide the ARN verification results here, including the detailed logs, to ensure the external table functions correctly.

cpp/include/milvus-storage/ffi_c.h
```diff
@@ -320,6 +324,13 @@ FFI_EXPORT LoonFFIResult loon_column_groups_create(const char** columns,
                                                    size_t file_lens,
                                                    LoonColumnGroups** out_column_groups);
 
+/** Add or replace one property on a file owned by LoonColumnGroups. */
+FFI_EXPORT LoonFFIResult loon_column_group_file_set_property(LoonColumnGroups* column_groups,
```
jiaqizho, line 328: This API make me very confuse. 

The column group sturct in public in FFI layer, don't open a function just for put the private metadata.

What situation does the caller need to modify the private metadata of a specific file independently?

cpp/src/ffi/column_groups_c.cpp
```diff
@@ -71,6 +83,57 @@ LoonFFIResult loon_column_groups_create(const char** columns,
   RETURN_UNREACHABLE();
 }
 
+LoonFFIResult loon_column_group_file_set_property(LoonColumnGroups* column_groups,
```
jiaqizho, line 86: don't do that.

jiaqizho (commented):
review isn't finished yet; continue later.

cpp/src/format/paimon/paimon_format_reader.cpp
```diff
@@ -0,0 +1,941 @@
...
+  while (begin < indices.size()) {
+    if (indices[begin] <= previous || indices[begin] < 0 ||
+        static_cast<uint64_t>(indices[begin]) >= metadata_->payload.record_count) {
+      return arrow::Status::Invalid("Paimon take indices must be sorted, unique, and in range");
+    }
+    size_t end = begin + 1;
+    while (end < indices.size() && indices[end] == indices[end - 1] + 1) {
+      ++end;
+    }
+    ARROW_ASSIGN_OR_RAISE(auto range_batches, cursor->ReadRange(static_cast<uint64_t>(indices[begin]),
```
jiaqizho, line 898: Implementing `take` using a streaming reader would be disastrous.

cpp/src/format/paimon/paimon_format_reader.cpp
```diff
@@ -0,0 +1,941 @@
+// Licensed under the Apache License, Version 2.0 (the "License");
```
jiaqizho, line 1: The current PR is simply too large; could you split it into several parts for merging? 
For example, 
- The first PR could implement just the Rust-side interfaces for C++.
- The second could implement the format reader/writer (or `write_test_dataset`)
- The third could implement `explore` and ARN support.

cpp/src/format/bridge/rust/src/paimon_bridge.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+std::vector<PaimonFileInfo> PlanFiles(const std::string& table_location,
+                                      int64_t snapshot_id,
+                                      const std::string& scan_mode,
+                                      const StorageOptions& storage_options) {
+  try {
+    rust::Vec<rust::String> keys;
+    rust::Vec<rust::String> values;
+    ConvertStorageOptions(storage_options, keys, values);
+    auto planned = ffi::paimon_plan_files(rust::Str(table_location), snapshot_id, rust::Str(scan_mode), std::move(keys),
```
jiaqizho, line 46: Why not use the data-split API? That way, there would be no need to do "apply delete" operations on the storage side.

cpp/src/format/bridge/rust/Cargo.toml
yanbinyang: paimon 0.3.0 only provides storage-azdls; it has no storage-azblob feature. This matches the current Iceberg integration, while Lance supports both. Therefore, this PR supports Azure ADLS Gen2 (abfs:// / abfss://), not Azure Blob.

cpp/src/format/bridge/rust/testdata/paimon-java-bitmap64.dv
yanbinyang: This bitmap64 deletion-vector fixture was generated by Paimon Java 1.2.0 to cover the compatibility issue raised in @xiaofan-luan’s comment above. It avoids testing the decoder with a fixture based on the same assumptions. Since Rust tests are not run in CI, I can move it to the C++ testdata directory and exercise it in the C++ integration suite. Do you think we should keep it, or remove it from this PR?

cpp/src/format/bridge/rust/src/paimon_bridge.cpp
yanbinyang: paimon-rust 0.3.0 exposes DataSplit as a sequential RecordBatchStream and does not provide random take(row_indices). Milvus needs FormatReader::take for random row retrieval; implementing it by rescanning the DataSplit stream would have the issue noted in your previous comment. Therefore, raw-convertible DV splits use the direct-file reader with file-local DV mapping, while MOR and other non-raw splits use DataSplit.

yanbinyang:
Thanks for the feedback. Based on the discussion, I’ll close this PR and split the work into smaller, focused PRs. I’ll carry the review feedback into the new submissions.

PR #598, opened 2026-07-27
enhance: add PackedRecordBatchReader::Make factory; fix abort and code-destroying paths in packed

czs007:
## Problem

Three related defects in the packed (V2 API) library-mode error paths — all instances of the "stringify-and-rethrow destroys the classification" / "abort instead of status" classes:

1. **`PackedRecordBatchReader`'s constructor throws.** On a failed open it does `throw std::runtime_error(status.ToString())` — the classification the lower layers carefully attached (`PackedInvalidArgs` / `PackedStorageIO` / `PackedMetadataCorrupted` details, the filesystem's ENOENT) is flattened into a string at the very last step. Direct-link consumers are forced to `try`-wrap construction and can only parse text.
2. **`ColumnGroup::Table()` does the same** (`throw std::runtime_error(result.status().message())` — even drops `ToString()`'s code prefix).
3. **A fifth unguarded `ValueOrDie` abort path** (missed by the #575 sweep, which fixed four): `file_reader.cpp` `FileRecordBatchReader::init`'s `schema==nullptr` branch calls `FieldIDList::Make(schema_).ValueOrDie()`. A parquet file whose stored schema lacks `PARQUET:field_id` metadata **aborts the whole process** — a data-dependent abort. The sibling `schema!=nullptr` branch already used `ARROW_ASSIGN_OR_RAISE`.

## Change

- **New `PackedRecordBatchReader::Make(...) → arrow::Result<std::unique_ptr<...>>`** — same parameters, reports open failures as a status with the classification intact. The throwing constructor is **kept** (documented as deprecated, "do not add new call sites") so direct-link consumers can migrate without a lockstep bump; it can be deleted once they have.
  - milvus side: the only constructor consumer is `segcore/packed_reader_c.cpp` (3 try-wrapped sites). Migrating those to `Make()` hands them a classified status — which is exactly the input the planned "wire the packed C-API through `ToSegcoreError`" follow-up needs (it currently hardcodes FileReadFailed/FileWriteFailed). Tracked for the milvus-side PR; not part of this change.
- **`ColumnGroup::Table()` returns `arrow::Result`**, wrapping the cause without destroying it. `ColumnGroup::Schema()` now reads the schema from the first batch (all batches in a group share one schema) instead of materializing the merged table. In-tree callers updated; there are no milvus-side callers of either.
- **`file_reader.cpp` abort path** → `ARROW_ASSIGN_OR_RAISE`, matching the sibling branch.

## Verification

- Full `make build` clean.
- New tests in `packed_error_status_test.cpp`:
  - `Make()` on a missing file → the wrap preserves the filesystem's not-found detail → `ToSegcoreError` = `ObjectNotExist` (the throwing constructor destroyed exactly this);
  - `Make()` with empty paths → `PackedInvalidArgs`;
  - `Make()` success path reads a batch.
- Packed/column-group suites: 27/29 passed — the single failure is the pre-existing upstream `ColumnGroupTest.MemoryUsageCalculation` (fails 3/3 in isolation on unmodified main; not touched here), one skipped.

Not covered (honest scope): the deprecated throwing constructor still exists (one `throw` site remains by design until milvus migrates); no fault-injection e2e.

Refs: milvus-io/milvus#50903. Follows the classification taxonomy of #574/#575.

### DIFF
--- a/cpp/include/milvus-storage/packed/column_group.h
+++ b/cpp/include/milvus-storage/packed/column_group.h
@@ -54,7 +54,10 @@ class ColumnGroup {
 
   /// Schema of the group's batches: the first batch's schema, or nullptr for
   /// an empty group. Does NOT validate that later batches match -- callers
-  /// needing validation should go through Table().
+  /// needing validation should go through Table(). Migration note: this
+  /// replaces the previous behavior of materializing the merged table (which
+  /// THREW std::runtime_error on an empty group or schema mismatch); check
+  /// for nullptr before dereferencing.
   std::shared_ptr<arrow::Schema> Schema() const;
 
   std::shared_ptr<arrow::RecordBatch> GetRecordBatch(size_t index) const;
--- a/cpp/include/milvus-storage/packed/column_group.h
+++ b/cpp/include/milvus-storage/packed/column_group.h
@@ -45,7 +45,7 @@ class ColumnGroup {
 
   arrow::Status Merge(const ColumnGroup& other);
 
-  std::shared_ptr<arrow::Table> Table() const;
+  arrow::Result<std::shared_ptr<arrow::Table>> Table() const;
 
   std::shared_ptr<arrow::Schema> Schema() const;
 
--- a/cpp/include/milvus-storage/packed/column_group.h
+++ b/cpp/include/milvus-storage/packed/column_group.h
@@ -45,8 +45,16 @@ class ColumnGroup {
 
   arrow::Status Merge(const ColumnGroup& other);
 
+  /// Merge all batches into one table. Source-compat note: this returns
+  /// arrow::Result since the abort/throw cleanup (the previous signature
+  /// threw std::runtime_error on merge failure). Schema consistency across
+  /// batches is validated here (by Table::FromRecordBatches), not in
+  /// AddRecordBatch.
   arrow::Result<std::shared_ptr<arrow::Table>> Table() const;
 
+  /// Schema of the group's batches: the first batch's schema, or nullptr for
+  /// an empty group. Does NOT validate that later batches match -- callers
+  /// needing validation should go through Table().
   std::shared_ptr<arrow::Schema> Schema() const;
 
   std::shared_ptr<arrow::RecordBatch> GetRecordBatch(size_t index) const;
@@ -71,7 +79,7 @@ class ColumnGroup {
   std::vector<std::shared_ptr<arrow::RecordBatch>> batches_;
   size_t memory_usage_;
   std::vector<int> origin_column_indices_;
-  int64_t total_rows_;
+  int64_t total_rows_ = 0;
 };
 
 struct ColumnGroupState {
--- a/cpp/include/milvus-storage/packed/reader.h
+++ b/cpp/include/milvus-storage/packed/reader.h
@@ -47,6 +47,19 @@ class PackedRecordBatchReader : public arrow::RecordBatchReader {
    * @param buffer_size The max buffer size of the packed reader.
    * @param reader_props The reader properties.
    */
+  /// Preferred entry point: reports open failures as a status (with the
+  /// classification the underlying layers attached) instead of throwing.
+  static arrow::Result<std::unique_ptr<PackedRecordBatchReader>> Make(
+      const std::shared_ptr<arrow::fs::FileSystem>& fs,
+      std::vector<std::string>& paths,
+      const std::shared_ptr<arrow::Schema>& schema,
+      int64_t buffer_size = DEFAULT_READ_BUFFER_SIZE,
+      ::parquet::ReaderProperties reader_props = ::parquet::default_reader_properties(),
+      ::parquet::ArrowReaderProperties arrow_reader_props = ::parquet::default_arrow_reader_properties());
+
+  /// Deprecated in favor of Make(): a failed open THROWS std::runtime_error
+  /// with the stringified status, destroying its classification. Kept only
+  /// until direct-link consumers migrate; do not add new call sites.
   PackedRecordBatchReader(
       const std::shared_ptr<arrow::fs::FileSystem>& fs,
       std::vector<std::string>& paths,
@@ -79,6 +92,10 @@ class PackedRecordBatchReader : public arrow::RecordBatchReader {
   arrow::Status Close() override;
 
   private:
+  /// Member-init only; used by Make() so initialization failures can be
+  /// reported as a status instead of a constructor exception.
+  explicit PackedRecordBatchReader(int64_t buffer_size);
+
   arrow::Status init(const std::shared_ptr<arrow::fs::FileSystem>& fs,
                      std::vector<std::string>& paths,
                      const std::shared_ptr<arrow::Schema>& origin_schema,
--- a/cpp/src/common/metadata.cpp
+++ b/cpp/src/common/metadata.cpp
@@ -178,9 +178,7 @@ int64_t RowGroupMetadata::row_offset() const { return row_offset_; }
 
 std::string RowGroupMetadata::ToString() const {
   std::stringstream ss;
-  ss << "memory_size=" << memory_size_ << ","
-     << "row_num=" << row_num_ << ","
-     << "row_offset=" << row_offset_;
+  ss << "memory_size=" << memory_size_ << "," << "row_num=" << row_num_ << "," << "row_offset=" << row_offset_;
   return ss.str();
 }
 
@@ -291,6 +289,12 @@ arrow::Result<std::shared_ptr<PackedFileMetadata>> PackedFileMetadata::Make(
     const std::shared_ptr<parquet::FileMetaData>& metadata) {
   // deserialize row group metadata
   auto key_value_metadata = metadata->key_value_metadata();
+  if (key_value_metadata == nullptr) {
+    // A foreign parquet file without any key-value metadata used to crash
+    // here on the null dereference below.
+    return arrow::Status::Invalid(fmt::format(
+        "Not a packed parquet file: no key-value metadata present. [num_row_groups={}]", metadata->num_row_groups()));
+  }
   auto row_group_meta = key_value_metadata->Get(ROW_GROUP_META_KEY);
   if (!row_group_meta.ok()) {
     return arrow::Status::Invalid(
--- a/cpp/src/format/parquet/file_reader.cpp
+++ b/cpp/src/format/parquet/file_reader.cpp
@@ -99,7 +99,9 @@ arrow::Status FileRowGroupReader::init(std::shared_ptr<arrow::fs::FileSystem> fs
       return status;
     }
     schema_ = file_schema;
-    field_id_list_ = FieldIDList::Make(schema_).ValueOrDie();
+    // FieldIDList::Make fails when a field lacks PARQUET:field_id metadata;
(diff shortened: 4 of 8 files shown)


bigsheeper:
Adversarial review of this PR surfaced 1 medium and 4 low severity issues; the core abort-to-status fix is sound, but a null dereference upstream of it and a few pre-existing gaps in `ColumnGroup` are worth addressing while this code is being touched.

### Medium

- **`cpp/src/common/metadata.cpp:293` — Null dereference in `PackedFileMetadata::Make`.** `metadata->key_value_metadata()` can return a null `shared_ptr` for parquet files without key-value metadata, but the result is immediately dereferenced via `->Get(ROW_GROUP_META_KEY)` with no null check. Reading a foreign parquet file will crash here before ever reaching the fix at `file_reader.cpp:104` — this is also the root cause making that fix's regression test hard to write for bare foreign files. Suggestion: add a null guard and return a proper error status; handle it together with the `file_reader.cpp:104` fix. *(raised by bigsheeper)*

### Low

- **`cpp/src/format/parquet/file_reader.cpp:104` — Missing regression test for the `ValueOrDie` → `ARROW_ASSIGN_OR_RAISE` fix.** The fix itself is effective (`FieldIDList::Make` has a `Contains` guard and returns `Status::Invalid` instead of aborting, with unit coverage in `metadata_test.cpp:122-135`), but no test exercises the integration-level error path through `FileRowGroupReader::init` to line 104 — the three new `Make` tests in `packed_error_status_test.cpp` don't reach it. Note the path is testable today: a parquet file carrying the three required KV keys but whose fields lack `PARQUET:field_id` passes `PackedFileMetadata::Make` at `file_reader.cpp:92` and reaches line 104. Suggestion: add that regression test now; once `metadata.cpp:293` is fixed, the bare foreign-file scenario can be covered as well. *(raised by xaxys, bigsheeper)*

- **`cpp/include/milvus-storage/packed/column_group.h:74` — `ColumnGroup::total_rows_` is never initialized.** It has no default member initializer and neither constructor's init list (`column_group.cpp:23-32`) includes it, so `AddRecordBatch`'s `total_rows_ += batch->num_rows()` operates on an indeterminate value (UB) and `GetTotalRows()` returns garbage. Additionally, the batch-taking constructor (used by both splitters) skips updating `total_rows_` and `batch_memory_usage_`, leaving `size()==1` with `GetRecordMemoryUsages().size()==0` — an out-of-bounds risk for any caller pairing them by index. Current impact is zero (both getters have no callers repo-wide), but the fix is trivial: initialize `total_rows_ = 0` and complete the constructors while this PR is already modifying the class. *(raised by bigsheeper)*

- **`cpp/include/milvus-storage/packed/column_group.h:48` — `ColumnGroup::Table()` return-type change is a source-level breaking change.** Changing the return type to `arrow::Result<std::shared_ptr<arrow::Table>>` breaks source compatibility for downstream consumers. The impact is limited: the library sets no soname/SOVERSION (no ABI promise; consumers must rebuild headers and library together), rebuilt consumers get a loud compile error rather than silent misbehavior, and the sole call site (`column_group_test.cpp:86`) is already updated in this PR. Suggestion: flag the change in release/PR notes as a source-compatibility note. *(raised by xaxys)*

- **`cpp/include/milvus-storage/packed/column_group.h:50` — `Schema()`/`Table()` no longer validate multi-batch schema consistency, and the new contract is undocumented.** The new `Schema()` simply returns the first batch's schema (`column_group.cpp:70-74`), and `AddRecordBatch` only checks for a null pointer, whereas the old implementation would surface schema mismatches via `Table::FromRecordBatches`. The failure scenario is unreachable inside the library today (both methods have zero callers in `cpp/src`, and both splitters construct single-batch groups), so this is a documentation gap rather than a defect: add contract comments at `column_group.h:48-50` noting the empty→nullptr behavior and that consistency is no longer validated. *(raised by xaxys, bigsheeper)*

czs007:
All five findings addressed in 8add0c2:

- **Medium (null deref in `PackedFileMetadata::Make`)**: guarded with a proper `Invalid` status. Worth noting: this proved **reachable, not theoretical** — the arrow `FileWriter` used in the new integration test produces a KV-less parquet file by default, and the first draft of that test hit the guard (pre-fix it would have crashed the process). Nice catch on the root cause blocking the regression test, too.
- **Low (integration test for the `file_reader.cpp` fix)**: added — a parquet file carrying a genuine packed file's key-value metadata (`FileWriter::AddKeyValueMetadata`) but a schema without `PARQUET:field_id`, driven through `FileRowGroupReader::Make`'s `schema==nullptr` path; asserts `Invalid`, not abort.
- **Low (`total_rows_` uninitialized / batch ctor incomplete)**: default-initialized; the batch-taking constructor now maintains `total_rows_` and `batch_memory_usage_` like `AddRecordBatch`.
- **Low (contract docs)**: `Table()`'s Result-return and `Schema()`'s first-batch/no-validation contract documented at the declarations.
- **Low (source compatibility)**: noted here per the suggestion — **`ColumnGroup::Table()` changes its return type to `arrow::Result<std::shared_ptr<arrow::Table>>`**, a source-level breaking change; there is no ABI promise (no soname), the only in-tree caller is updated in this PR, and there are no milvus-side callers.

Packed suites 28/29 (the one failure is the pre-existing upstream `ColumnGroupTest.MemoryUsageCalculation`).

tinswzy:
## Adversarial Review Summary

Verification confirmed 5 issues (1 high, 3 medium, 1 low); the status-propagation fixes themselves are sound, but the public-header signature change on `ColumnGroup::Table()` needs an explicit migration decision before merge.

### High

- **`cpp/include/milvus-storage/packed/column_group.h:53` — Silent ABI break from `Table()` return-type change.** The return type changed from `std::shared_ptr<arrow::Table>` to `arrow::Result<std::shared_ptr<arrow::Table>>`. Under the Itanium C++ ABI, return types don't participate in name mangling, so old and new symbols are identical: a stale caller passes sret storage sized for a 16-byte `shared_ptr` while the new implementation writes a larger `arrow::Result` — memory corruption rather than a link error. The blast radius is limited: in-repo, `.Table()` is only called from tests, and the FFI shared library exports only C symbols via `ffi_exports.map`, so the risk is confined to external consumers that direct-link the C++ API without rebuilding (e.g. milvus segcore). The real inconsistency is that this same PR keeps a deprecated throwing constructor on `PackedRecordBatchReader` as a migration shim (`reader.h:60-62`) while changing `Table()` in place. Suggestion: state in the PR description whether any cross-binary caller of `Table()` exists on the milvus side, and unify the migration strategy — either add a `TryTable()` and keep the old symbol, or explicitly declare this a breaking change requiring a lockstep rebuild. *(raised by tinswzy, chyezh)*

### Medium

- **`cpp/src/packed/column_group.cpp:31` — Batch constructor dereferences a null batch before its null guard.** `GetRecordBatchMemorySize(batch)` (line 31) unconditionally dereferences `batch`, and line 30 stores it, both before the `total_rows_ += batch ? batch->num_rows() : 0` guard at line 36 — making that guard dead code, since the process has already crashed at line 31. Unlike `AddRecordBatch`, which returns `PackedInvalidArgs` on null, the constructor has no status path at all. Suggestion: reject null at entry (mirroring `AddRecordBatch`, ideally via a `Result`-returning factory), or remove the misleading half-guard. *(raised by tinswzy, chyezh)*

- **`cpp/src/packed/column_group.cpp:69` — `WrapExtendError` drops the underlying status code, degrading error classification to `StorageError`.** Contrary to the "propagate it unchanged" comment, `arrow::Table::FromRecordBatches` returns an `Invalid` status with no detail for empty groups or mismatched schemas; `WrapExtendError` then falls through to `MakeExtendError`, which maps `PackedUnexpected` to `IOError`. Callers lose `IsInvalid()`, and `ToSegcoreError` classifies malformed data as `StorageError` (2044) instead of `DataFormatBroken` (2024) — recreating the classification loss this PR aims to eliminate. Suggestion: preserve the original code and detail (e.g. `result.status().WithMessage(...)`) and add a regression assertion for the schema-mismatch error code. *(raised by chyezh)*

- **`cpp/src/packed/column_group.cpp:78` — `Schema()` no longer detects cross-batch schema inconsistency (documented contract; limited residual risk).** Neither `AddRecordBatch` nor `Merge` validates schema equality, and `Schema()` returns `batches_.front()->schema()`, so it can report a schema that `Table()` would reject; pre-PR, `Schema()` surfaced the mismatch by materializing the table. Verification notes this is a deliberate, documented division of responsibility (`column_group.h:55-57` explicitly says validation happens in `Table()`), and no non-test in-repo caller of `Schema()` exists. The residual risk is that external callers relying on `Schema()` for pre-validation now discover mismatches late in the write path. Suggestion (optional): add a schema `Equals` check in `AddRecordBatch`/`Merge` to fail early, and add a regression test asserting `Table()`'s error code on mismatch. *(raised by tinswzy, chyezh)*

### Low

- **`cpp/src/packed/column_group.cpp:78` — Empty-group `Schema()` changed from throwing to returning `nullptr` (documented, but a silent behavior change for existing callers).** Previously `Schema()` materialized the table and threw `std::runtime_error` on an empty group; now it returns `nullptr`, so an unaudited caller doing `group.Schema()->num_fields()` goes from a catchable exception to a null-pointer dereference. The `nullptr` contract itself is documented in the header (`column_group.h:55-56`); what's missing is a note that it replaces prior throwing behavior. (A related concern — `batches_.front()` being null — was found unreachable during verification: the constructor crashes on a null batch first, and `AddRecordBatch` rejects null.) Suggestion: add a migration note to the header comment or changelog, or return `arrow::Result` for symmetry with `Table()`. *(raised by tinswzy)*

czs007:
Round-2 findings addressed in b723d21:

- **High (Table() ABI hazard) — explicit decision, as requested**: declared a **source-level breaking change requiring a lockstep rebuild**, now stated in the PR description with the rationale: the library ships no soname/ABI promise; a repo-wide + milvus-side audit found **zero cross-binary callers of `Table()`/`Schema()`** (in-repo callers are tests only; milvus consumes packed via `packed_reader_c.cpp`, which does not touch ColumnGroup); and the conan consumption path always rebuilds against the new headers, so the silent-sret-corruption window requires a consumer that hand-links a stale binary — none exists. The migration-strategy asymmetry you flagged is intentional and now documented: `PackedRecordBatchReader` keeps a deprecated throwing-constructor shim because it has **3 live milvus call sites** needing a migration window; `Table()` has none, so no `TryTable()` dual API (long-term surface cost for a zero-caller scenario).
- **Medium (Table() wrapped away the Invalid)**: confirmed — `WrapExtendError`→`PackedUnexpected` rewrote `Invalid` into `IOError`, degrading 2024→2044, contradicting the comment. `Table()` now preserves the original StatusCode/detail via `WithMessage`; regression test pins schema-mismatch → `Invalid` → `DataFormatBroken`.
- **Medium (null deref before the guard)**: confirmed — the round-1 bookkeeping fix left the half-guard dead. Null is now rejected at constructor entry (empty group; `AddRecordBatch` remains the status-reporting path); test pins no-crash + empty-group.
- **Medium (early schema validation, optional)**: took the test half (the mismatch→2024 regression above anchors `Table()`'s contract); left `AddRecordBatch` behavior unchanged to avoid a write-path behavior change beyond review scope — noted as possible follow-up.
- **Low (throw→nullptr migration note)**: added to the header contract comment.

Packed error suites 13/13.

PR #599, opened 2026-07-28
enhance: support Azure ARN with broker-issued SAS tokens

jiaqizho:
Azure cross-tenant access now obtains short-lived SAS tokens from the cloud credential broker instead of relying on a long-lived account key or a one-off static SAS value. The new configuration carries the broker endpoint, client and tenant IDs, Azure account name, region, container, requested lifetime, and request timeout through the existing filesystem properties. It also rejects incomplete or invalid broker configurations early and includes the broker identity in filesystem cache keys so credentials cannot be reused across different tenants or accounts.

On the C++ filesystem path, add an Azure SDK HTTP policy that calls the broker, parses and validates the returned account name, SAS token, and expiration time, and attaches the SAS parameters to each Blob or Data Lake request through Azure’s URL API. The policy caches tokens, refreshes them ahead of expiration, reports detailed broker and refresh failures, falls back to the last cached token when one exists, and stops the Azure request pipeline when no usable token can be obtained.

Carry the same broker configuration into the Rust bridge so Lance can refresh SAS values through its storage options accessor for both reads and writes. Iceberg fetches a short-lived SAS token before planning files and passes it to OpenDAL as adls.sas-token. Cloud-specific credential setup is now selected through a bridge-private cloud_provider value in one place, keeping the existing AWS AssumeRole, GCP impersonation, and Aliyun role flows alongside the new Azure path without leaking internal routing options into upstream libraries.

Keep the existing Azure authentication modes unchanged, including account keys, workload identity, anonymous access, and caller-supplied SAS tokens. Broker authentication is only selected when the Azure broker properties are present, so current deployments continue using their existing credential path.


jiaqizho:
without iceberg tested.

cpp/src/filesystem/azure/azure_sas_token_policy.cpp
```diff
@@ -0,0 +1,255 @@
...
+      return arrow::Status::IOError("Azure SAS credential broker response body read failed: endpoint=",
+                                    state_->config.endpoint, ", error=", error.what());
+    }
+    return arrow::Status::IOError("Azure SAS credential broker returned an invalid expiration time: expiredAt=",
+                                  broker_response.token.expired_at, ", error=", error.what());
+  }
+}
+
+arrow::Result<std::string> AzureSasTokenPolicy::GetSasToken() const {
+  std::lock_guard<std::mutex> lock(state_->mutex);
```
liliu-z, line 202: GetSasToken() holds state_->mutex across the broker HTTP fetch (lock at line 194, FetchSasToken at 201), so a slow or unreachable broker serializes every Azure request — even ones that only need to read a still-valid cached token — behind one in-flight attempt, each blocking up to request_timeout_ms. On failure it returns the cached SAS unconditionally (lines 212-213), re-applying an already-expired token that Azure 403s instead of failing closed. The original 'permanent 403 / no self-recovery' framing overstates it: the code re-attempts the broker on every non-fresh request and recovers once it returns (see test RefreshFailureReturnsOldTokenAndRetriesEveryRequest), so the real cost is refresh-window serialization plus a wasted round-trip on expired tokens. Fetch outside the lock (single-flight) and fail closed once the SAS is genuinely expired.

cpp/src/format/bridge/rust/src/azure_sas_provider.rs
```diff
@@ -0,0 +1,529 @@
...
+        {
+            let cached = self.cached.read().await;
+            if let Some(credential) = cached.as_ref()
+                && Self::is_fresh(credential, now)
+            {
+                return Ok(Some(Self::to_options(credential)));
+            }
+        }
+
+        let mut cached = self.cached.write().await;
```
liliu-z, line 332: fetch_storage_options() holds the cached.write() guard across fetcher.fetch(now).await (write lock at 327, fetch at 335), so when a refresh is needed a slow or unreachable broker serializes all callers behind the single in-flight attempt. On failure it returns the previously cached credential unconditionally (353-354) without invalidating its past-expiry timestamp, handing object_store an expired SAS that 403s rather than failing closed. The 'never self-heals' claim is inaccurate — the next non-fresh call re-fetches and succeeds once the broker returns (test caches_and_retries_failed_refresh_with_old_token). Release the write lock before the fetch and fail closed once the credential is genuinely past expiry.

cpp/src/filesystem/azure/azure_sas_token_policy.cpp
```diff
@@ -0,0 +1,255 @@
...
+      {"durationSeconds", state_->config.duration_seconds},
+      {"azureClientId", state_->config.client_id},
+      {"azureTenantId", state_->config.tenant_id},
+      {"azureAccountName", state_->config.account_name},
+  };
+
+  const auto request_body = request_json.dump();
+  const std::vector<uint8_t> request_bytes(request_body.begin(), request_body.end());
+  Azure::Core::IO::MemoryBodyStream body_stream(request_bytes);
+  Azure::Core::Http::Request request(Azure::Core::Http::HttpMethod::Post, Azure::Core::Url(state_->config.endpoint),
```
liliu-z: Azure::Core::Url(state_->config.endpoint) is constructed at line 108, outside the try that begins at line 117, and neither GetSasToken() nor Send() wraps the call — so any endpoint that passes the arrow::util::Uri validation in fs.cpp but the pinned Azure Url parser rejects throws an uncaught exception that escapes the policy's error handling instead of surfacing as the intended 401. The two parsers need not agree; an IPv6 literal such as http://[::1]:80 is a plausible divergence. Move the Url construction inside the try block.

cpp/src/filesystem/azure/azure_sas_token_policy.cpp
```diff
@@ -0,0 +1,255 @@
...
+        "Azure SAS token unavailable: " + sas_token.status().ToString());
+    response->SetHeader("x-ms-error-code", "AuthenticationFailed");
+    response->SetHeader("Content-Length", "0");
+    return response;
+  }
+
+  // SAS authentication is encoded as URL query parameters. Parse the token
+  // with Azure::Core::Url and append it without replacing operation parameters
+  // already added by the Storage SDK, such as comp or restype.
+  const auto sas_query_parameters = Azure::Core::Url("?" + sas_token.ValueOrDie()).GetQueryParameters();
```
liliu-z, line 261: Same throwing-Url-ctor pattern as the endpoint case, this time on the hot path: Send() parses the broker-supplied SAS with Azure::Core::Url("?" + sas_token.ValueOrDie()) at line 248 inside a Send() body (228-253) that has no try/catch, so a SAS containing invalid percent-encoding would throw an uncaught exception into the Azure pipeline rather than producing a clean auth failure. Reachability is low because the token comes from a trusted broker, but wrapping the parse keeps the policy's failure handling uniform.

cpp/src/filesystem/azure/azure_sas_token_policy.cpp
jiaqizho: This behavior is intentional and matches the existing C++ credential-provider semantics. AWS SDK C++ STSAssumeRoleCredentialsProvider holds m_credsMutex across AssumeRole(), and our Aliyun, Tencent, and Huawei providers likewise upgrade to the writer lock before Reload(), which performs the network request. This provides single-flight refresh; simply fetching outside the lock would allow concurrent refreshes unless additional in-flight state were introduced. Retaining the cached SAS, including after expiry, is also intentional: the unchanged expiry causes every subsequent lookup to retry the broker, while Azure rejects the expired token until refresh recovers. We will keep these documented AWS-compatible semantics in this PR.

cpp/src/format/bridge/rust/src/azure_sas_provider.rs
jiaqizho: This is intentional for the same reason. The Rust Aliyun and GCP providers keep the write guard across the async refresh, and Lance StorageOptionsAccessor itself holds its write lock while awaiting StorageOptionsProvider::fetch_storage_options(). The old expiry is intentionally preserved so the next lookup retries the broker. Changing only this provider would not remove the actual serialization and would diverge from the native and existing provider behavior, so we are keeping the current semantics.

cpp/src/filesystem/azure/azure_sas_token_policy.cpp
jiaqizho: ACK

cpp/src/filesystem/azure/azure_sas_token_policy.cpp
jiaqizho: In the pinned azure-core 1.11.3 implementation, the Url constructor does not percent-decode query components. It calls AppendQueryParameters(), which stores the encoded key and value strings verbatim, and GetQueryParameters() returns that encoded map. Invalid percent sequences are rejected only by an explicit Url::Decode() call, which this path never performs. Therefore this constructor does not throw for malformed percent encoding, so no change is needed here.

cpp/src/filesystem/fs.cpp
```diff
@@ -149,6 +237,30 @@ arrow::Status ArrowFileSystemConfig::create_file_system_config(const milvus_stor
...
+      return arrow::Status::Invalid(
+          "Azure credential broker mode requires fs.azure_client_id, fs.azure_tenant_id, "
+          "fs.azure_credential_endpoint, fs.access_key_id, fs.bucket_name, fs.region, and a positive "
+          "fs.request_timeout_ms");
+    }
+
+    arrow::util::Uri endpoint;
+    auto endpoint_status = endpoint.Parse(result.azure_credential_endpoint);
+    if (!endpoint_status.ok() || endpoint.host().empty() ||
+        (endpoint.scheme() != "http" && endpoint.scheme() != "https")) {
```
liliu-z, line 260: The broker-endpoint validation accepts plaintext `http://` (it only rejects schemes that are neither `http` nor `https`). The policy then POSTs to that endpoint and reads back the SAS `sessionToken` — a bearer credential granting broad container access — so over an `http` endpoint the token is transmitted in cleartext and an on-path observer can capture and replay it against Azure over HTTPS until it expires. Require the broker endpoint scheme to be `https`, with at most an explicit loopback exception for local testing.

cpp/src/format/bridge/rust/src/azure_sas_provider.rs
```diff
@@ -0,0 +1,529 @@
...
+            || config.bucket.is_empty()
+            || config.duration_seconds == 0
+            || config.request_timeout_ms == 0
+        {
+            bail!("incomplete Azure credential broker configuration");
+        }
+
+        let endpoint = url::Url::parse(&config.endpoint)
+            .map_err(|_| anyhow!("Azure credential broker endpoint is not a valid URL"))?;
+        if (endpoint.scheme() != "http" && endpoint.scheme() != "https") || !endpoint.has_host() {
```
liliu-z, line 106: The Rust bridge's broker-config validation accepts plaintext `http://` exactly like the C++ native path, and `AzureBrokerClient::fetch` then POSTs to that endpoint and parses the SAS `session_token` from the response. So Lance reads/writes and Iceberg file planning will transmit the issued SAS bearer token in cleartext over an `http` broker endpoint. This is the same cleartext-credential exposure flagged on the C++ path and should be fixed together: require `https`, with at most a loopback exception for tests.

cpp/src/filesystem/fs.cpp
jiaqizho: HTTP is optional by user. Should not be disabled.

cpp/src/format/bridge/rust/src/azure_sas_provider.rs
jiaqizho: same reason

jiaqizho:
> without iceberg tested.

iceberg verified.

cpp/src/filesystem/azure/azure_sas_token_policy.cpp
```diff
@@ -0,0 +1,261 @@
...
+    if (broker_response.token.account_name != state_->config.account_name) {
+      return arrow::Status::IOError("Azure SAS credential broker returned a token for the wrong account: expected=",
+                                    state_->config.account_name, ", actual=", broker_response.token.account_name);
+    }
+
+    auto sas_token = std::move(broker_response.token.sas_token);
+    if (!sas_token.empty() && sas_token.front() == '?') {
+      sas_token.erase(0, 1);
+    }
+    if (sas_token.empty()) {
```
liliu-z, line 159: The broker token is validated only for non-emptiness (line 159) and a future expiredAt (line 166), never for containing a `sig` parameter. A non-empty but signature-less token carrying a far-future expiredAt passes, is cached, and — because IsFresh keys off expires_at rather than token validity — is treated as fresh for the whole TTL, so every request 403s with no re-fetch until expiry. The Rust fetch() (`if token.is_empty()`) has the identical gap. Require a non-empty `sig` parameter before accepting and caching the token.

jiaqizho: ack

PR #600, opened 2026-07-29
enhance: make parquet open_async non-blocking asynchronous

jiaqizho:
Previously, open_async only moved the synchronous Parquet open path onto a Folly executor. Opening a remote file could still block a worker while fetching the file size and footer, so the async API did not provide a fully non-blocking open flow.

The new path uses ParquetFileReader::OpenAsync and bridges the Arrow futures back to the Folly result. When a usable footer-size hint is available, it reads that suffix directly, parses the trailer and metadata, and fetches only the missing metadata range when the hint is too small. Encrypted files and unusable hints fallback to Parquet’s native asynchronous footer handling.

The non-blocking random-access interface now also supports asynchronous size lookup, backed by S3 CRT HeadObjectAsync with cached size reuse and existing error mapping. Open-time continuations are transferred onto the executor supplied through via(), while normal data reads return to their original I/O path after the reader has finished opening.

### DIFF
--- a/cpp/src/filesystem/s3/s3_filesystem.cpp
+++ b/cpp/src/filesystem/s3/s3_filesystem.cpp
@@ -723,7 +723,7 @@ class ObjectInputFile final : public arrow::io::RandomAccessFile {
 };
 
 #ifdef WITH_CRT
-class ObjectCrtInputFile final : public arrow::io::RandomAccessFile, public NonBlockingReadAtFile {
+class ObjectCrtInputFile final : public arrow::io::RandomAccessFile, public NonBlockingRandomAccessFile {
   protected:
   struct ReadState {
     explicit ReadState(int64_t size) : content_length(size) {}
@@ -796,6 +796,59 @@ class ObjectCrtInputFile final : public arrow::io::RandomAccessFile, public NonB
     ARROW_RETURN_NOT_OK(CheckClosed());
     ARROW_RETURN_NOT_OK(EnsureHeadObject(/*need_metadata=*/false));
     return GetCachedContentLength();
+  }
+
+  Future<int64_t> GetSizeAsync() override {
+    auto status = CheckClosed();
+    if (!status.ok()) {
+      return FailedReadFuture(status);
+    }
+
+    const auto content_length = GetCachedContentLength();
+    if (content_length != kNoSize) {
+      return Future<int64_t>::MakeFinished(content_length);
+    }
+
+    auto maybe_client_lease = holder_->Acquire();
+    if (!maybe_client_lease.ok()) {
+      return FailedReadFuture(maybe_client_lease.status());
+    }
+
+    auto ctx = std::make_shared<AsyncHeadContext>();
+    ctx->future = Future<int64_t>::Make();
+    ctx->client_lease = std::move(maybe_client_lease).ValueOrDie();
+    ctx->read_state = read_state_;
+    ctx->path = path_;
+    ctx->request.SetBucket(ToAwsString(path_.bucket));
+    ctx->request.SetKey(ToAwsString(path_.key));
+
+    ctx->client_lease->HeadObjectAsync(
+        ctx->request, [ctx](const Aws::S3Crt::S3CrtClient*, const S3CrtModel::HeadObjectRequest&,
+                            const S3CrtModel::HeadObjectOutcome& outcome,
+                            const std::shared_ptr<const Aws::Client::AsyncCallerContext>&) mutable {
+          if (!outcome.IsSuccess()) {
+            if (outcome.GetError().GetResponseCode() == Aws::Http::HttpResponseCode::NOT_FOUND) {
+              ctx->future.MarkFinished(arrow::Result<int64_t>(PathNotFound(ctx->path)));
+              return;
+            }
+            ctx->future.MarkFinished(arrow::Result<int64_t>(
+                ErrorToStatus(std::forward_as_tuple("When reading information for key '", ctx->path.key,
+                                                    "' in bucket '", ctx->path.bucket, "': "),
+                              "HeadObject", outcome.GetError())));
+            return;
+          }
+
+          const auto content_length = outcome.GetResult().GetContentLength();
+          if (content_length < 0) {
+            ctx->future.MarkFinished(
+                arrow::Result<int64_t>(arrow::Status::IOError("HeadObject returned a negative Content-Length")));
+            return;
+          }
+
+          ctx->read_state->content_length.store(content_length, std::memory_order_release);
+          ctx->future.MarkFinished(content_length);
+        });
+    return ctx->future;
   }
 
   arrow::Status Seek(int64_t position) override {
@@ -1048,6 +1101,16 @@ class ObjectCrtInputFile final : public arrow::io::RandomAccessFile, public NonB
     std::shared_ptr<FilesystemMetrics> metrics;
     int64_t position = 0;
     int64_t nbytes = 0;
+  };
+
+  struct AsyncHeadContext {
+    // Keep the same non-owning CRT client lifetime model as AsyncReadContext.
+    // The path and read state remain valid without owning the file or holder.
+    Future<int64_t> future;
+    S3CrtClientLease client_lease;
+    S3CrtModel::HeadObjectRequest request;
+    std::shared_ptr<ReadState> read_state;
+    S3Path path;
   };
 
   std::shared_ptr<S3CrtClientHolder> holder_;
--- a/cpp/src/format/parquet/parquet_format_reader.cpp
+++ b/cpp/src/format/parquet/parquet_format_reader.cpp
@@ -15,8 +15,10 @@
 #include "milvus-storage/format/parquet/parquet_format_reader.h"
 
 #include <cstring>
+#include <exception>
 #include <limits>
 #include <memory>
+#include <new>
 #include <numeric>
 #include <optional>
 #include <unordered_map>
@@ -35,7 +37,9 @@
 #include <arrow/type.h>
 #include <arrow/util/async_generator.h>
 #include <arrow/util/key_value_metadata.h>
+#include <arrow/util/ubsan.h>
 #include <folly/futures/Promise.h>
+#include <parquet/file_reader.h>
 #include <parquet/arrow/schema.h>
 #include <parquet/metadata.h>
 #include <parquet/type_fwd.h>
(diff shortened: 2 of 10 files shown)


cpp/src/format/parquet/parquet_format_reader.cpp
```diff
@@ -495,10 +558,332 @@ arrow::Status ParquetFormatReader::open() {
...
+            auto parquet_reader = open_future.MoveResult().MoveValueUnsafe();
+            std::unique_ptr<::parquet::arrow::FileReader> arrow_reader;
+            auto status = ::parquet::arrow::FileReader::Make(arrow::default_memory_pool(), std::move(parquet_reader),
+                                                             state->arrow_reader_properties, &arrow_reader);
+            if (!status.ok()) {
+              complete(std::move(status));
+              return;
+            }
+
+            status = state->reader->finish_open(std::shared_ptr<::parquet::arrow::FileReader>(std::move(arrow_reader)));
```
liliu-z: finish_open() is called inside the OpenAsync completion callback here with no surrounding try/catch, but it can throw: finish_open -> create_row_group_infos -> try_build_row_group_infos -> RowGroupMetadata::Deserialize throws std::runtime_error / std::stoull on malformed row-group metadata (metadata.cpp:203,206). A thrown exception escapes the arrow Future callback instead of being converted to arrow::Status, so state->promise is never set and the caller observes a folly BrokenPromise (or, when driven inline, the raw C++ exception) rather than a cleanly failed future. Wrap the finalization body in try/catch and convert thrown exceptions to arrow::Status before calling complete().

cpp/src/format/parquet/parquet_format_reader.cpp
```diff
@@ -495,10 +558,332 @@ arrow::Status ParquetFormatReader::open() {
...
+          });
+    };
+
+    auto prepare_metadata = [state](int64_t file_size) -> arrow::Future<std::shared_ptr<::parquet::FileMetaData>> {
+      if (file_size < 0) {
+        return arrow::Future<std::shared_ptr<::parquet::FileMetaData>>::MakeFinished(arrow::Status::Invalid(
+            fmt::format("Parquet file size is negative. [path={}, file_size={}]", state->reader->path_, file_size)));
+      }
+      state->file_size = file_size;
+      state->reader->file_size_ = static_cast<uint64_t>(file_size);
```
liliu-z: The file size resolved by the async HEAD is written only to the transient reader's file_size_ here; MetaTrait::Payload has no size field and create_metadata_from_reader never stores it, so the resolved size never reaches the cached Metadata. When the caller-supplied kPropertyFileSize is 0 (the same case that required the async HEAD), cache-hit reconstruction via create_from_metadata opens with file_size=0, and Parquet's SerializedFile constructor calls source_->GetSize() unconditionally before the metadata check (parquet/file_reader.cc:304), forcing ObjectCrtInputFile::GetSize -> EnsureHeadObject to issue a blocking synchronous S3 HEAD even though the footer metadata is cached. Persist the resolved size (e.g. add it to Payload and prefer it in create_from_metadata) so reconstruction reuses it.

cpp/src/format/parquet/parquet_format_reader.cpp
```diff
@@ -297,21 +352,45 @@ static arrow::Result<std::shared_ptr<::parquet::arrow::FileReader>> create_parqu
...
+  } else {
+    ARROW_ASSIGN_OR_RAISE(parquet_file, fs->OpenInputFile(file_path));
+  }
+
+  if (!key_retriever && footer_size > 0 && !metadata && file_size > 0 && footer_size <= file_size) {
+    auto footer_buffer = try_read_footer_buffer(parquet_file, file_size, footer_size);
+    metadata = try_parse_footer_metadata(footer_buffer, reader_props);
+  }
+
+  ARROW_RETURN_NOT_OK(builder.Open(std::move(parquet_file), reader_props, metadata));
```
liliu-z, line 370: The cache-backed reconstruction path falls back to a synchronous `builder.Open()` that performs blocking footer I/O, and because the guard at line 363 is `if (key_retriever)` (retriever-existence, not footer encryption) any retriever-configured reader — plaintext included — takes this path, not just encrypted footers. This blocks the async executor thread on a network footer read, defeating the async-open contract, and is the limitation flagged by the existing FIXME. Reconstruct and open the footer asynchronously so retriever-configured readers don't stall the caller.

cpp/src/filesystem/s3/s3_filesystem.cpp
```diff
@@ -788,6 +788,58 @@ class ObjectCrtInputFile final : public arrow::io::RandomAccessFile, public NonB
...
+
+          const auto content_length = outcome.GetResult().GetContentLength();
+          if (content_length < 0) {
+            ctx->future.MarkFinished(
+                arrow::Result<int64_t>(arrow::Status::IOError("HeadObject returned a negative Content-Length")));
+            return;
+          }
+
+          ctx->self->SetCachedContentLength(content_length);
+          ctx->future.MarkFinished(content_length);
```
liliu-z, line 849: When open_async's future is consumed with an inline executor (e.g. `.via(&folly::InlineExecutor::instance())`), FollyArrowExecutor forwards each TransferAlways continuation straight into `executor_->add`, so the next step runs inline inside the AWS callback that publishes the previous result. GetSizeAsync's HEAD callback calls `ctx->future.MarkFinished` (s3_filesystem.cpp:838) while `AsyncHeadContext` still holds the finalizer `std::shared_lock`, and the chained footer read then calls `holder_->Lock()` again (s3_filesystem.cpp:870), recursively acquiring a second shared_lock on the same finalizer mutex on one thread — undefined behavior, and a deadlock against a concurrent `FinalizeS3()` writer on a writer-preferring `shared_mutex`. The new footer-to-missing-metadata read pair self-chains the same way (both reads go through ReadAtAsyncInto, which also holds its lock across `MarkFinished`), so the result must be published after the client/finalizer guard is released in both GetSizeAsync and the read path, not only for the HEAD-to-GET hop.

cpp/src/format/parquet/parquet_format_reader.cpp
jiaqizho: ack

cpp/src/format/parquet/parquet_format_reader.cpp
jiaqizho: ack

cpp/src/format/parquet/parquet_format_reader.cpp
```diff
@@ -248,21 +303,45 @@ static arrow::Result<std::shared_ptr<::parquet::arrow::FileReader>> create_parqu
...
+    arrow::fs::FileInfo file_info(file_path, arrow::fs::FileType::File);
+    file_info.set_size(static_cast<int64_t>(file_size));
+    ARROW_ASSIGN_OR_RAISE(parquet_file, fs->OpenInputFile(file_info));
+  } else {
+    ARROW_ASSIGN_OR_RAISE(parquet_file, fs->OpenInputFile(file_path));
+  }
+
+  if (!key_retriever && footer_size > 0 && !metadata && file_size > 0 && footer_size <= file_size) {
+    auto footer_buffer = try_read_footer_buffer(parquet_file, file_size, footer_size);
+    metadata = try_parse_footer_metadata(footer_buffer, reader_props);
```
liliu-z, line 367: This call site feeds the footer suffix into `try_parse_footer_metadata`, whose length check `footer_length + PARQUET_FOOTER_TRAILER_SIZE > static_cast<uint64_t>(suffix->size())` (line 286) evaluates in 32-bit unsigned arithmetic: a footer whose length field is near `0xFFFFFFFF` wraps to a small value, passes the guard, and then line 291 computes `data + suffix->size() - 8 - footer_length`, a pointer roughly 4 GiB before the buffer that Thrift decoding then dereferences (the `catch (...)` at line 294 cannot recover from the resulting fault). The identical wrap exists in `try_read_footer_buffer` at line 262, so it does not filter the value first. The arithmetic itself is pre-existing, but this PR relocates the call and adds an overflow-safe twin (`try_parse_footer_trailer`, line 640, which widens to `uint64_t` before comparing) — the sync path should use the same widening or reuse that helper.

cpp/src/format/parquet/parquet_format_reader.cpp
```diff
@@ -459,10 +525,351 @@ arrow::Status ParquetFormatReader::open() {
...
+
+      // footer_size is an optimization hint, not part of Parquet correctness.
+      // Encrypted files and unusable hints are delegated to Parquet's native
+      // open path by returning null metadata.
+      const auto footer_size = static_cast<int64_t>(state->reader->footer_size_);
+      if (state->reader->key_retriever_ || footer_size < kParquetFooterTrailerSize || footer_size > file_size) {
+        return arrow::Future<std::shared_ptr<::parquet::FileMetaData>>::MakeFinished(nullptr);
+      }
+
+      return read_file_async_and_transfer(state->file, file_size - footer_size, footer_size, state->executor)
```
liliu-z: A failure of the hinted footer read (and of the follow-up metadata read at line 845) propagates all the way out and fails `open_async()`, because the error flows through `.Then` into `metadata_future.AddCallback`, which calls `complete(metadata_result.status())`. The synchronous `open()` path treats the same failed read as merely an unusable optimization — `try_read_footer_buffer` returns `nullptr`, `try_parse_footer_metadata` returns `nullptr`, and `builder.Open` re-reads the footer through Parquet. So a transient S3 error during the hint read now aborts an async open that the sync path recovers from. Map read failures on this optimization path to null metadata so `ParquetFileReader::OpenAsync` retries natively.

cpp/src/format/parquet/parquet_format_reader.cpp
jiaqizho: ACK

cpp/src/format/parquet/parquet_format_reader.cpp
jiaqizho: won't be, the size will be cached in s3filesystem. And GetSizeAsync() will be call before the  GetSize().

cpp/src/filesystem/s3/s3_filesystem.cpp
jiaqizho: checkout https://github.com/milvus-io/milvus-storage/pull/612

cpp/src/format/parquet/parquet_format_reader.cpp
```diff
@@ -459,10 +528,351 @@ arrow::Status ParquetFormatReader::open() {
...
+      return self->fs_->OpenInputFile(self->path_);
+    }();
+    if (!maybe_file.ok()) {
+      return folly::makeSemiFuture(maybe_file.status());
+    }
+    auto file = std::move(maybe_file).ValueUnsafe();
+
+    // Bridge the Arrow open chain back to the Folly return type without
+    // blocking either the CRT callback thread or a Folly worker.
+    auto promise = std::make_shared<folly::Promise<arrow::Status>>();
```
somfornot: As a general guideline, the same piece of state should not be captured by multiple lambdas at the same time. Doing so makes state transitions very easy to lose track of and makes the code much harder to maintain. At any given point, a piece of state should be owned by only one callback. If the state needs to flow from one callback to another, it should be moved through the function chain rather than shared as it is today.

For example, `promise` is only used by the `complete` lambda, so it only needs to be move-captured by that lambda. There is no need to put it into `state`.

Similarly, `state` itself should only be owned by one callback at a time. Right now it is held by both `start_parquet_open` and `metadata_future.AddCallback`, which is also not ideal. If `metadata_future.AddCallback` can pass the state to `start_parquet_open` as an argument, then `start_parquet_open` can take ownership of it and eventually call `complete`.

A basic principle for this kind of code that bridges multiple asynchronous/state-management mechanisms is: avoid shared ownership of state. Prefer explicit ownership transfer through moves.

PR #601, opened 2026-07-29
enhance: add async reader reconstruction from metadata

jiaqizho:
Async column group reads can load cached metadata through a future, but they still rebuild the format reader synchronously after that metadata becomes available. Reader reconstruction may reopen the underlying file and perform blocking remote I/O, so the public async API can still block the thread that starts the operation. Add `create_from_metadata_async` to the async metadata contract and use it in both eager and lazy column group readers. This keeps metadata loading and reader reconstruction in the same future chain, while formats without a complete async metadata path continue using the existing synchronous fallback.

Parquet and Vortex now provide deferred adapters around their existing metadata-based constructors. Reconstruction starts only when the SemiFuture is driven, allowing the work to follow the caller-provided executor and leaving room for each format to adopt a native asynchronous implementation later.

### DIFF
--- a/cpp/src/format/parquet/parquet_format_reader.cpp
+++ b/cpp/src/format/parquet/parquet_format_reader.cpp
@@ -475,6 +475,19 @@ arrow::Result<std::shared_ptr<ParquetFormatReader>> ParquetFormatReader::MetaTra
   return reader;
 }
 
+folly::SemiFuture<arrow::Result<std::shared_ptr<ParquetFormatReader>>>
+ParquetFormatReader::MetaTrait::create_from_metadata_async(MetadataPtr metadata,
+                                                           const api::ColumnGroupFile& file,
+                                                           const std::shared_ptr<arrow::Schema>& read_schema,
+                                                           const std::vector<std::string>& needed_columns,
+                                                           const std::string& predicate) {
+  return folly::makeSemiFuture().deferValue(
+      [metadata = std::move(metadata), file, read_schema, needed_columns,
+       predicate](folly::Unit) -> arrow::Result<std::shared_ptr<ParquetFormatReader>> {
+        return create_from_metadata(std::move(metadata), file, read_schema, needed_columns, predicate);
+      });
+}
+
 arrow::Status ParquetFormatReader::open() {
   assert(file_reader_ == nullptr);
 
--- a/cpp/include/milvus-storage/format/format_reader.h
+++ b/cpp/include/milvus-storage/format/format_reader.h
@@ -214,15 +214,23 @@ concept FormatReaderWithMetadata =
       { metadata->file_schema } -> std::same_as<const std::shared_ptr<arrow::Schema>&>;
     };
 
-// Identifies formats whose immutable metadata loader itself returns a future;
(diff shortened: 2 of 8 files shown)


cpp/src/format/parquet/parquet_format_reader.cpp
```diff
@@ -475,6 +475,19 @@ arrow::Result<std::shared_ptr<ParquetFormatReader>> ParquetFormatReader::MetaTra
...
+folly::SemiFuture<arrow::Result<std::shared_ptr<ParquetFormatReader>>>
+ParquetFormatReader::MetaTrait::create_from_metadata_async(MetadataPtr metadata,
+                                                           const api::ColumnGroupFile& file,
+                                                           const std::shared_ptr<arrow::Schema>& read_schema,
+                                                           const std::vector<std::string>& needed_columns,
+                                                           const std::string& predicate) {
+  return folly::makeSemiFuture().deferValue(
+      [metadata = std::move(metadata), file, read_schema, needed_columns,
+       predicate](folly::Unit) -> arrow::Result<std::shared_ptr<ParquetFormatReader>> {
+        return create_from_metadata(std::move(metadata), file, read_schema, needed_columns, predicate);
```
buqian-zilliz, line 487: **[P2] Async Parquet reconstruction still blocks the consumer executor**

`create_from_metadata_async()` only queues the existing synchronous constructor onto the consumer's current executor. The caller was already in a deferred continuation, and `create_from_metadata()` immediately reaches `OpenInputFile()`/`builder.Open()`; encrypted readers cannot reuse cached footer metadata, so `.get()` and single-thread event-loop consumers still block on file/footer work, while a thread-pool consumer merely loses one worker. Use a native asynchronous reconstruction path or explicitly offload this blocking work to a designated blocking executor.

cpp/src/format/parquet/parquet_format_reader.cpp
```diff
@@ -475,6 +475,19 @@ arrow::Result<std::shared_ptr<ParquetFormatReader>> ParquetFormatReader::MetaTra
...
 }
 
+folly::SemiFuture<arrow::Result<std::shared_ptr<ParquetFormatReader>>>
+ParquetFormatReader::MetaTrait::create_from_metadata_async(MetadataPtr metadata,
+                                                           const api::ColumnGroupFile& file,
+                                                           const std::shared_ptr<arrow::Schema>& read_schema,
+                                                           const std::vector<std::string>& needed_columns,
+                                                           const std::string& predicate) {
+  return folly::makeSemiFuture().deferValue(
+      [metadata = std::move(metadata), file, read_schema, needed_columns,
```
buqian-zilliz, line 485: **[P3] Do not deep-copy reconstruction state twice**

The common continuations already own `file`, `read_schema`, and `needed_columns` (and the eager path also owns `predicate`), but this adapter takes borrowed parameters and captures them again. That deep-copies `ColumnGroupFile::properties`, the projection strings, and the predicate once per file or split task, while the non-mutable closure also turns `std::move(metadata)` into a `shared_ptr` copy. Accept owned arguments and move them from a mutable outer continuation, or retain only the scalar fields needed for reconstruction.

PR #602, opened 2026-07-29
feat: support Apache Paimon direct-file reads

yanbinyang:
Part of #593

## What changed

Add Apache Paimon as a read-only external table format using the released paimon-rust 0.3.0 crate through the existing Rust/CXX bridge.

- Plan pinned Paimon snapshots and route raw-convertible Parquet and Vortex files through the existing direct-file readers.
- Apply file-local deletion vectors inside the Paimon reader so chunk, range, and take operations use logical row positions.
- Persist versioned direct-file metadata in column group file properties.
- Forward supported S3, OSS, Azure, and GCS storage options.
- Expose Paimon scan and snapshot properties through the C and Python APIs and the `loon` inspection path.

The paimon-rust dependency uses the exact published `=0.3.0` release, and `Cargo.lock` is committed for reproducible builds.

## Compatibility and scope

- Paimon tables are read-only.
- `auto` selects direct-file only when the split is safely raw-convertible; `direct-file` requires that path explicitly.
- Splits that require native `DataSplit` execution fail with `NotImplemented`; native DataSplit reads will be added separately.
- Delegated role ARN credentials are outside this PR and will be added separately.
- Parquet and compatible Vortex files are supported. ORC direct-file reads are outside the initial scope.
- Existing Parquet, Vortex, Lance, and Iceberg behavior is unchanged.

## Test plan

- C++ Paimon format reader and integration tests for append-only tables, deletion vectors, snapshot selection, Parquet, Vortex, malformed metadata, and unsupported DataSplit descriptors.
- `loon` manifest and read tests for the direct-file path.
- Rust planning, deletion-vector, and storage-option unit coverage.
- Downstream Milvus C++/Go tests and client/v3 Paimon E2E on Linux.


yanbinyang:
Updated with two follow-up fixes:

- `b143ef28` rejects Vortex predicate pushdown when a deletion vector is present, avoiding incorrect physical-row deletion mapping.
- `85b9bb9` classifies deterministic Paimon configuration and deletion-vector validation failures as `Invalid` while keeping storage and network failures retryable.

cpp/include/milvus-storage/format/paimon/paimon_format_reader.h
```diff
@@ -0,0 +1,87 @@
...
+
+namespace milvus_storage::paimon {
+
+class PaimonFormatReader final : public FormatReader {
+  public:
+  struct MetaTrait {
+    struct Payload {
+      std::string data_format;
+      parquet::ParquetFormatReader::MetaTrait::MetadataPtr parquet_metadata;
+      vortex::VortexFormatReader::MetaTrait::MetadataPtr vortex_metadata;
```
jiaqizho:
```
private: 
friend class PaimonFormatReader;
union {
     parquet::ParquetFormatReader::MetaTrait::MetadataPtr parquet_metadata;
      vortex::VortexFormatReader::MetaTrait::MetadataPtr vortex_metadata;
} direct_file_cache;
```
packed the direct file cache into union? Also declare it as a friend to prevent circular dependencies.

jiaqizho: Also, please add a comment explaining why it is necessary to reference the metacache of another format here.

cpp/src/format/bridge/rust/Cargo.toml
```diff
@@ -101,6 +101,12 @@ url = "2"
 hmac = "0.12"
 sha1 = "0.10"
 base64 = "0.22"
+crc32fast = "1"
+roaring = "0.11"
+
+# Apache Paimon planner and test writer. Keep the exact release and Cargo.lock
+# together. Paimon uses the Arrow 58 already present for Lance.
+paimon = { version = "=0.3.0", default-features = false, features = ["storage-fs", "storage-s3", "storage-oss", "storage-azdls", "storage-gcs", "vortex"] }
```
jiaqizho, line 109: paimon support `storage-azblob` or not？

cpp/src/format/paimon/paimon_format_reader.cpp
```diff
@@ -0,0 +1,619 @@
...
+    value = folly::parseJson(json);
+  } catch (const std::exception& error) {
+    return arrow::Status::Invalid("Cannot parse Paimon metadata: ", error.what());
+  }
+  if (!value.isObject()) {
+    return arrow::Status::Invalid("Paimon metadata must be a JSON object");
+  }
+  try {
+    auto version = value.getDefault("version", 0).asInt();
+    if (version != 1) {
```
jiaqizho: if (version != 1) will return fail, but the defualt value is 0.

cpp/src/format/paimon/paimon_format_reader.cpp
```diff
@@ -0,0 +1,619 @@
...
+constexpr uint64_t kVariableWidthColumnBytes = 32;
+constexpr uint64_t kUnknownColumnBytes = 16;
+
+struct ParsedMetadata {
+  std::string data_format;
+  uint64_t record_count = 0;
+  folly::dynamic deletion_file;
+};
+
+arrow::Result<ParsedMetadata> ParseMetadata(const std::string& json) {
```
jiaqizho, line 43: there are two try...catch in the same function. just combine it into single try...catch.

cpp/src/format/paimon/paimon_format_reader.cpp
```diff
@@ -0,0 +1,619 @@
...
+  folly::dynamic value;
+  try {
+    value = folly::parseJson(json);
```
jiaqizho: use the `nlohmann_json`, do not use the folly json.

cpp/src/format/paimon/paimon_format_reader.cpp
```diff
@@ -0,0 +1,619 @@
...
+  } catch (const std::exception& error) {
+    return arrow::Status::Invalid("Cannot parse Paimon metadata: ", error.what());
+  }
+  if (!value.isObject()) {
+    return arrow::Status::Invalid("Paimon metadata must be a JSON object");
+  }
+  try {
+    auto version = value.getDefault("version", 0).asInt();
```
jiaqizho: just 
```
auto version = value.getDefault("version", 0).asInt();
auto read_path = value.getDefault("read_path", "").asString();
auto record_count = value.getDefault("record_count", 0).asInt();

// then check the json result.
```

cpp/src/format/paimon/paimon_format_reader.cpp
```diff
@@ -0,0 +1,619 @@
...
+    if (version != 1) {
+      return arrow::Status::Invalid("Unsupported Paimon metadata version: ", version);
+    }
+    auto read_path = value.getDefault("read_path", "").asString();
+    if (read_path == "data-split") {
+      return arrow::Status::NotImplemented("Paimon data-split reads are not supported");
+    }
+    if (read_path != "direct-file") {
+      return arrow::Status::Invalid("Paimon metadata read_path must be direct-file");
+    }
```
jiaqizho: `if (read_path == "data-split")`  and `if (read_path != "direct-file")` is dup. combine into the same `if`.

cpp/src/format/paimon/paimon_format_reader.cpp
```diff
@@ -0,0 +1,619 @@
...
+      return kVariableWidthColumnBytes;
+    default: {
+      auto width = type->byte_width();
+      return width > 0 ? static_cast<uint64_t>(width) : kUnknownColumnBytes;
+    }
+  }
+}
+
+// Estimated decoded bytes per row from the file schema.
+uint64_t EstimateRowByteWidth(const std::shared_ptr<arrow::Schema>& schema) {
```
jiaqizho: what is this? Just making random guesses?

Is there no better solution here? This will lead to OOM or extremely poor performance during the milvus `LOAD`.

cpp/src/format/paimon/paimon_format_reader.cpp
```diff
@@ -0,0 +1,619 @@
...
+  } catch (const std::exception& error) {
+    return arrow::Status::Invalid("Paimon metadata has invalid field types: ", error.what());
+  }
+}
+
+uint64_t SaturatingMultiply(uint64_t left, uint64_t right) {
+  if (left == 0 || right == 0) {
+    return 0;
+  }
+  if (left > std::numeric_limits<uint64_t>::max() / right) {
```
jiaqizho:
```
if (left > (std::numeric_limits<uint64_t>::max() / right))
```

Although they are equivalent, the latter is more readable.

cpp/src/format/paimon/paimon_format_reader.cpp
```diff
@@ -0,0 +1,619 @@
...
+  }
+  for (size_t i = 0; i < batches.size(); ++i) {
+    const auto& physical = metadata_->payload.direct_physical_row_groups[indices[i]];
+    ARROW_ASSIGN_OR_RAISE(auto filtered, filter_direct_batch(batches[i], physical.start_offset));
+    output.push_back(std::move(filtered));
+  }
+  return output;
+}
+
+int64_t PaimonFormatReader::logical_to_physical(int64_t logical_offset) const {
```
jiaqizho: uint64_t ? then u don't need do the static_cast. and check others pls .

cpp/src/format/paimon/paimon_format.cpp
```diff
@@ -0,0 +1,93 @@
...
+}
+
+arrow::Result<std::shared_ptr<FormatReader>> PaimonFormat::create_reader(
+    const std::shared_ptr<arrow::Schema>& read_schema,
+    const api::ColumnGroupFile& file,
+    const api::Properties& properties,
+    const std::vector<std::string>& needed_columns,
+    const std::function<std::string(const std::string&)>& key_retriever) {
+  ARROW_ASSIGN_OR_RAISE(auto metadata,
+                        paimon::PaimonFormatReader::MetaTrait::load_metadata(file, properties, key_retriever));
```
jiaqizho, line 73: The memcache can be closed. 

Do u have test without the cache?

cpp/include/milvus-storage/format/paimon/paimon_format_reader.h
yanbinyang: Only one underlying metadata cache is active. I’ll use std::variant instead of a raw union and add a comment. The cache avoids reopening and reparsing the Parquet/Vortex file.

cpp/src/format/bridge/rust/Cargo.toml
yanbinyang: Paimon 0.3.0 has no storage-azblob feature.

cpp/src/format/paimon/paimon_format_reader.cpp
yanbinyang: This is a workaround for Vortex files without uncompressed-size statistics. We can get the exact compressed file size, but it may underestimate LOAD memory. Should we use that as the fallback, or report the estimate as unavailable?

cpp/src/format/paimon/paimon_format_reader.cpp
yanbinyang: Thanks for pointing this out. This was missed when I trimmed down the previous PR. I’ll fix it.

cpp/src/format/paimon/paimon_format_reader.cpp
yanbinyang: Fine. I’ll use nlohmann_json here and simplify the parsing and validation.

cpp/src/format/paimon/paimon_format_reader.cpp
yanbinyang: OK. I’ll add the parentheses for clarity.

cpp/src/format/bridge/rust/Cargo.toml
jiaqizho: Then azure only support dfs?

yanbinyang: Yes, only covers ADLS Gen2.

cpp/include/milvus-storage/format/paimon/paimon_format_reader.h
yanbinyang: Updated in [4ff630b6](https://github.com/milvus-io/milvus-storage/commit/4ff630b678f5a33b6ed7fce949067e833d51852a).

cpp/src/format/paimon/paimon_format_reader.cpp
yanbinyang: Updated in [9232e140](https://github.com/milvus-io/milvus-storage/commit/9232e1400b8bcce343acc3aba30f2f4766592972). version is kept as a reserved field with a default of 0, but is no longer rejected before version-specific behavior is introduced.

cpp/src/format/paimon/paimon_format_reader.cpp
yanbinyang: Updated in [9232e140](https://github.com/milvus-io/milvus-storage/commit/9232e1400b8bcce343acc3aba30f2f4766592972). Parsing and typed field extraction are now handled by a single try/catch.

cpp/src/format/paimon/paimon_format_reader.cpp
yanbinyang: Updated in [9232e140](https://github.com/milvus-io/milvus-storage/commit/9232e1400b8bcce343acc3aba30f2f4766592972). Paimon reader and planning metadata now use nlohmann::json.

cpp/src/format/paimon/paimon_format_reader.cpp
yanbinyang: Updated in [9232e140](https://github.com/milvus-io/milvus-storage/commit/9232e1400b8bcce343acc3aba30f2f4766592972). Fields are read with typed defaults and validated after parsing.

cpp/src/format/paimon/paimon_format_reader.cpp
yanbinyang: Updated in [9232e140](https://github.com/milvus-io/milvus-storage/commit/9232e1400b8bcce343acc3aba30f2f4766592972). read_path is now checked once: data-split returns NotImplemented, while other non-direct values return Invalid.

cpp/src/format/paimon/paimon_format_reader.cpp
yanbinyang: Updated in [9232e140](https://github.com/milvus-io/milvus-storage/commit/9232e1400b8bcce343acc3aba30f2f4766592972).

cpp/src/format/paimon/paimon_format.cpp
yanbinyang: Added in [66ab091a](https://github.com/milvus-io/milvus-storage/commit/66ab091a76e55ac8f55a20a6687c8872c654cc5e). The tests cover successful append and deletion-vector reads with metadata caching disabled. The DV case also verifies that repeated reader creation does not populate the caller cache and that the cache is populated again after re-enabling it.

cpp/src/format/paimon/paimon_format_reader.cpp
yanbinyang: Updated in [3176e7c8](https://github.com/milvus-io/milvus-storage/commit/3176e7c852eec4e6877d6e5e53c667a0df93e848).

cpp/src/format/paimon/paimon_format_reader.cpp
jiaqizho: no need to do that. if external have no enable the `Stat::UncompressedSizeInBytes`, just return the `arrow::status` no ok. The milvus have fallback logical which will sampling some data to do the estimation.

yanbinyang: Updated in [2b884a23](https://github.com/milvus-io/milvus-storage/commit/2b884a231fbdaa5526146c384c279593b785845d), following the latest FormatReader::get_rg_column_memsz() design from #605. I removed the schema-width guess and delegated per-column estimates to the underlying direct-file reader. If Vortex does not provide UncompressedSizeInBytes, the reader returns NotImplemented, allowing Milvus to use its existing sampling fallback.

cpp/src/format/bridge/rust/src/paimon_bridgeimpl.rs
```diff
@@ -0,0 +1,1200 @@
...
+        if file.file_name.to_ascii_lowercase().contains(".vector.") {
+            return Some((
+                RouteReason::DedicatedVectorFile,
+                format!(
+                    "data file '{}' is a dedicated vector-store file that requires Paimon column merge",
+                    file.file_name
+                ),
+            ));
+        }
+        if file.delete_row_count.is_some_and(|count| count > 0) {
```
xiaofan-luan, line 402: [P1] Please reject legacy primary-key files whose `delete_row_count` is unknown here. For a non-DV deduplicate table, Paimon 0.3's own `pk_split_needs_merge` requires every file to have a known delete count: scan planning may still mark a legacy `None` value raw-convertible for compatibility, but the read side deliberately falls back to the merge reader because that file may hide retract rows. Accepting `None` here can raw-read stale/deleted physical rows. Please mirror that guard and add a legacy-file coverage case.

cpp/src/format/paimon/paimon_format_reader.cpp
```diff
@@ -0,0 +1,533 @@
...
+    if (deleted > physical_rows) {
+      return arrow::Status::Invalid("Paimon deletion count exceeds physical row group size");
+    }
+    auto logical_rows = physical_rows - deleted;
+    uint64_t logical_memory_size = 0;
+    if (group.memory_size_available && physical_rows != 0) {
+      logical_memory_size =
+          static_cast<uint64_t>(static_cast<unsigned __int128>(group.memory_size) * logical_rows / physical_rows);
+    }
+    result.push_back(RowGroupInfo{.start_offset = static_cast<size_t>(logical_start),
```
xiaofan-luan, line 102: [P1] Please avoid exposing a zero-length logical row group when the deletion vector removes every row from one physical group. `ColumnGroupReader` turns this `start_offset == end_offset` entry into a zero-row chunk; in its batched `get_chunks` path, a preceding non-empty batch can advance `rbs_idx` and the following zero-row chunk then dereferences past the end of `rbs_in_file`. Please either skip zero-row chunks while retaining a physical-to-logical row-group mapping, or make the shared chunk path handle them safely. A multi-row-group test with the middle or trailing group fully deleted would cover this.

cpp/src/format/bridge/rust/src/paimon_bridge.cpp
```diff
@@ -0,0 +1,78 @@
...
+    auto planned = ffi::paimon_plan_files(rust::Str(table_location), snapshot_id, rust::Str(scan_mode), std::move(keys),
+                                          std::move(values));
+    std::vector<PaimonFileInfo> result;
+    result.reserve(planned.size());
+    for (const auto& info : planned) {
+      result.push_back(PaimonFileInfo{std::string(info.path), info.file_size, std::string(info.metadata_json)});
+    }
+    return result;
+  } catch (const rust::cxxbridge1::Error& error) {
+    throw PaimonException(error.what());
```
xiaofan-luan: This exception-based bridge currently blocks the PR: `error-handling-ratchet` reports six new `throw` sites (`48 -> 54`). Please return `arrow::Result` / `arrow::Status` from the bridge and carry terminal-vs-retryable classification with status detail instead of throwing and parsing marker strings at callers.

xiaofan-luan (commented):
I found two correctness issues in the direct-file path, plus one error-handling CI blocker. Details are inline.

cpp/src/format/bridge/rust/src/paimon_bridge.cpp
yanbinyang: Fixed in [1c84079](https://github.com/milvus-io/milvus-storage/commit/1c84079ed43a684033d6ad799119d39087aa1813). The Paimon bridge APIs now return arrow::Result / arrow::Status, with bridge errors mapped to terminal, not-found, or retryable transient statuses as appropriate. Unclassified errors remain plain IOError.

cpp/src/format/paimon/paimon_format_reader.cpp
yanbinyang: Fixed in [4a6e130](https://github.com/milvus-io/milvus-storage/commit/4a6e130bf497d5bf4e881556c555026923bfc18a). The reader retains the physical row-group metadata and mapping, but no longer exposes zero-length logical groups as chunks. The test covers a multi-row-group file whose trailing group is fully deleted.

cpp/src/format/bridge/rust/src/paimon_bridgeimpl.rs
yanbinyang: Fixed in [cad28fc1](https://github.com/milvus-io/milvus-storage/commit/cad28fc1bc8fb201fe2b5c15c59276fb98f0b0b3). Auto routing now rejects non-DV deduplicate splits when a raw file has no delete_row_count. Append-only, DV, and compacted first-row routes remain allowed. The tests cover both the rejected and allowed cases.

cpp/src/format/bridge/rust/src/paimon_bridgeimpl.rs
```diff
@@ -0,0 +1,1291 @@
...
+    }) {
+        return Some((
+            RouteReason::UnsupportedFormat,
+            "direct-file currently supports Parquet and Vortex data files only".to_string(),
+        ));
+    }
+    None
+}
+
+fn decide_route(
```
jiaqizho, line 453: Partitioned tables can pass this direct-file check, but partition columns are not stored in the data file. Paimon's reader rebuilds them from DataSplit.partition(), while this descriptor drops the partition values. Please reject partitioned tables here or preserve and materialize the partition values.

cpp/src/format/paimon/paimon_format_reader.cpp
```diff
@@ -0,0 +1,523 @@
...
+arrow::Result<std::shared_ptr<FormatReader>> PaimonFormatReader::clone_reader() {
+  ARROW_ASSIGN_OR_RAISE(auto cloned,
+                        MetaTrait::create_from_metadata(metadata_, file_, read_schema_, needed_columns_, predicate_));
+  ARROW_RETURN_NOT_OK(cloned->open());
+  return std::static_pointer_cast<FormatReader>(cloned);
+}
+
+std::shared_ptr<arrow::Schema> PaimonFormatReader::get_schema() const { return metadata_->file_schema; }
+
+arrow::Status PaimonFormatReader::set_predicate(const std::string& predicate) {
```
jiaqizho: Just ignore this overwrite function. remove it is fine.

cpp/src/format/paimon/paimon_format_reader.cpp
```diff
@@ -0,0 +1,523 @@
...
+  }
+  for (size_t i = 0; i < batches.size(); ++i) {
+    const auto& physical = metadata_->payload.direct_physical_row_groups[indices[i]];
+    ARROW_ASSIGN_OR_RAISE(auto filtered, filter_direct_batch(batches[i], physical.start_offset));
+    output.push_back(std::move(filtered));
+  }
+  return output;
+}
+
+uint64_t PaimonFormatReader::logical_to_physical(uint64_t logical_offset) const {
```
jiaqizho: The offsets passed to take are already sorted. You can assert this assumtion under #ifndef NDEBUG, change logical_to_physical to accept the complete offset array, and scan it together with the sorted deletions so the complexity becomes O(N + D). Do not add a hashmap to the reader for this optimization. The deletion vector can be very large, and the extra memory is not acceptable.

cpp/test/format/format_reader_cache_test.cpp
```diff
@@ -39,6 +39,7 @@
 #include "milvus-storage/format/iceberg/iceberg_common.h"
 #include "milvus-storage/format/iceberg/iceberg_format_reader.h"
 #include "milvus-storage/format/lance/lance_table_reader.h"
+#include "milvus-storage/format/paimon/paimon_format_reader.h"
```
jiaqizho, line 42: also need add paimon into 
```
INSTANTIATE_TEST_SUITE_P(
    Formats,
    FormatReaderMetadataCacheParamTest,
    ::testing::Values(LOON_FORMAT_PARQUET, LOON_FORMAT_VORTEX, LOON_FORMAT_LANCE_TABLE, LOON_FORMAT_ICEBERG_TABLE));
```

And check others TEST_P

cpp/src/format/bridge/rust/src/paimon_bridgeimpl.rs
```diff
@@ -0,0 +1,1291 @@
...
+            .copy_with_time_travel(HashMap::from([(
+                "scan.snapshot-id".to_string(),
+                snapshot_id.to_string(),
+            )]))
+            .await?;
+        // The probe above already proved existence; reaching this branch means
+        // the snapshot disappeared in between (expiration race) or its content
+        // is unreadable. Both are terminal states for this descriptor.
+        ensure!(
+            table.has_resolved_travel_snapshot(),
```
xiaofan-luan: [P1] This still loses retryable semantics after the existence probe. `exists()` and `copy_with_time_travel()` are separate storage operations: the probe can return `true`, then the later snapshot/schema read can fail with a timeout, 503, throttling, etc. Paimon 0.3's time-travel path swallows that resolution error and returns an unresolved table, so this check turns a retryable failure into `Invalid("snapshot no longer exists; refresh...")`. It also mislabels corrupt snapshot content as non-existence; the current `corrupt_snapshot_content_does_not_claim_nonexistence_bounds` test codifies that wrong behavior.

Please resolve the pin strictly: use `SnapshotManager::get_snapshot(snapshot_id).await`, load `SchemaManager::schema(snapshot.schema_id()).await`, add `scan.snapshot-id` to that historical schema's options, and construct `Table` from it. Only a confirmed `NotFound` should become terminal `Invalid` with refresh advice; temporary storage failures must remain retryable, and malformed metadata should remain a corruption/invalid-data error without claiming expiration. Please add a fault-injection test where the existence check succeeds but the subsequent snapshot read fails transiently, and assert retryable classification with no refresh advice.

xiaofan-luan (commented):
One blocking snapshot-pinning error-classification issue remains.

xiaofan-luan (commented):
Strong PR overall. This is a thoughtful, production-oriented first Paimon integration rather than a thin format switch. I especially like that the direct-file routing is fail-closed: splits that require Paimon merge semantics, schema/data evolution, row-range handling, vector sidecars, or unsupported formats are rejected instead of risking silently incorrect rows. Reusing the existing Parquet/Vortex readers keeps the common reader contract aligned with the other external formats, while the versioned descriptor, deletion-vector handling across chunk/range/take, storage-option forwarding, and the recent typed error-bridge cleanup are all solid design choices.

The test matrix is also strong for an initial integration: it exercises append-only and primary-key routing, Parquet and Vortex, deletion vectors, malformed metadata, snapshot selection, and unsupported paths. The implementation has clearly been developed with correctness and future extensibility in mind.

Once the outstanding blocking review comments—especially strict snapshot resolution/error classification—and their regression tests are addressed, I would recommend approving this PR.

cpp/src/format/paimon/paimon_common.cpp
```diff
@@ -0,0 +1,108 @@
...
+  if (config.cloud_provider == kCloudProviderAliyun) {
+    set("fs.oss.endpoint", endpoint());
+    set("fs.oss.accessKeyId", config.access_key_id);
+    set("fs.oss.accessKeySecret", config.access_key_value);
+    return options;
+  }
+  if (config.cloud_provider == kCloudProviderAzure) {
+    set("azure.endpoint", endpoint());
+    set("azure.account-name", config.access_key_id);
+    if (config.use_iam) {
```
liliu-z, line 63: The Azure use_iam path forwards only service-principal env vars (client-id/secret/tenant-id/authority-host). On an AKS workload-identity cluster the credential is a projected federated token (AZURE_FEDERATED_TOKEN_FILE) with no client secret, and opendal 0.58's AzdlsConfig has no field to carry a federated token, so ToStorageOptions emits an incomplete client-credentials config that fails at read time — unlike role-ARN and GCS impersonation, which fail closed with NotImplemented. Either resolve the federated token to a bearer credential or reject workload-identity setups explicitly.

cpp/src/format/paimon/paimon_common.cpp
```diff
@@ -0,0 +1,108 @@
...
+      if (const auto* value = std::getenv("AZURE_AUTHORITY_HOST"))
+        set("azure.authority-host", value);
+    } else {
+      set("azure.account-key", config.access_key_value);
+    }
+    return options;
+  }
+  if (config.cloud_provider == kCloudProviderGCP) {
+    set("gcs.endpoint", endpoint());
+    set("gcs.credential", config.gcp_credential_json);
```
liliu-z: GCS static credentials are written as raw JSON into gcs.credential, but opendal's GcsConfig.credential must be base64-encoded. The backend calls StaticCredentialProvider::from_base64(content) and, on decode failure, silently omits the credential provider (no error), so a read configured via gcp_credential_json drops the credential and falls back to the default chain and fails to authenticate off-GCP. Base64-encode the JSON before setting gcs.credential (paimon-rust exposes no raw-JSON inline field, only gcs.credential-path to a file) and fix the assertion at paimon_storage_options_test.cpp:126.

cpp/src/format/bridge/rust/src/paimon_bridgeimpl.rs
yanbinyang: Fixed in [03dc9304](https://github.com/milvus-io/milvus-storage/commit/03dc9304e855166d6c6feb7da174af1374afc792). Direct-file planning now rejects partitioned tables because partition values are reconstructed from DataSplit metadata and are not stored in the data files.

cpp/src/format/paimon/paimon_format_reader.cpp
yanbinyang: Removed in [41db999f](https://github.com/milvus-io/milvus-storage/commit/41db999fe8b2dd491b2cbc32d9d0ce11adf7d71c). Paimon now uses the default no-op predicate hook for both cached and uncached readers.

cpp/src/format/paimon/paimon_format_reader.cpp
yanbinyang: Updated in [99e6f79d](https://github.com/milvus-io/milvus-storage/commit/99e6f79da54663aae8ed0ea528a1bef913f214f3). Sorted logical offsets and deletion positions are now merged in O(N + D), without adding another deletion-vector index.

cpp/test/format/format_reader_cache_test.cpp
yanbinyang: Added in [176e5dae](https://github.com/milvus-io/milvus-storage/commit/176e5dae0b0ee02714d6c6b29864fb683c4b0817).

cpp/src/format/bridge/rust/src/paimon_bridgeimpl.rs
yanbinyang: Fixed in [9e05d31f](https://github.com/milvus-io/milvus-storage/commit/9e05d31fa1834b461efd151910f72f9d3cf21915). Pinned snapshots and historical schemas are now resolved through the strict manager APIs. Only confirmed NotFound gets refresh advice; transient and corrupt-metadata errors retain their respective classifications. Regression tests cover all three cases.

cpp/src/format/paimon/paimon_common.cpp
yanbinyang: Fixed in [4dde08f0](https://github.com/milvus-io/milvus-storage/commit/4dde08f0d1abe5fcde9d3e39467de4fd51fba0fb). Workload identity now fails closed when a federated token is present and a complete service-principal credential tuple is unavailable. A complete service-principal configuration remains supported.

cpp/src/format/paimon/paimon_common.cpp
yanbinyang: Fixed in [c2702786](https://github.com/milvus-io/milvus-storage/commit/c2702786d2d8cb67c6b6d6f1e2d05ea712a649fe). The inline JSON credential is now base64-encoded before being passed as gcs.credential, and the test verifies the encoded value.

cpp/src/format/bridge/rust/src/paimon_bridgeimpl.rs
yanbinyang: Follow-up in [ac1e5444](https://github.com/milvus-io/milvus-storage/commit/ac1e54442d84d78b15ebcf34f866d8c5df84641e): the strict snapshot path now combines the historical schema with the current table options and validates the final scan options after injecting scan.snapshot-id. Tests cover schema reconstruction and unsupported scan options for both latest and pinned scans.

PR #603, opened 2026-07-29
enhance: define storage error taxonomy and consumer contract

xiaofan-luan:
## Summary

Part 1 of the six-PR split of the original #603.

- Define one X-macro table for exact storage code, name, category, and transient/retryable hint.
- Keep typed `ExtendStatusDetail` authoritative over Arrow fallback classification.
- Preserve the public `LoonFFIResult` layout as `(err_code, message)`.
- Export code/name/category/retryable queries so consumers do not infer retry from Arrow StatusCode.
- Keep unknown, retired, and untyped failures conservative.
- Document in the C API example that `retryable=true` describes an observed transient cause, not whether a stateful operation is safe to replay.

Refs #640.

## Review scope

This foundation PR contains only the native taxonomy, mappings, exports, and focused tests. Cloud backends, Rust format bridges, writer rollback, Packed validation, and public reader/FFI lifetime changes are split into the five dependent PRs below.

Compatibility names already present on `main` remain only while a dependent layer still uses them. Each is retired in the first layer after its final producer or consumer migrates: code 106 in #647, code 54 in #650, and the remaining C/C++ aliases in #651.

## Validation

- Release C++ build passed.
- Focused taxonomy/FFI C++ tests passed.
- Pure-C ABI tests passed.
- `git diff --check` and clang-format 17 passed.

## Stack

Review and merge in order:

1. #603 — taxonomy foundation
2. #647 — cloud filesystems, credentials, and lifecycle
3. #648 — Rust format error envelope
4. #649 — writer failure and output ownership
5. #650 — Packed metadata and missing-field projections
6. #651 — public API lifetime and consumer migration

Each PR is based on the preceding stack branch. After one PR merges, rebase the remaining stack onto the updated `main` (required for squash merges), update the synthetic branches with force-with-lease, and then retarget the next PR to `main`.


cpp/include/milvus-storage/ffi_internal/ffi_error_code.h
```diff
@@ -14,6 +14,30 @@
...
+// Transient). A single `retryable` bool is deliberately NOT the primary field:
+// User and Permanent are both non-retriable but need opposite handling --
+// report a User error back to whoever made the request, alert an operator for
+// a Permanent one.
+//
+// The category is a pure function of the error code, so it needs no room in
+// LoonFFIResult and a consumer can classify a code it has never seen.
+// ===========================================================================
+#define LOON_ERROR_CATEGORY_UNKNOWN 0
+#define LOON_ERROR_CATEGORY_USER 1
```
jiaqizho, line 34: For milvus-storage, the upper layers are segcore and some Go code that calls it through FFI. Is the purpose of classifying errors as User here to consistently report configuration errors to Milvus users?

If so, at the storage layer we cannot actually distinguish whether an error was caused by invalid user input or by an internal Milvus error. 

For example, `LOON_INVALID_PROPERTIES` is not necessarily caused by properties set by a Milvus user; most of these properties are hard-coded in segcore. Similarly, `LOON_INVALID_ARGS` could also be caused by segcore calling the API incorrectly, in which case it would be an internal logic error.

xiaofan-luan: You're right, and this is the sharpest problem with the PR as written. Let me answer the question directly and then say what I think the fix is.

**No — the category is not meant to say "show this to a human."** It says where the fault lies *relative to storage*: outside this library, in the request we were handed. Whether that request originated from a person typing a URI or from segcore passing a null pointer is knowledge only segcore has, and you're correct that we cannot recover it here.

The problem is that the PR doesn't stop at saying that. `ToSegcoreErrorCode` maps every `User` code to `2042 InvalidParameter`, and 2042 in milvus *does* mean "the end user's parameter is bad" and gets surfaced as such. So the mapping asserts exactly the thing you're saying storage can't know. That's a real defect, not a naming quibble.

Concretely, the `User` codes today are two different things:

| Code | What actually happened | Who must act |
|---|---|---|
| `LOON_SOURCE_NOT_FOUND` 13, `LOON_SOURCE_ACCESS_DENIED` 14, `LOON_SOURCE_URI_INVALID` 116 | a location string the user supplied | the end user |
| `LOON_INVALID_ARGS` 1, `LOON_INVALID_PROPERTIES` 7, `PackedInvalidArgs` 50 | the API was called wrong | a developer |

The first group is safe because those codes are only minted at the two entry points that take a user-supplied location (`loon_exttable_explore`, `loon_exttable_get_file_info`) — storage doesn't guess there, the entry point *knows*. The second group is exactly your case: `LOON_INVALID_PROPERTIES` fires on `loon_properties_create`'s duplicate-key and bad-index checks, and most of those properties are hard-coded in segcore, so calling it a user error is wrong.

By this PR's own stated rule — one category per *consumer action* — those two rows need different categories, because "return it to the requester, don't alert" and "alert a developer, this is a bug" are opposite actions. So the fix I'd propose:

- **1, 7, 50 become `Permanent`** ("our bug or the data is gone → alert a developer, never retry"), and map to `2001 UnexpectedError` rather than `2042 InvalidParameter`. Non-retriable either way, so no retry-behaviour change — but segcore is told "you called us wrong" instead of milvus telling its user "your parameter is bad."
- **`User` stays, and shrinks to 13/14/116** — the codes where storage provably knows the input was a user-supplied location. `User → 2042` then becomes true rather than aspirational, and the invariant test can keep enforcing it.

This costs no new category and removes the guess you're pointing at. It does change the segcore code for `LOON_INVALID_ARGS`, which is the highest-frequency code at the boundary, so I'd rather have your agreement before pushing it than surprise you with it.

One thing I'd push back on slightly: `LOON_NOT_SUPPORT` (9) is also currently `User`, and I think that one is defensible where it is — asking for a format this build lacks is a property of the request, and no developer action fixes it. Happy to be argued out of that too.

Separately, thanks for reading closely enough to hit this — the invariant tests in this PR would never have caught it, since they only check that the tables agree with each other.

xiaofan-luan:
**On `DefaultRetryableForExtendStatusCode` (raised in rounds 4-7) — settling it here.**

The mechanical point is correct: the old symbol was out-of-line, so the `[[deprecated]] inline` forwarder an earlier revision carried restored *source* compatibility only — the exported symbol was gone either way. The shim was removed rather than "fixed" because the obligation itself does not exist here:

- **Zero call sites.** Not in this repo, not in milvus (the only grep hit is the vendored copy of this header itself), not in any downstream we ship.
- **This library makes no ABI promise.** There is no SOVERSION, and every consumer (milvus, via conan) pins a commit and rebuilds from source in lockstep.
- Keeping a permanently dead compatibility surface for a never-called function is exactly the kind of second source of truth this PR exists to remove.

If an out-of-tree consumer ever surfaces, the fix on their side is a one-line rename. Please treat this as settled for this PR; happy to revisit if someone can name a real consumer.

**On structured Java exceptions (round 7, P2)** — agreed on the end state, and it is deliberately a follow-up rather than part of this PR: `java/` currently has no exception classes at all, so doing it right needs a new `LoonStorageException` carrying `errCode` / `category` / `retryable`, JNI construction plumbing, and Java build wiring — a reviewable unit of its own. Today's boundary maps invalid-args → `IllegalArgumentException` and OOM → `OutOfMemoryError`; everything else is a `RuntimeException` whose code lives only in the message text. Will be tracked as a follow-up issue.

xiaofan-luan:
**On the `DefaultRetryableForExtendStatusCode` API/ABI break — accepting it explicitly, per the waive option.**

The finding is factually right and I want to record it as such rather than argue it away. At the merge base (`63c29c67`) the function is declared in the installed header (`extend_status.h:101`) and defined **out-of-line** (`extend_status.cpp:119`), so it is a real exported C++ symbol, and the current dylib exports only the new name — `nm -gU libmilvus-storage.dylib | grep DefaultRetryable` returns nothing. Anything compiled against the old header and linked against the new library gets an undefined symbol. That is a break, not a theoretical one.

Accepting it, for three reasons that are specific to this symbol rather than general hand-waving:

1. **No consumer exists.** At the merge base the only callers are inside this repo (`result_c.cpp:103` and three assertions in `azure_error_classification_test.cpp`), all renamed by this PR. milvus consumes the **C** ABI (`loon_*`), not the C++ API; the sole grep hit there is the vendored copy of this header. No out-of-tree consumer has been identified in four rounds of looking.
2. **It was already invisible in the shipped configuration.** `cpp/ffi_exports.map` lists only `loon_*` symbols, so under `WITH_PYTHON_BINDING` the version script hid this mangled C++ symbol at the merge base too.
3. **This library makes no ABI promise.** No SOVERSION; consumers pin a commit via conan and rebuild in lockstep.

Restoring a non-inline forwarder would mean keeping a permanently dead exported symbol for a function nothing calls — the kind of second surface this PR exists to remove. If a real consumer surfaces the fix on their side is a one-line rename. Marking this waived.

jiaqizho:
Is the current PR ready for review?

xiaofan-luan:
**Scope note**

本 PR 聚焦存储层的错误分类体系本身：单一 X-macro 错误表（码值 / category / S3 对应 / segcore 落点的唯一事实源）、分类在 producer → 各层 → FFI/segcore 边界上的保真传递，以及配套的 CI 门禁（文档漂移、死码、ratchet、Rust 单测接入）。

FFI 接口的形态设计不是本 PR 的关注点。这里改动到的 FFI 部分只是让错误码/category 能跨 C ABI 完整传出去；接口本身的取舍请另开 issue 或 PR 讨论，以免评审焦点被带偏。

czs007:
Reviewed the taxonomy and the mapping against `main`. The fact/policy split, the X-macro table with a mandatory category column, and the provenance gate on `UserSourceErrorCodeFromStatus` all hold up well. Three things I could not reconcile.

---

### 1. `AwsErrorNoSuchUpload` silently loses its retryable classification

`main` gave it a dedicated arm with a specific rationale:

```cpp
case ExtendStatusCode::AwsErrorNoSuchUpload:
  // The SDK has exhausted retries for the failed multipart upload state,
  // but an outer operation retry can create a fresh upload and succeed.
  return milvus::StorageTransientError;  // 2045
```

This PR folds it into the `AwsErrorNotFound` arm, where it inherits reasoning written for a different failure:

> The object/bucket is gone ... "data missing" (stale loadinfo, GC'd file) ... Never transient/2045: **a retry/reroute hits the same shared object store and fails identically**.

None of that describes `NoSuchUpload`: nothing was ever completed, no data is missing, and a retry does **not** hit the same thing — it creates a new upload id.

The producer in this same PR says so explicitly (`s3_internal.h:259-262`):

```cpp
case Aws::S3::S3Errors::NO_SUCH_UPLOAD:
  // A dead upload id cannot be repaired by replaying the same multipart
  // request. The owner of the upload may choose to start a new one.
```

And "start a new one" is precisely what this taxonomy's own `Retryable` contract prescribes — *"a failed stateful writer must be destroyed and recreated, never reused"* — which the PR description restates as *"Recovery must destroy the failed writer and create a new writer."*

**Impact:** `loon_ffi_is_retryable_errcode` flips 1 -> 0 for this code, so milvus's outer `retry.Do` around the packed writer (`internal/flushcommon/syncmgr/pack_writer_v3.go:227`) stops retrying an operation that a fresh upload would complete. The change is not mentioned in the PR description, which documents comparable decisions in detail — it reads as an unintended consequence of merging the case arms.

**Suggested:** category `RETRYABLE`, mapping back to `StorageTransientError`.

---

### 2. Value 54 is reused, against this PR's own rule, for a code that is never produced

```
54  PackedArrowError   (System, real producer on main: packed/reader.cpp:370)
 -> PackedIOTransient  (Retryable, listed in RESERVED_PRODUCERLESS -- "never produced again")
```

Meanwhile 106 gets the opposite treatment in the same header:

```c
// 106 retired (was LOON_AWS_ERROR_NON_RETRYABLE). Never reuse. ...
// while `loon_errcode_aws_non_retryable` stays exported as a tombstone so an
// older binding does not fail to load against a newer library
```

and the table's own contract says `code` *"is the C ABI contract, never renumber."*

Three things make this worth changing rather than documenting:

- `PackedArrowError` was **in service** (not a never-emitted reservation), so 54 has shipped meaning "permanent internal error"
- the semantic flip runs in the dangerous direction: a stale consumer asking `loon_ffi_is_retryable_errcode(54)` goes from **false to true**
- the reuse buys nothing — `PackedIOTransient` is declared producer-less, and 114 / 116 / 123+ are free

**Suggested:** give `PackedIOTransient` a fresh value; retire 54 the way 106 was retired.

---

### 3. `User` carries two different meanings of "user"

`LOON_SOURCE_INVALID` (13) is the end user: *"the external source **the user named** is unusable ... the credentials **they supplied** were rejected."* It is minted only through `SourceProvenance`, and `docs/error-codes.md:225` is explicit — *"Do not call it anywhere else."*

`LOON_USER_INVALID_ARGUMENT` (10) is a different caller. Its producers are FFI argument validation in `column_groups_c.cpp` / `exttable_c.cpp`:

```c
RETURN_ERROR(LOON_USER_INVALID_ARGUMENT, "Invalid arguments: col_lens must be greater than 0");
```

and the doc describes it as *"a structurally valid public API call contains an invalid caller-owned value ... malformed packed-writer offsets, zero parallelism."* No end user reaches packed-writer offsets — the caller here is milvus, and there is no provenance gate on this path at all.

That leaves the category with two frames under one name, and `docs/error-codes.md:27` — *"`User` is minted only where ownership is known"* — holds for 13 but not for 10.

**Impact on the consumer side:** milvus reads `User` as "my caller's fault", and milvus's caller is the end user. A milvus-side null pointer therefore surfaces as an input error, which makes the proxy abort cross-replica failover and tell the user their request is invalid.

**Suggested:** move 10 to `System` — it matches that category's own definition (*"configuration, missing internal data, unsupported operation, **bug**"*). `User` then means exactly what the provenance gate mints, and the name stops being layer-relative.

---

*(For the record, one downstream note rather than a review point: `PackedStorageIO` and `AwsErrorNonRetryable` are referenced by `internal/core/unittest/test_storage_error_code.cpp` on milvus master and 3.0, so the pin bump needs a matching milvus PR. That's ours to handle.)*

czs007:
## CI triage for `b2af27b` — 6 failing checks, 5 independent causes

I reproduced each one locally against the PR head. Four belong to this PR, one is pre-existing and unrelated, one is a consequence.

### 1. `error-handling-ratchet` / `check` — exit 141 (SIGPIPE). **This is a race in code this PR adds.**

```
bash cpp/scripts/error_handling_ratchet.sh check /tmp/base_baseline.tsv
##[error]Process completed with exit code 141
```

141 = 128 + 13 = SIGPIPE. `cpp/scripts/error_handling_ratchet.sh:97`, in the new `verify_stripper()`:

```bash
set -euo pipefail
probe_file=$(git -c core.quotePath=false ls-files -- 'cpp/src/**.cpp' | head -n 1)
```

`head -n 1` exits after the first line and closes the read end. If `git ls-files` has not finished writing by then it dies of SIGPIPE, `pipefail` makes the pipeline's status 141, and `set -e` aborts the script — after `probe_file` has already been assigned, so `bash -x` shows the assignment succeed and the script die on that same line.

It is timing-dependent, which is why it is worth fixing rather than re-running: **the identical script at `aec6c44` passed this job on 08-22**, and `b2af27b` failed it on 08-23 with no change to the file. Locally it fails on a cold worktree and passes on every warm re-run:

```
$ for i in 1 2 3; do bash -c 'set -euo pipefail; p=$(git ls-files -- "cpp/src/**.cpp" | head -n 1); echo ok=$p'; echo rc=$?; done
rc=141
ok=cpp/src/common/arrow_util.cpp
rc=0
ok=cpp/src/common/arrow_util.cpp
rc=0
```

A gate that fails at random under `pipefail` is worse than no gate, so the producer needs to run to EOF. Minimal fix — `awk` drains stdin, so nothing ever gets SIGPIPE:

```bash
probe_file=$(git -c core.quotePath=false ls-files -- 'cpp/src/**.cpp' | awk 'NR==1')
```

(`|| true` does not help: the failure is the pipeline status under `pipefail`, and suppressing it would also mask a genuinely empty result, which the existing `[ -z "$probe_file" ]` check is there to catch.) This is the only early-exit pipe consumer in the script.

### 2. `clang-format` / `check` — `make check-format` fails

`-Wclang-format-violations` in two files this PR touches:

- `cpp/src/common/metadata.cpp`: 99, 100, 299, 407, 408
- `cpp/src/writer.cpp`: 221, 224

`make format` fixes it.

### 3. `lint` (Python CI) — `black --check`

```
would reformat python/tests/test_error_taxonomy.py
1 file would be reformatted, 15 files would be left unchanged.
make: *** [Makefile:9: format-check] Error 1
```

Two over-length expressions in the new file (the `header = Path(...)` assignment and the `re.findall(...)` call) need wrapping.

### 4. `unittest` — 20 failing tests, one root cause + one semantic question

`1416 tests from 97 test suites ran … 20 FAILED TESTS`:

| test | count |
|---|---|
| `MetadataTest.{TestGroupFieldIDListSerde, TestFieldIDList, TestRowGroupMetadata}` | 3 |
| `SegmentReaderTextTest.{CreateWithNullFs, CreateWithNullSchema, CreateWithNullColumnGroups, OpenWithNullFs, OpenWithNullSchema, OpenWithNullManifest}` | 6 |
| `Formats/FormatReaderMetadataCacheParamTest.{GetOrOpenNullMetadataDoesNotPoisonCache, GetOrOpenAsyncFailuresDoNotPoisonCacheAndAllowRetry}` | 2 × 5 params |
| `VortexLocalFormatTest.PublicArgumentValidationReturnsPlainInvalid` | 1 |

**Root cause (mechanical).** This PR converts a batch of `arrow::Status::Invalid(...)` sites to `MakeExtendErrorMsg(code, ...)`, which produces an **`IOError` carrying an `ExtendStatusDetail`**, not an `Invalid`. The assertions still read `status().IsInvalid()` (and in two suites, `UnwrapStatus(...) == nullptr`), so they all flip:

```
Value of: RowGroupMetadata::TryDeserialize("invalid").status().IsInvalid()
  Actual: false / Expected: true

IOError: Format reader metadata loader returned null metadata. Detail: InternalInvariantViolated:
```

All four test files are already modified by this PR, so this reads as an incomplete sweep of the assertions rather than an unexpected behaviour change. Worth noting that the verification section lists a focused 144-test suite; the full `milvus_test` binary is what catches these.

**One case is not just an assertion update.** `cpp/test/common/metadata_test.cpp:69`:

```cpp
auto classified = WrapExtendError(ExtendStatusCode::PackedMetadataCorrupted,
                                  "Invalid persisted group field id metadata", persisted_error);
auto detail = ExtendStatusDetail::UnwrapStatus(classified);
EXPECT_EQ(detail->code(), ExtendStatusCode::PackedMetadataCorrupted);   // 0x0034
// actual: 0x0075 == DataCorrupted, i.e. the cause's own code
```

`metadata.cpp:58` now returns `MakeExtendErrorMsg(ExtendStatusCode::DataCorrupted, ...)`, so the cause already carries a detail and `WrapExtendError` keeps the inner code instead of applying the outer one. That is a contract decision for the taxonomy, and it should be stated in `docs/error-codes.md` either way:

- *inner code wins when the cause is already classified* → the expectation becomes `DataCorrupted`, and `WrapExtendError`'s outer code is documented as a fallback for unclassified causes only; or
- *outer code wins* → `WrapExtendError` has to overwrite, and every existing wrap site needs re-checking for whether its outer code is really the more specific one.

Given the PR's stated principle that a code reports an observed fact, the first reading is the more consistent one — the format-aware decoder that confirmed the corruption is the more authoritative observer. Either way this should be an explicit decision recorded in the docs, not a by-product of which branch happens to run first.

### 5. `mac-build` — pre-existing, unrelated to this PR

```
ERROR: libavrocpp/1.12.1.1@milvus/dev: Error in source() method, line 46
  NotFoundException: Not found: https://dlcdn.apache.org/avro/avro-1.12.1/avro-src-1.12.1.tar.gz
```

`dlcdn.apache.org` only serves current releases and avro 1.12.1 has been moved to `archive.apache.org`. Workflow history for `cpp-macos-optional`: last success on this branch was 08-11; every run from 08-13 onward fails on this same 404. The fix belongs in the `libavrocpp/1.12.1.1@milvus/dev` conan recipe (archive URL or a mirror), not in this PR — the job is marked optional.

### 6. `codecov/patch/cpp` (`0.00% of diff hit`) and `codecov/project/cpp` (`0.00%`)

Downstream of #4 — the cpp coverage report is not usable while `milvus_test` fails. The python codecov statuses pass. These should clear once the tests do.

---

### Minor

`python3 cpp/scripts/check_error_table.py` on `b2af27b` prints:

```
check_error_table: OK -- 33 codes (21 ExtendStatusCode), live producers and reserved tombstones verified; doc rows in sync
```

The PR description's verification section says 35 codes / 23 `ExtendStatusCode`, so that section is stale relative to the pushed head.

---

**Summary:** 1–3 are mechanical. 4 is mechanical except for the `WrapExtendError` code-precedence question, which is worth settling explicitly before the assertions are rewritten around whichever behaviour is currently in place. 5 blocks nothing here and needs a separate recipe fix. 6 follows 4.

cpp/src/filesystem/s3/s3_filesystem.cpp
```diff
@@ -1247,37 +1494,59 @@ class CustomOutputStream final : public arrow::io::OutputStream {
...
     }
 
     return arrow::Status::OK();
   }
 
   arrow::Status CleanupIfFailed(Status status) {
     if (!status.ok()) {
-      ARROW_RETURN_NOT_OK(CleanupAfterClose());
-      return status;
+      return CloseAfterFailure(std::move(status));
```
jiaqizho: This logical is different. it will make origin status lost.

cpp/src/filesystem/s3/s3_filesystem.cpp
```diff
@@ -874,10 +926,12 @@ class ObjectCrtInputFile final : public arrow::io::RandomAccessFile, public NonB
...
           }
-          ctx->future.MarkFinished(bytes_read);
+          finish(bytes_read);
         });
-    // Keep executor selection at the caller-owned continuation boundary.
-    return ctx->future;
+    // Arrow callbacks run inline by default. Transfer the user-visible future
+    // so releasing its last filesystem owner cannot finalize the CRT client
+    // from inside this native callback while the operation lease is active.
+    return arrow::io::default_io_context().executor()->Transfer(ctx->future);
```
jiaqizho: never do that.

Current CRT callback can't do in CRT thread which is an implementation issue within CRT client itself. However, this is already strictly enforced at the caller level.

Use `arrow::io::default_io_context()` to execute the callback will make decoder/uncompress happen in arrow thread.  In the current design, this CPU operation should always take place within a folly executor.

cpp/include/milvus-storage/filesystem/s3/provider/credential_resolution.h
```diff
@@ -0,0 +1,162 @@
...
+///
+/// ResolveForRequest returns the credentials and their typed failure as one
+/// value. S3's Standard and CRT holders call it before handing a client to an
+/// object request, so an expired credential whose refresh failed never falls
+/// through as an anonymous request and later masquerades as AccessDenied.
+class RequestCredentialsResolver {
+  public:
+  virtual ~RequestCredentialsResolver() = default;
+
+  [[nodiscard]] virtual arrow::Result<Aws::Auth::AWSCredentials> ResolveForRequest() = 0;
```
jiaqizho: checkout `GetAWSCredentials`, why we need `ResolveForRequest` to do the same thing?

cpp/src/filesystem/s3/s3_client.cpp
```diff
@@ -379,6 +384,25 @@ arrow::Result<S3ClientLock> S3ClientHolder::Lock() {
...
+  std::shared_ptr<S3Client> client;
+  std::shared_ptr<RequestCredentialsResolver> credentials_resolver;
+  {
+    std::unique_lock lock(mutex_);
+    client = client_;
+    credentials_resolver = credentials_resolver_;
+  }
+  if (credentials_resolver != nullptr) {
+    auto credentials = credentials_resolver->ResolveForRequest();
+    if (!credentials.ok()) {
```
jiaqizho: ARROW_RETURN_NOT_OK

cpp/src/filesystem/s3/s3_client.cpp
```diff
@@ -379,6 +384,25 @@ arrow::Result<S3ClientLock> S3ClientHolder::Lock() {
...
+  // the barrier and destruct after Aws::ShutdownAPI().
+  std::shared_ptr<S3Client> client;
+  std::shared_ptr<RequestCredentialsResolver> credentials_resolver;
+  {
+    std::unique_lock lock(mutex_);
+    client = client_;
+    credentials_resolver = credentials_resolver_;
+  }
+  if (credentials_resolver != nullptr) {
+    auto credentials = credentials_resolver->ResolveForRequest();
```
jiaqizho: Every s3 request requires a call to `ResolveForRequest`.

cpp/include/milvus-storage/filesystem/s3/provider/credential_resolution.h
jiaqizho: i just found u change the `GetAWSCredentials`. So u just want get the error message before we call `GetAWSCredentials`? 

I do think add a error log in `GetAWSCredentials` rather than pass the `RequestCredentialsResolver` to `s3_client/s3_crt_client`?

cpp/src/filesystem/s3/s3_filesystem.cpp
```diff
@@ -406,6 +427,19 @@ std::string FormatRange(int64_t start, int64_t length) {
   return ss.str();
 }
 
+template <typename ErrorType>
+arrow::Status ObjectReadErrorToStatus(const S3Path& path,
```
jiaqizho: If we need more detail in error log. Just move it into s3_internal? As defined by the `ErrorToStatus` location.

cpp/src/filesystem/s3/s3_filesystem.cpp
```diff
@@ -1247,37 +1507,59 @@ class CustomOutputStream final : public arrow::io::OutputStream {
...
   arrow::Status Close() override {
+    auto check_status = writer_status_->Check();
+    if (!check_status.ok()) {
+      return CloseAfterFailure(std::move(check_status));
+    }
+    return writer_status_->RecordFirstFailure(CloseImpl());
+  }
+
+  arrow::Status CloseAfterFailure(arrow::Status primary_status = arrow::Status::OK()) {
+    (void)writer_status_->RecordFirstFailure(primary_status);
```
jiaqizho: Current close already failed. Why still call `RecordFirstFailure` will always not record the `primary_status`?

cpp/src/filesystem/s3/s3_filesystem.cpp
```diff
@@ -1247,37 +1507,59 @@ class CustomOutputStream final : public arrow::io::OutputStream {
...
-      return status;
+      return CloseAfterFailure(std::move(status));
     }
     return arrow::Status::OK();
   }
 
   arrow::Status Close() override {
+    auto check_status = writer_status_->Check();
+    if (!check_status.ok()) {
+      return CloseAfterFailure(std::move(check_status));
```
jiaqizho: CloseImpl already have this logical at begin.
```
auto check_status = writer_status_->Check();
if (!writer_status_->Check().ok()) {
   return CloseAfterFailure();
}
```
So in `Close()`, we don't need call it twice.

cpp/src/filesystem/s3/s3_filesystem.cpp
```diff
@@ -1599,43 +1989,46 @@ class CustomOutputStream final : public arrow::io::OutputStream {
...
-      // Besides, it's not generally sound to write to an OutputStream from
-      // several threads at once.
-      fut.MarkFinished(state->status);
+    // Publishing may allocate. If it fails the exception propagates and the
+    // process dies. That is deliberate: the publish slot above has already been
+    // consumed, so anything that swallows the failure here leaves
+    // pending_uploads_completed unfinished with no later completion able to
+    // claim it, and every FlushAsync/CloseAsync waiter blocks forever. A crash
+    // is restartable; that hang is not.
+    auto completion_status = arrow::io::internal::ObserveBackgroundIOStatus(writer_status.get(), status);
```
jiaqizho: in `ObserveBackgroundIOStatus(writer_status, status)`
```
  if (status.ok()) {
    return arrow::Status::OK();
  }
```

So if current outputstream have been cancel.Then writer_status do have a error status. But it won't do `    ock(state->mutex);`. `upload` and `abort` can happen in same time.

Then the upload is success the future will got a status.ok in callback.

cpp/include/milvus-storage/common/macro.h
```diff
@@ -18,39 +18,23 @@
...
-  std::cout << "Case-" << #name << " Execute time: (" << name##_duration.count() << " us), ("                \
-            << name##_duration.count() / 1000.0 << " ms), (" << name##_duration.count() / 1000000.0 << " s)" \
-            << std::endl;
-#endif
-
 #define CONCAT_IMPL(x, y) x##y
 
 #define CONCAT(x, y) CONCAT_IMPL(x, y)
 
-#if !defined(LIKELY) && !defined(UNLIKELY)
```
jiaqizho, line 37: no need remove `LIKELY`.

`LIKELY` and `UNLIKELY` are paired definitions.

cpp/src/format/parquet/folly_arrow_executor.cpp
```diff
@@ -15,6 +15,7 @@
 #include "milvus-storage/format/parquet/folly_arrow_executor.h"
 
 #include <exception>
+#include <new>
```
jiaqizho: what for?

PR #604, opened 2026-07-29
enhance: classify Azure storage errors into the ExtendStatusCode taxonomy

### LINKED ISSUE #595: [Enhancement]: Azure/GCS filesystems lack ExtendStatusCode classification (not-found / auth / throttle all surface as plain IOError)

xiaofan-luan:
## Problem

The ExtendStatusCode taxonomy (#574 transients 107–110, #575 permanent S3 tags 104–106) only covers the S3 path. On Azure, `ExceptionToStatus` tagged exactly one condition — `PreconditionFailed`, plus `Conflict`+`BlobAlreadyExists` — and everything else fell through to a plain `Status::IOError`:

https://github.com/milvus-io/milvus-storage/blob/d3eebb1ff653f023bb02c885e139d69daca0f1d2/cpp/src/filesystem/azure/azurefs.cc#L360-L375

An untagged status reaches segcore as `StorageError`/2044 — permanent and non-retriable. So on Azure a **throttle or a 503 was reported as a permanent storage error and never retried**, which is the availability-relevant half of the gap. Closes #595.

## Change

`internal::ClassifyAzureError(http_status, error_code)` maps the failure onto the existing taxonomy; `ExceptionToStatus` tags the status when it returns a code and keeps today's plain `IOError` when it does not.

| Azure | ExtendStatusCode | segcore | retry |
|---|---|---|---|
| no response (transport: conn refused/reset, DNS, TLS) | `StorageTransientNetwork` 107 | 2045 | yes |
| 408 Request Timeout | `StorageTransientTimeout` 108 | 2045 | yes |
| 429 Too Many Requests | `StorageTransientThrottling` 109 | 2045 | yes |
| 503 + `ServerBusy` | `StorageTransientThrottling` 109 | 2045 | yes |
| 503 (other), 500, 502, 504 | `StorageTransientService` 110 | 2045 | yes |
| 404 | `AwsErrorNotFound` 104 | 2017 `ObjectNotExist` | no |
| 401, 403 | `AwsErrorAccessDenied` 105 | 2044 | no |
| 412, 409 + `BlobAlreadyExists` | `AwsErrorPreConditionFailed` 103 | 2044 | no | 
| anything else (incl. 409 non-`BlobAlreadyExists`) | untagged | 2044 | no |

Notes on the two judgement calls:

- **Transport failures are the `StatusCode == None` (0) case.** `Azure::Core::Http::TransportException` derives from `RequestFailedException` and `ExceptionToStatus` is already called with it (the `catch` in the HNS check), so it arrives here with `StatusCode` left at its `None` default. That is the discriminator, and it is the one class where "no response at all" makes retry unambiguously right.
- **503 is overloaded on Azure Storage**: `ServerBusy` is throttling, everything else is an availability blip. Both retriable, but keeping them apart lets a consumer apply throttle-aware backoff only where it belongs. Same split #574 settled on for S3.

Codes are cloud-neutral despite the `AwsError*` prefix (per the naming discussion in #574) — no new values needed, as #595 predicted.

The classifier takes the raw HTTP status rather than the SDK exception type specifically so it can be unit-tested **without an Azure account**. Every credential-bearing Azure test is skipped in CI, so a mapping tested only through the filesystem would be untested in practice.

## Two corrections to #595

Both were load-bearing for its proposed scope, and both turn out to be wrong. Verified against `d3eebb1`:

1. **"GCS: zero `MakeExtendError` call sites in the whole subtree. All GCS errors surface as untagged status."** — Not so. `GcpFileSystemProducer` builds **milvus-storage's own** `S3FileSystem`, the one carrying `ErrorToStatus`:

   https://github.com/milvus-io/milvus-storage/blob/d3eebb1ff653f023bb02c885e139d69daca0f1d2/cpp/src/filesystem/gcp/gcp_filesystem_producer.cpp#L255-L259

   (`#include "milvus-storage/filesystem/s3/s3_filesystem.h"` at L41, inside `namespace milvus_storage` at L45.) So the GCS **data path already inherits the full S3 classification**, including the HTTP-status channel that is vendor-neutral. What is genuinely unclassified is only the init path, which this PR fixes as a one-liner: it was stringifying the cause into `Status::Invalid`, turning a credential/network init failure into a data-class error (`DataFormatBroken`/2024) and destroying whatever classification the cause carried. The remaining GCS gap is the exception-*name* channel (`SlowDown`, `XMinioServerNotInitialized` are AWS/MinIO dialect), which is a long tail, not the whole taxonomy.

2. **"`PathNotFound()` returns arrow's generic `PathNotFound` IOError with no `ExtendStatusDetail` and no errno detail, so it cannot reach `ObjectNotExist`/2017."** — It does carry an errno detail. `arrow::fs::internal::PathNotFound` is `Status::IOError(...).WithDetail(StatusDetailFromErrno(ENOENT))`, and `ToSegcoreError` checks ENOENT before anything else, so that path already reaches 2017. No change needed, and none made.

Net: #595's Azure analysis holds and is what this PR implements; its GCS half shrinks to the init-path one-liner.

## Verification

- `make build` clean; clang-format-18 clean.
- New `cpp/test/filesystem/azure_error_classification_test.cpp`, which runs in CI without credentials: every transient status is retriable and maps to 2045; every permanent one is non-retriable, maps to its expected segcore code, and is asserted **not** to be 2045; unidentified statuses stay untagged and land on 2044; and the two conditions #595 named are stated as the end-to-end verdict a consumer sees.
- Full suite: **1071 ran / 915 passed / 154 skipped (cloud credentials) / 2 failed.** Both failures — `MetadataTest.TestFieldIDList` and `FileReaderTest.SchemaEvolutionWithInvalidFieldID` — are pre-existing: I reverted the change in place, rebuilt, and reproduced them identically on unmodified `origin/main`. They are in the `FieldIDList` area #598 is separately fixing.

## Not covered

- The Azure **name** channel: this classifies on HTTP status plus `ErrorCode` only where the status is ambiguous (409, 503). Azure error codes such as `OperationTimedOut` (which arrives as 500) are covered by their status, but a name-only signal with no distinguishing status is not.
- The Go/FFI consumer still discards the classification: milvus's `HandleLoonFFIResult` wraps every loon failure in one `ErrLoonTransient` sentinel and never reads `err_code`. The C++/segcore half is wired as of milvus-io/milvus#51530. So this makes Azure's errors *classifiable*, and they now reach segcore correctly, but the Go retry path still cannot tell them apart. Tracked in milvus-io/milvus#50903.
- No fault-injection e2e against a live Azure account.

Refs: #595, #574, #575, milvus-io/milvus#50903.

https://claude.ai/code/session_011HDgMok3nR8ZNugKWWQQK2

### DIFF
--- a/cpp/src/filesystem/azure/azurefs.cc
+++ b/cpp/src/filesystem/azure/azurefs.cc
@@ -115,11 +115,21 @@ using arrow::fs::internal::ValidateAbstractPathParts;
 // The codes are cloud-neutral despite the `AwsError*` prefix (see the naming
 // discussion in #574): they carry the meaning, not the vendor.
 std::optional<milvus_storage::ExtendStatusCode> ClassifyAzureError(
-    int http_status, std::string_view error_code) {
-  // No response at all: connection refused/reset, DNS failure, TLS handshake
-  // failure. The textbook retriable case.
-  if (http_status == 0) {
+    int http_status, std::string_view error_code, bool transport_failure) {
+  // The request never reached the service: connection refused/reset, DNS
+  // failure, TLS handshake failure. The textbook retriable case.
+  //
+  // This is keyed on the exception's dynamic type, NOT on `http_status == 0`.
+  // `RequestFailedException(std::string)` also leaves StatusCode at None, and
+  // PollUntilDone raises exactly that when a copy operation ends in a failed
+  // state -- treating that as a network blip would turn a definitively failed
+  // copy into an infinitely retriable one.
+  if (transport_failure) {
     return milvus_storage::ExtendStatusCode::StorageTransientNetwork;
+  }
+  // No response and not a transport failure: not positively identified.
+  if (http_status == 0) {
+    return std::nullopt;
   }
 
   switch (http_status) {
@@ -416,10 +426,14 @@ std::string BuildBaseUrl(const std::string& scheme, const std::string& authority
 template <typename... PrefixArgs>
 Status ExceptionToStatus(const Azure::Core::RequestFailedException& exception,
                          PrefixArgs&&... prefix_args) {
-  // A TransportException never received a response, so StatusCode is left at
-  // its None (0) default -- see internal::ClassifyAzureError.
+  // Discriminate on the dynamic type: TransportException means the request
+  // never reached the service. Its StatusCode is None, but so is that of a
+  // plain RequestFailedException(std::string), which PollUntilDone raises for
+  // a failed copy -- see internal::ClassifyAzureError.
+  const bool transport_failure =
+      dynamic_cast<const Azure::Core::Http::TransportException*>(&exception) != nullptr;
   auto code = internal::ClassifyAzureError(static_cast<int>(exception.StatusCode),
-                                           exception.ErrorCode);
+                                           exception.ErrorCode, transport_failure);
   if (!code.has_value()) {
     return Status::IOError(std::forward<PrefixArgs>(prefix_args)..., " Azure Error: [",
                            exception.ErrorCode, "] ", exception.what());
@@ -2751,6 +2765,11 @@ class AzureFileSystem::Impl {
                                    location.path, ": ", container_client.GetUrl());
         }
         std::vector<std::string> failed_blob_names;
+        // Keep the first per-blob failure as a classified Status. Discarding
+        // the exception and synthesizing a plain IOError below would drop the
+        // classification of every sub-response, so a 429/503 inside a batch
+        // delete would still be reported as a permanent storage error.
+        std::optional<Status> first_failure_status;
         for (auto& [blob_name_view, deferred_response] : deferred_responses) {
           bool success = true;
           try {
@@ -2758,20 +2777,36 @@ class AzureFileSystem::Impl {
             success = delete_result.Value.Deleted;
           } catch (const Storage::StorageException& exception) {
             success = false;
+            if (!first_failure_status.has_value()) {
+              first_failure_status = ExceptionToStatus(exception, "Failed to delete blob: ",
+                                                       blob_name_view, ": ");
+            }
           }
           if (!success) {
             failed_blob_names.emplace_back(blob_name_view);
           }
         }
         if (!failed_blob_names.empty()) {
+          std::stringstream message_stream;
           if (failed_blob_names.size() == 1) {
-            return Status::IOError("Failed to delete a blob: ", failed_blob_names[0],
-                                   ": " + container_client.GetUrl());
+            message_stream << "Failed to delete a blob: " << failed_blob_names[0] << ": "
+                           << container_client.GetUrl();
           } else {
-            return Status::IOError("Failed to delete blobs: [",
-                                   arrow::internal::JoinStrings(failed_blob_names, ", "),
-                                   "]: " + container_client.GetUrl());
+            message_stream << "Failed to delete blobs: ["
+                           << arrow::internal::JoinStrings(failed_blob_names, ", ")
+                           << "]: " << container_client.GetUrl();
           }
+          auto message = message_stream.str();
+          if (first_failure_status.has_value()) {
+            // Reuse the cause's code and detail so the classification survives;
+            // the message still names every blob that failed. A sub-response
+            // that reports Deleted=false without throwing carries nothing to
+            // preserve, so those fall back to the plain IOError below.
+            return Status(first_failure_status->code(),
+                          message + " " + first_failure_status->message(),
+                          first_failure_status->detail());
+          }
+          return Status::IOError(message);
         }
       }
       return Status::OK();
--- a/cpp/src/filesystem/azure/azurefs.cc
+++ b/cpp/src/filesystem/azure/azurefs.cc
@@ -108,6 +108,62 @@ using arrow::fs::internal::RemoveLeadingSlash;
 using arrow::fs::internal::RemoveTrailingSlash;
 using arrow::fs::internal::SplitAbstractPath;
 using arrow::fs::internal::ValidateAbstractPathParts;
+
+// Declared in azurefs_internal.h. Kept out of the anonymous namespace so it can
+// be unit-tested without an Azure account.
+//
+// The codes are cloud-neutral despite the `AwsError*` prefix (see the naming
+// discussion in #574): they carry the meaning, not the vendor.
+std::optional<milvus_storage::ExtendStatusCode> ClassifyAzureError(
+    int http_status, std::string_view error_code) {
+  // No response at all: connection refused/reset, DNS failure, TLS handshake
+  // failure. The textbook retriable case.
+  if (http_status == 0) {
+    return milvus_storage::ExtendStatusCode::StorageTransientNetwork;
+  }
+
+  switch (http_status) {
+    case 412:  // Precondition Failed
+      return milvus_storage::ExtendStatusCode::AwsErrorPreConditionFailed;
+    case 409:  // Conflict
+      // Only the "blob already exists" flavour is the precondition-style
+      // conflict this maps to. Other 409s (lease held, container being deleted)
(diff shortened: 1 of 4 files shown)


cpp/src/filesystem/azure/azurefs.cc
```diff
@@ -108,6 +108,62 @@ using arrow::fs::internal::RemoveLeadingSlash;
 using arrow::fs::internal::RemoveTrailingSlash;
 using arrow::fs::internal::SplitAbstractPath;
 using arrow::fs::internal::ValidateAbstractPathParts;
+
+// Declared in azurefs_internal.h. Kept out of the anonymous namespace so it can
+// be unit-tested without an Azure account.
+//
+// The codes are cloud-neutral despite the `AwsError*` prefix (see the naming
+// discussion in #574): they carry the meaning, not the vendor.
+std::optional<milvus_storage::ExtendStatusCode> ClassifyAzureError(
```
jiaqizho, line 117: One small gap in [this batch-delete path](https://github.com/milvus-io/milvus-storage/blob/65c8240e359cea19b7d290536884221353f8caf3/cpp/src/filesystem/azure/azurefs.cc#L2753-L2773): `GetResponse()` can throw a `StorageException`, but we swallow it and only keep the blob name. We then build a fresh plain `IOError`, so a per-blob 429/503 or another transient failure loses the `ExtendStatusCode` added by this PR and still reaches the caller as non-retryable 2044. Could we preserve one of the original statuses and run it through `ExceptionToStatus` when reporting the batch failure?

cpp/test/filesystem/azure_error_classification_test.cpp
```diff
@@ -0,0 +1,129 @@
...
+#include <gtest/gtest.h>
+
+#include <optional>
+#include <string_view>
+
+#include "common/EasyAssert.h"
+#include "milvus-storage/common/extend_status.h"
+#include "milvus-storage/filesystem/azure/azurefs_internal.h"
+
+namespace milvus_storage::fs::internal {
```
jiaqizho, line 30: use test_env.h to make sure current test only in azure cloud env.

no need running in local or s3 env.

xiaofan-luan:
Both findings are correct and both are now fixed in `873e5b0`. Thanks — the first one is the exact transient/permanent inversion this taxonomy exists to prevent, and I introduced it.

**1. `StatusCode == None` was the wrong discriminator.** Confirmed against the SDK header: `explicit RequestFailedException(std::string const& what)` leaves `StatusCode` at its `None` default, so it is indistinguishable from `TransportException` by status alone. And it is not hypothetical — the second catch in `Impl::CopyFile` documents itself as handling exactly that case:

```cpp
} catch (const Azure::Core::RequestFailedException& exception) {
  // A GetProperties call inside PollUntilDone returned a failed CopyStatus.
```

So a copy that definitively failed was being reported as `StorageTransientNetwork`/2045 and would be retried forever.

`ClassifyAzureError` now takes an explicit `transport_failure` discriminator that `ExceptionToStatus` derives from the dynamic type (`dynamic_cast<const Azure::Core::Http::TransportException*>`). Status 0 without it is no longer classified at all and falls back to the conservative non-retriable bucket. I kept the discriminator in the signature rather than doing the `dynamic_cast` inside the classifier so the case stays unit-testable without an SDK exception. New test `StatusZeroWithoutTransportIsNotRetriable` pins both directions.

**2. Batch delete did swallow the classification.** `DeleteDirContentsOnContainer` caught each blob's `StorageException`, kept only a bool, and synthesized a fresh plain `IOError` — so a 429/503 inside a batch delete was unaffected by this PR and still landed on 2044. It now keeps the first per-blob failure as a classified Status and returns it with the cause's code and detail, while the message still names every failed blob. Sub-responses that report `Deleted=false` without throwing carry nothing to preserve, so those keep the plain `IOError`.

Verified: `AzureErrorClassification` 5/5; full suite 1072 ran / 916 passed / 154 skipped (cloud credentials) / 2 failed — the same two pre-existing `FieldIDList` failures that reproduce on unmodified `origin/main`.

Still true and unchanged from the PR description: I have no Azure account, so the mapping is verified against the SDK type definitions and unit tests, never against a live service.

PR #605, opened 2026-07-30
fix: compute per-column chunk memory estimates on demand

jiaqizho:
The original per-column estimate path materialized and retained a full size vector for every row group, then built another field-name map for every chunk. That metadata scaled with both chunk count and column count, which is why BuildColumnMemorySizes became a major jemalloc allocation even though callers only need the values when they actually request an estimate.

This change keeps the lightweight total memory estimate in row-group and chunk metadata, but moves the column split behind FormatReader and calculates it on demand. Parquet uses the cached footer metadata, aggregates nested physical leaves into top-level Arrow fields, and normalizes them against the row-group total. Lance and Vortex retain one shared column-weight vector per fragment or file instead of copying it into every row group, while Iceberg reuses the Parquet values and rescales them for rows left after positional deletes.

ColumnGroupReader now requests all column estimates for a chunk in one call and maps the physical file schema back to the logical column-group order by field name. This keeps results stable when different files use schemas such as [a, b], [b, a], or [a, b, c], returns zero for logical columns missing from a file, excludes extra physical columns from the logical total, and rejects ambiguous duplicate field names.

The redistribution logic also keeps zero-weight columns at zero and assigns integer rounding residue to the last non-zero-weight column. This preserves the existing estimate behavior, avoids recalculating the same vector once per requested field, and reduces long-lived metadata from O(chunks × columns) allocations to row-group totals plus, where needed, a single O(columns) weight vector per file or fragment.

### DIFF
--- a/cpp/include/milvus-storage/reader.h
+++ b/cpp/include/milvus-storage/reader.h
@@ -108,10 +108,9 @@ class ChunkReader {
   /**
    * @brief Returns the estimated decoded-memory size of every chunk.
    *
-   * The result is indexed by chunk. When per-column metadata is available, each
-   * estimate is the sum of the logical column group's fields, independently of
-   * the active projection; fields present only in a physical file are excluded.
-   * Formats without per-column metadata retain their aggregate estimate.
+   * The result is indexed by chunk and contains the aggregate estimate recorded
+   * for that chunk. It is independent of the active projection and may include
+   * physical-file fields that are not present in the logical column group.
    * Estimates may be zero when a chunk has no live rows or its known
    * decoded-memory size is zero. Unavailable estimates are returned as an
    * error, typically arrow::Status::NotImplemented, rather than as zero.
@@ -119,20 +118,6 @@ class ChunkReader {
    * @return One estimated size in bytes per chunk, or an error status.
    */
   [[nodiscard]] virtual arrow::Result<std::vector<uint64_t>> get_chunk_estimated_size() = 0;
-
-  /**
-   * @brief Returns the estimated decoded-memory size of a top-level field in every chunk.
-   *
-   * The field is resolved against the logical column group, independently of
-   * the active projection. A field absent from a particular physical file has
-   * a known zero estimate for that chunk. Unavailable statistics are returned
-   * as an error rather than as zero.
-   *
-   * @param field_name Name of a unique top-level field in the column group.
-   * @return One estimated size in bytes per chunk, or an error status.
-   */
-  [[nodiscard]] virtual arrow::Result<std::vector<uint64_t>> get_chunk_column_estimated_size(
-      const std::string& field_name) = 0;
 
   /**
    * @brief Returns estimated decoded-memory sizes for every top-level field and chunk.
--- a/cpp/src/format/column_group_reader.cpp
+++ b/cpp/src/format/column_group_reader.cpp
@@ -16,10 +16,8 @@
 
 #include <algorithm>
 #include <future>
-#include <limits>
 #include <numeric>
 #include <sstream>
-#include <string_view>
 #include <unordered_map>
 #include <unordered_set>
 #include <utility>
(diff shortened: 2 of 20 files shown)


cpp/src/format/column_group_reader.cpp
```diff
@@ -716,54 +655,48 @@ arrow::Result<uint64_t> ColumnGroupReaderImpl<ReaderT>::get_chunk_estimated_size
...
-      return arrow::Status::Invalid("Duplicate column in column group: ", column_name);
-    }
-    ARROW_ASSIGN_OR_RAISE(auto column_size, get_chunk_column_estimated_size(chunk_index, static_cast<int>(col_idx)));
-    if (column_size > std::numeric_limits<uint64_t>::max() - total_size) {
-      return arrow::Status::Invalid("Chunk column memory sizes exceed the uint64_t range");
-    }
-    total_size += column_size;
-  }
-  return total_size;
+  return chunk_info.avg_memory_size;
```
liliu-z, line 658: get_chunk_estimated_size now returns chunk_info.avg_memory_size (the physical row group's total) instead of the sum of the logical column group's per-column estimates, so for column groups whose physical files retain columns dropped from the logical group the aggregate over-reports and no longer equals the sum of get_chunk_column_estimated_size(). Milvus consumes this aggregate via get_chunk_size() to size cache cells, so a retained large column (e.g. a dropped vector) inflates that cell's memory accounting and can drive premature eviction or load pressure. Either restore the logical-column total or confirm the physical-total semantics are acceptable for the cache-sizing consumer.

jiaqizho: expected. we should not allow pass the column group info into the format reader

cpp/include/milvus-storage/reader.h
```diff
@@ -120,20 +119,6 @@ class ChunkReader {
...
-   * as an error rather than as zero.
-   *
-   * @param field_name Name of a unique top-level field in the column group.
-   * @return One estimated size in bytes per chunk, or an error status.
-   */
-  [[nodiscard]] virtual arrow::Result<std::vector<uint64_t>> get_chunk_column_estimated_size(
-      const std::string& field_name) = 0;
-
   /**
    * @brief Returns estimated decoded-memory sizes for every top-level field and chunk.
```
liliu-z, line 123: This removes the pure virtual get_chunk_column_estimated_size(const std::string&) from the public ChunkReader interface, but Milvus master calls this exact string overload at ManifestGroupTranslator.cpp:188 (and references it in ManifestGroupTranslatorTest.cpp and test_sealed.cpp), so Milvus core will not compile against the new header. Removing a pure virtual also shifts the subsequent ChunkReader vtable slots, so swapping the shared library without a rebuild is silent ABI breakage. Retain the overload or land a coordinated Milvus-side migration before removing it.

jiaqizho: i know, expect

cpp/src/format/column_group_reader.cpp
buqian-zilliz, line 658: **[P2] Exclude physical-only fields from preload estimates**

Returning the raw row-group total here includes physical-only fields that are not read for this logical column group. `PackedRecordBatchReader::preload_column_group()` uses this estimate to stop scheduling chunks at `memory_usage_limit_`, so a schema-evolved file with a large retired field can make a small logical read load only one chunk per round and severely reduce scan throughput. Please preserve a logical-column aggregate, for example by caching a per-file logical weight or ratio, while still computing per-row-group column vectors on demand.

cpp/src/format/parquet/parquet_format_reader.cpp
```diff
@@ -503,6 +454,52 @@ folly::SemiFuture<arrow::Status> ParquetFormatReader::open_async() {
...
+    return arrow::Status::Invalid(fmt::format("Parquet reader metadata is not initialized. [path={}]", path_));
+  }
+
+  auto metadata = file_reader_->parquet_reader()->metadata();
+  if (!metadata || row_group_index >= metadata->num_row_groups()) {
+    return arrow::Status::Invalid(
+        fmt::format("Parquet row group metadata is not available. [path={}, row_group={}]", path_, row_group_index));
+  }
+
+  const auto leaf_column_offsets_by_field = get_leaf_column_offsets_by_field(*schema_);
```
buqian-zilliz, line 471: **[P2] Cache Parquet leaf offsets across row groups**

`get_chunk_column_estimated_size()` invokes this method once per chunk, but this line recursively rebuilds and heap-allocates the identical schema-to-leaf mapping for every row group before doing the unavoidable row-group metadata scan. Wide Parquet files with many row groups therefore pay O(row_groups × fields) extra traversal and one avoidable allocation per chunk in the estimate path. Cache or share the offsets once per reader/schema; this retains only O(fields) memory without restoring the removed O(row_groups × fields) metadata.

cpp/src/format/column_group_reader.cpp
weiliu1031, line 658: **[P2] Preserve the logical-column aggregate for cache sizing**

Returning `chunk_info.avg_memory_size` here includes physical-only fields that are absent from the logical column group. This value is consumed by `PackedRecordBatchReader::preload_column_group()` to enforce the reader memory limit, and by the Milvus `ManifestGroupTranslator` full-projection path to determine cache-cell grouping and loading overhead. With schema evolution, a retired large vector column can therefore make a small logical read appear much larger, reducing prefetch and creating premature cache or load pressure.

Please compute the aggregate on demand from `get_chunk_column_estimated_sizes()` with overflow checking, or retain a per-file logical ratio. That preserves the metadata-memory reduction while keeping the aggregate equal to the logical-column estimates.

cpp/src/format/parquet/parquet_format_reader.cpp
weiliu1031, line 471: **[P2] Cache the Parquet schema-to-leaf offsets**

`get_chunk_column_estimated_size()` invokes this path once per chunk, but `get_leaf_column_offsets_by_field(*schema_)` recursively rebuilds and heap-allocates the same mapping for every row group. Wide or nested Parquet files with many row groups therefore pay `O(row_groups × fields)` redundant traversal and allocation on the estimate path.

Please cache the offsets once per reader or shared metadata. This retains only `O(fields)` memory and does not restore the removed per-row-group column vectors.

cpp/src/format/column_group_reader.cpp
jiaqizho: expected

cpp/src/format/parquet/parquet_format_reader.cpp
jiaqizho: memory >> cpu

PR #608, opened 2026-07-30
enhance: classify iceberg planning errors

xiaofan-luan:
> **Depends on #597** — branched from it, since it introduces `bridge_error.rs`. The diff shown against `main` will include #597's commits until that lands. Review the top commit (`9cf5ddc`) only.

## Problem

iceberg was the last unclassified producer reachable from the Rust bridges. `iceberg_bridge.cpp` already decodes the marker, but nothing on the Rust side ever emitted one, so every planning failure crossed as an untagged string and landed on `StorageError`/2044 — a throttle on manifest reads looked exactly like a malformed table.

**Scope is narrower than it sounds.** iceberg's Rust does *planning only* (`iceberg_plan_files`); the data files it returns are read by the C++ parquet reader on the C++ `ArrowFileSystem`, which already classifies. So this covers metadata/manifest access and snapshot resolution, not the read path.

## Why the classification has to come from the source chain

`iceberg-storage-opendal` collapses every IO failure into a single kind:

```rust
// iceberg-storage-opendal-0.9.0/src/utils.rs
pub(crate) fn from_opendal_error(e: opendal::Error) -> iceberg::Error {
    iceberg::Error::new(iceberg::ErrorKind::Unexpected, "Failure in doing io operation")
        .with_source(e)
}
```

So `iceberg::ErrorKind` carries **no IO signal at all** — everything is `Unexpected`. What survives is the typed `opendal::Error` in the source chain. `classify_anyhow_error` walks the anyhow chain (anyhow keeps concrete types through `?`) and downcasts `iceberg::Error` then `opendal::Error`; first positive identification wins.

opendal is the friendlier of the two IO backends to classify against — it has a typed `ErrorKind` **and** its own `is_temporary()` bit, so unlike the `object_store` path nothing has to be recovered from prose.

| source | code | segcore |
|---|---|---|
| `opendal::NotFound`, `iceberg::TableNotFound`/`NamespaceNotFound` | 104 | 2017 `ObjectNotExist` |
| `opendal::PermissionDenied` | 105 | 2044 |
| `opendal::ConditionNotMatch`, `iceberg::PreconditionFailed` | 103 | 2044 |
| `opendal::RateLimited` | 109 | 2045 retriable |
| no specific kind but `is_temporary()` | 110 | 2045 retriable |
| `opendal::Unsupported`, `iceberg::FeatureUnsupported` | 1002 | `NotImplemented` |
| `iceberg::DataInvalid` | 1001 | 2024 `DataFormatBroken` |

## Deliberately left untagged

- `opendal::ConfigInvalid` — a caller/operator mistake, but there is no user-error code on this channel yet. Mislabelling it as a storage failure is worse than leaving it in the conservative bucket; revisit when the user/system axis lands (#603).
- iceberg's write-path conflicts (`TableAlreadyExists`, `NamespaceAlreadyExists`, `CatalogCommitConflicts`) — reusing the CAS-specific conflict code would dilute it, and `plan_files` is read-only anyway.

## Entry point

`iceberg_plan_files` becomes a thin wrapper that calls the unchanged anyhow-based body and maps once through `BridgeError::from`, so `?` stays ergonomic inside and classification happens at exactly one place.

## Verification

- `cargo test --lib bridge_error::tests` → **12 passed** (4 new).
- Full `make build` clean.
- The version pin is a **runtime** test, not a compile-time one: it reproduces `from_opendal_error`'s exact shape and asserts the chain walk recovers the code. An opendal version skew between this crate and `iceberg-storage-opendal` would make the downcast silently return `None` and degrade every iceberg IO failure to untagged — this catches that. A compile-time pin is impossible here because `with_source` accepts any error type.
- The C++ iceberg integration tests all skip without cloud credentials, as they do on main.

## Not covered

The iceberg *read* path needs nothing here (it is C++), and `ConfigInvalid` waits on the user/system axis. paimon remains the last unclassified stack (its own opendal, tracked with #602).

https://claude.ai/code/session_011HDgMok3nR8ZNugKWWQQK2


xiaofan-luan:
Closing as superseded by #603. The current #603 head already contains the Iceberg/OpenDAL typed source-chain classification and converts iceberg_plan_files failures through BridgeError at the boundary. Any additional mappings that were not carried into the current taxonomy can be revisited separately as a focused follow-up.

Issue #609, opened 2026-07-30
Stringified status throws destroy the error classification (vortex, packed reader, column_group)


xiaofan-luan:
Found while auditing throw sites for the error taxonomy (#603). These are one class of bug: a `Status` that already carries a classification is turned into a `std::string` and thrown as an untyped exception, so the classification is destroyed at the throw.

#603 defines the categories and #597 reduces the number of throws, but neither of them fixes these — the exception still escapes, it just escapes with a classification that nobody can read.

### Why it matters

`ExtendStatusDetail` survives every hop up the arrow stack via `ARROW_*_RAISE`. The moment it becomes `status.ToString()` inside a `std::runtime_error`, the only thing left is a message. The consumer's fallback is "unclassified ⇒ permanent", so **a retriable failure becomes a non-retriable one**. A throttled object store during a vortex cache miss was never retried, for exactly this reason.

`ToSegcoreError(const arrow::Status&)` already exists and returns a `milvus::SegcoreError`, which *is* an exception type carrying a `milvus::ErrorCode`. Where we must throw, that is what we should throw.

### Sites

| File:line | Current | Loses |
|---|---|---|
| `cpp/src/format/vortex/vortex_translater.cpp:180` | `throw std::runtime_error(status.ToString())` | the full `ExtendStatusDetail` |
| `cpp/src/packed/reader.cpp:88` | `throw std::runtime_error(status.ToString())` | the full `ExtendStatusDetail` |
| `cpp/src/packed/column_group.cpp:64` | `throw std::runtime_error(result.status().message())` | the detail **and** the arrow `StatusCode` — `.message()` drops even that |

The vortex one has a fix in flight (branch `fix/no-stringified-status-throw`) with a regression test that fails a throttled read through the translator and asserts the code survives.

Two notes on the other two:

- `packed/reader.cpp:88` is the deprecated throwing `PackedRecordBatchReader` constructor. If that constructor is going away anyway, deleting it is a better fix than reclassifying it — worth deciding before patching.
- `packed/column_group.cpp:64` is `arrow::Table::FromRecordBatches` failing, which is a schema mismatch across batches, i.e. our own invariant. That is `PackedUnexpected`, not a passthrough.

### Related, same audit, different shape

[...]

PR #610, opened 2026-07-30
fix: preserve classifications across C++ exception boundaries

xiaofan-luan:
## Problem

Two C++ exception boundaries received an `arrow::Status` with a usable storage classification and then discarded its type by throwing `std::runtime_error(status.ToString())`:

- `VortexTranslater::get_cells`, used by the cache cell-loader path.
- The deprecated direct-link `PackedRecordBatchReader` constructor.

A throttled Vortex read could therefore become an unclassified permanent failure, and Packed constructor callers could only inspect text.

## Fix

Use the existing `ToSegcoreError(status)` conversion at both boundaries. `milvus::SegcoreError` carries the consumer-facing error code, so the exception contract remains intact without parsing strings or adding another exception mechanism.

The deprecated Packed constructor is retained for source compatibility because current Milvus still calls it directly. New callers should use `PackedRecordBatchReader::Make()`; removal can happen after downstream migration.

## Verification

- Release build passed on current `main`.
- Focused Packed and Vortex tests: 41 passed, 1 existing metadata-null skip.
- Both regression tests assert the caught exception is `milvus::SegcoreError` with the expected code.
- clang-format dry run and `git diff --check` passed.

## Scope

`ColumnGroup::Table()` already preserves the original Status on current main. This PR resolves the remaining live Status-to-string exception boundaries tracked by #609 without changing the public constructor signature.

Refs #609.

### DIFF
--- a/cpp/src/format/vortex/vortex_translater.cpp
+++ b/cpp/src/format/vortex/vortex_translater.cpp
@@ -22,6 +22,8 @@
 #include <arrow/io/interfaces.h>
 #include <fmt/format.h>
 
+#include "milvus-storage/common/extend_status.h"
+
 namespace milvus_storage::vortex {
 
 namespace {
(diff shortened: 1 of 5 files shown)


cpp/src/format/vortex/vortex_translater.cpp
jiaqizho, line 161: Should this `throw` also be handled?

jiaqizho, line 148: ditto, and check others

PR #612, opened 2026-07-31
enhance: prevent deadlocks during S3 CRT client finalization

jiaqizho:
S3CrtClient destruction waits for native CRT shutdown callbacks. The previous operation guard shared ownership of the client, allowing the last reference to be released from a CRT callback and making that callback wait for its own completion. Global S3 shutdown could also race with client construction or destruction.

Make S3CrtClientHolder the sole owner of each CRT client and replace shared client locks with move-only, non-owning operation leases. Finalization now closes admission, waits for client construction and active operations to drain, destroys clients outside synchronization locks, and tracks live clients until every destructor has completed.

Keep asynchronous read state independent of the holder, release Rust reader ownership on its task executor, and route Arrow work through the consumer-provided Folly executor while rejecting inline-like executors. Add lifecycle, concurrency, native CRT read.

### DIFF
--- a/cpp/src/filesystem/s3/s3_crt_client.cpp
+++ b/cpp/src/filesystem/s3/s3_crt_client.cpp
@@ -17,10 +17,11 @@
 #ifdef WITH_CRT
 
 #include <algorithm>
+#include <condition_variable>
+#include <cstddef>
 #include <memory>
 #include <mutex>
 #include <optional>
-#include <shared_mutex>
 #include <utility>
 #include <vector>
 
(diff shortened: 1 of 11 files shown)


cpp/src/filesystem/s3/s3_crt_client.cpp
```diff
@@ -54,82 +55,311 @@ inline arrow::Status ErrorS3Finalized() { return arrow::Status::Invalid("S3 subs
...
-  std::shared_ptr<Aws::S3Crt::S3CrtClient> client;
+Aws::S3Crt::S3CrtClient* S3CrtClientLease::operator->() const { return client_; }
+
+void S3CrtClientLease::Release() {
+  // Clear the borrowed pointer before publishing that this operation is done.
+  // Moving operation_state_ out also makes repeated Release() calls harmless,
+  // which is required by move assignment and moved-from destruction.
+  client_ = nullptr;
+  auto operation_state = std::move(operation_state_);
+  if (!operation_state) {
```
liliu-z, line 163: Finalize() waits unconditionally for active_operations to reach zero and is reached straight from ~S3CrtClientHolder, so whichever thread drops the last holder reference while a lease is still outstanding blocks forever on its own stack. The header states this as a precondition ('The last holder reference must not be released from a native CRT callback') but nothing enforces it, and the FFI async read does capture an owning input_file shared_ptr in the Arrow callback that is destroyed on the CRT thread when MarkFinished runs inline. That path stays safe today only because ReaderHandle::drop calls loon_filesystem_reader_close first and ObjectCrtInputFile::Close() nulls holder_ off the CRT thread — an implicit cross-language dependency that is not documented and not asserted. Worth either enforcing (e.g. a debug check that Finalize is not entered from a CRT event-loop thread) or spelling out the Close()-before-destroy dependency next to the precondition comment.

cpp/src/format/parquet/folly_arrow_executor.cpp
```diff
@@ -72,7 +74,16 @@ class FollyArrowExecutor final : public arrow::internal::Executor {
...
 
-std::shared_ptr<arrow::internal::Executor> MakeFollyArrowExecutor(folly::Executor::KeepAlive<> executor, int capacity) {
+arrow::Result<std::shared_ptr<arrow::internal::Executor>> MakeFollyArrowExecutor(folly::Executor::KeepAlive<> executor,
+                                                                                 int capacity) {
+  // InlineLikeExecutor::add() runs work on the submitting thread. Arrow may
+  // submit Parquet decode work while an S3 future is completing on a CRT event
+  // loop, so accepting an inline executor could move that decode onto the CRT
+  // thread. Do not choose a fallback here; require the storage caller to bind a
+  // non-inline executor.
+  if (dynamic_cast<folly::InlineLikeExecutor*>(executor.get()) != nullptr) {
```
liliu-z, line 87: The dynamic_cast guard rejects only direct InlineLikeExecutor instances, which is weaker than the guarantee the header advertises ('InlineLikeExecutor is rejected so Arrow work cannot run on a CRT callback thread'). The in-tree deferred path already demonstrates the gap: MakeFollyArrowExecutor accepts the executor Folly binds for SemiFuture::get(), and the test then asserts the Arrow task ran synchronously on the waiting thread. A SerialExecutor over an inline parent would slip through the same way. Either soften the header wording to describe a best-effort check, or detect synchronous dispatch rather than a single RTTI type.

cpp/test/filesystem/s3_crt_build_test.cpp
```diff
@@ -44,17 +61,475 @@ arrow::Status EnsureS3InitializedForTest() {
...
+    last_lease = S3CrtClientLease{};
+    client_leases.clear();
+    finalized.wait();
+    FAIL() << "Finalize did not close new client operations";
+  }
+
+  std::promise<void> release_promise;
+  auto release_signal = release_promise.get_future().share();
+  std::vector<std::future<void>> release_futures;
+  release_futures.reserve(kConcurrentOperations);
```
liliu-z, line 389: `release_futures.emplace_back(std::async(std::launch::async, ...))` with `kConcurrentOperations = 100` (same pattern at line 456) throws `std::system_error` when a thread can't be created — realistic under a container `pids_limit`, `RLIMIT_NPROC`, or a sanitizer build. If it throws at iteration k, threads 0..k-1 park on `release_signal.wait()`, the `release_promise.set_value()` at the bottom is never reached, and unwinding destroys `release_futures`, whose destructors block joining tasks waiting on a promise nobody will fulfil, so CI hangs silently instead of failing. Cap the fan-out and set the promise from a scope guard.

jiaqizho: The failure sequence is valid, but it is confined to the test harness under thread-resource exhaustion. Production code uses neither std::async nor this barrier. kConcurrentOperations = 100 is intentional stress coverage for simultaneous lease release and inline future completion, and the test treats capacity for that fan-out as a prerequisite. We are keeping the stress setup unchanged in this PR.

cpp/src/filesystem/s3/s3_crt_client.cpp
jiaqizho: That ordering is enforced by the Rust ownership path rather than being an accidental timing dependency. read_async_via_ffi keeps Arc<ReaderHandle> in the async future frame and deliberately does not store it in AsyncReadCallbackState. The reader is dropped only after receiver.await resumes on the task executor. ReaderHandle::drop then calls loon_filesystem_reader_close before destroy, so ObjectCrtInputFile::holder_ is cleared off the CRT callback thread. If the task resumes before the C++ AddCallback capture is destroyed, Close clears the holder first; otherwise the wrapper still owns the input file and that capture is not the last owner. This ordering is documented beside drop(reader) and ReaderHandle. We therefore keep the holder precondition and do not add a CRT-thread runtime check.

cpp/src/format/parquet/folly_arrow_executor.cpp
jiaqizho: This guard is intentionally scoped to the Folly InlineLikeExecutor marker; dynamic_cast also rejects subclasses of that marker. folly::Executor has no reliable generic capability for detecting whether an arbitrary wrapper will dispatch inline, and probing add() would be racy. The SemiFuture::get() path is intentional: it is driven by the waiting consumer thread, not a CRT completion callback. For via(), supported callers must bind a non-inline executor; wrapping an inline parent would violate that caller contract. We are keeping the explicit guard as-is.

buqian-zilliz, line 87: **[P3] MakeFollyArrowExecutor does not reject an empty KeepAlive executor**

`executor` can be an empty `folly::Executor::KeepAlive<>` (for example a default-constructed token), in which case `executor.get()` returns `nullptr`. A `dynamic_cast` on a null pointer yields `nullptr`, so the new `InlineLikeExecutor` guard does not reject it and the function returns a `FollyArrowExecutor` whose `executor_` is null. The first `SpawnReal` call then dereferences `executor_` through `executor_->add(...)`, which is undefined behavior. Add an explicit emptiness check such as `if (!executor || dynamic_cast<folly::InlineLikeExecutor*>(executor.get()) != nullptr)` so an empty executor is rejected instead of being stored.

cpp/src/format/parquet/folly_arrow_executor.cpp
```diff
@@ -72,7 +74,16 @@ class FollyArrowExecutor final : public arrow::internal::Executor {
...
+    return arrow::Status::Invalid("Parquet async reads require a non-inline Folly executor");
```
buqian-zilliz, line 88: **[P3] Rejecting InlineLikeExecutor changes the behavior of Parquet async read consumers**

`MakeFollyArrowExecutor` now fails with `Parquet async reads require a non-inline Folly executor` whenever the bound executor is an `InlineLikeExecutor` (or any executor whose `add()` runs work on the submitting thread). This is a deliberate safety change, but it is a public-behavior change for `ParquetFormatReader::ReadAsync`/`take_async`: any consumer that drives those `SemiFuture` values with an inline executor will now fail only at future-consumption time. Please confirm every caller binds a non-inline executor (the C++ FFI layer appears to use thread-pool executors); otherwise synchronous/inline consumption paths regress.

PR #613, opened 2026-08-04
enhance: reuse AWS and Aliyun credential providers across Lance and Iceberg

jiaqizho:
Role-based cloud access was rebuilding its credential provider every time Lance opened a dataset or Iceberg created a FileIO. That gets expensive fast when callers create thousands of readers or writers, because the same filesystem identity repeatedly goes through provider setup, credential warm-up, STS calls, and, for Aliyun, OIDC token-file loading even though the role configuration has not changed. The goal is to reuse credential state at the process level without sharing mutable reader, writer, session, or FileIO state.

This change adds a bounded global LRU cache keyed by the existing filesystem cache key. C++ forwards that key for AWS and Aliyun role-based configurations, and the Rust bridge consumes it before passing storage options to upstream libraries. Lance and Iceberg keep separate typed caches for their provider and factory objects, serialize same-key initialization so only one caller creates an entry, leave failed initialization uncached, and evict the least recently used entry when the cache reaches its limit.

For AWS, AssumeRole now uses a shared single-flight credential provider. It warms credentials before publishing the provider, reuses them while they are fresh, and coordinates refreshes near expiration so concurrent callers do not duplicate STS work. Lance still creates a fresh session for each open, while Iceberg creates a fresh FileIO from the cached factory, keeping operation-specific state isolated.

For Aliyun, the cached Lance provider and Iceberg factory now own a refreshable OSS store. The OIDC-to-AssumeRole chain runs when the cache entry is first created and later only when credentials actually need refreshing, instead of running for every dataset open or metadata operation. The resolved configuration and refreshed STS credentials stay with that store, while the filesystem cache key keeps different roles, sessions, external IDs, endpoints, and tenants isolated.

This reduces repeated STS traffic, token-file reads, and provider allocations for high-volume reader and writer workloads. Cache-entry creation is logged without exposing credential material, and bridge-private options are removed before upstream storage code sees them so the caching behavior remains an internal implementation detail.

### DIFF
--- a/cpp/src/format/bridge/rust/src/aliyun_oss_provider.rs
+++ b/cpp/src/format/bridge/rust/src/aliyun_oss_provider.rs
@@ -88,7 +88,7 @@ use lance::session::Session;
 use lance::{Error as LanceError, Result as LanceResult};
 use lance_io::object_store::{
     DEFAULT_CLOUD_IO_PARALLELISM, ObjectStore, ObjectStoreParams, ObjectStoreProvider,
-    ObjectStoreRegistry, StorageOptions,
+    ObjectStoreRegistry,
 };
 
 use iceberg::io::{
@@ -107,6 +107,10 @@ const OSS_DEFAULT_DOWNLOAD_RETRIES: usize = 3;
 const ALIYUN_OSS_STORE_NAME: &str = "aliyun_oss";
 const OSS_CREDENTIAL_REFRESH_SECS_KEY: &str = "oss_credential_refresh_secs";
 const DEFAULT_OSS_CREDENTIAL_REFRESH_SECS: u64 = 50 * 60;
+const DEFAULT_ALIYUN_STS_DURATION_SECS: u64 = 60 * 60;
+const ALIYUN_STS_MIN_SESSION_DURATION_SECS: u64 = 15 * 60;
+const ALIYUN_STS_REFRESH_GRACE_SECS: u64 = 3 * 60;
+const ALIYUN_STS_MAX_SESSION_DURATION_SECS: u64 = 12 * 60 * 60;
 const REFRESH_LOCK_RETRY_MS: u64 = 10;
 
 // ============================================================================
@@ -298,13 +302,14 @@ pub(crate) fn build_oss_config_from_iceberg_opts(
 pub(crate) async fn apply_oidc_chain_if_requested(
     config_map: &mut HashMap<String, String>,
 ) -> Result<(), String> {
-    apply_oidc_chain_if_requested_with_expiration(config_map)
+    apply_oidc_chain_if_requested_with_expiration(config_map, DEFAULT_ALIYUN_STS_DURATION_SECS)
         .await
         .map(|_| ())
 }
 
 async fn apply_oidc_chain_if_requested_with_expiration(
     config_map: &mut HashMap<String, String>,
+    duration_seconds: u64,
 ) -> Result<Option<u64>, String> {
     // RAM mode owns this config_map; do not double-resolve.
     if std::env::var(ram::AUTH_MODE_ENV).as_deref() == Ok(ram::AUTH_MODE_RAM) {
@@ -351,6 +356,7 @@ async fn apply_oidc_chain_if_requested_with_expiration(
         &target_role_arn,
         &session_name,
         &external_id,
+        duration_seconds,
     )
     .await?;
     let expires_at_ms = outer.expires_at_ms;
@@ -432,18 +438,19 @@ async fn call_assume_role_with_oidc(
 /// callers, plain AK/SK callers, and any caller without the env var flipped
 /// fall straight through. Lance `role_arn` stores wrap this one-shot swap in
 /// [`RefreshableAliyunOssStore`] so long-lived scans rebuild the opendal store
-/// on the configured refresh interval. Iceberg intentionally calls this from
-/// `create_operator()` for each I/O operation instead of keeping a cache.
+/// shortly before the server-provided expiration. Iceberg role factories own
+/// the same refreshable store shape for their configured bucket.
 pub(crate) async fn apply_ram_mode_if_requested(
     config_map: &mut HashMap<String, String>,
 ) -> Result<(), String> {
-    apply_ram_mode_if_requested_with_expiration(config_map)
+    apply_ram_mode_if_requested_with_expiration(config_map, DEFAULT_ALIYUN_STS_DURATION_SECS)
         .await
         .map(|_| ())
 }
 
 async fn apply_ram_mode_if_requested_with_expiration(
     config_map: &mut HashMap<String, String>,
+    duration_seconds: u64,
 ) -> Result<Option<u64>, String> {
     if std::env::var(ram::AUTH_MODE_ENV).as_deref() != Ok(ram::AUTH_MODE_RAM) {
         return Ok(None);
@@ -456,7 +463,9 @@ async fn apply_ram_mode_if_requested_with_expiration(
         .cloned()
         .unwrap_or_else(ram::default_session_name);
     let external_id = config_map.get("external_id").cloned().unwrap_or_default();
-    let creds = ram::fetch_assume_role_creds(&role_arn, &session_name, &external_id).await?;
+    let creds =
+        ram::fetch_assume_role_creds(&role_arn, &session_name, &external_id, duration_seconds)
+            .await?;
     let expires_at_ms = creds.expires_at_ms;
     ram::apply_credentials(config_map, creds);
     Ok(expires_at_ms)
@@ -474,7 +483,7 @@ fn parse_aliyun_expiration_to_ms(s: &str) -> Result<u64, String> {
     u64::try_from(dt.timestamp_millis()).map_err(|_| format!("pre-epoch Expiration: {s}"))
 }
 
-fn parse_oss_credential_refresh_interval(
+fn parse_oss_credential_duration(
     storage_options: &HashMap<String, String>,
 ) -> Result<Duration, String> {
     let Some(raw) = storage_options.get(OSS_CREDENTIAL_REFRESH_SECS_KEY) else {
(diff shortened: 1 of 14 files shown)


cpp/src/format/bridge/rust/src/aliyun_oss_provider.rs
```diff
@@ -846,17 +864,61 @@ fn from_opendal_error(e: opendal::Error) -> IcebergError {
...
+            storage_options,
+        )
+        .map_err(|error| IcebergError::new(IcebergErrorKind::DataInvalid, error))?;
+        let refresh_interval = parse_oss_credential_refresh_interval(storage_options)
+            .map_err(|error| IcebergError::new(IcebergErrorKind::DataInvalid, error))?;
+        let store = Arc::new(RefreshableAliyunOssStore::new(
+            config_map,
+            refresh_interval,
+        ));
+        store.current_store().await.map_err(|error| {
```
liliu-z, line 926: The Aliyun STS AssumeRole request (`ram::call_assume_role`) sends no `DurationSeconds`, so sessions get Aliyun's fixed 3600s default. `validate_refresh_interval_before_expiration` rejects any refresh interval >= the credential lifetime, so a role config with `load_frequency >= ~3600` fails deterministically during warm-up here (`current_store()` runs the OIDC→AssumeRole chain and then validates). The default `load_frequency=900` and any value below ~3600 are unaffected. This PR newly routes Iceberg through this validating refreshable store (Iceberg previously resolved credentials per-op with no such check) and newly forwards `oss_credential_refresh_secs` for Iceberg in iceberg_common.cpp, so the failure now reaches Iceberg in addition to the pre-existing Lance path. Send an explicit `DurationSeconds` >= the refresh interval (bounded by the role's MaxSessionDuration, <=43200) in the AssumeRole request.

cpp/src/format/bridge/rust/src/aws_arn_provider.rs
```diff
@@ -0,0 +1,383 @@
...
+
+        let _refresh = self.refresh_lock.lock().await;
+        {
+            let cached = self.cached.read().await;
+            if let Some(credential) = self.current_credential(cached.as_ref()) {
+                return Ok(credential);
+            }
+        }
+
+        let refreshed = self
```
liliu-z, line 85: In `SingleFlightAwsCredentialProvider::get_credential`, when the cached credential is inside the 300s pre-expiry refresh window (still valid, not yet expired) and the STS refresh fails, the error is propagated instead of falling back to the still-valid cached credential, producing a spurious auth failure during an otherwise-recoverable transient STS error. On failure the cache is also left unchanged and the refresh lock released, so subsequent callers each re-attempt STS one at a time (no single-flight protection on the failure path). Serve the unexpired cached credential (with a warning) when a refresh attempt fails.

cpp/src/format/bridge/rust/src/cloud_provider_cache.rs
```diff
@@ -0,0 +1,151 @@
...
+    where
+        F: FnOnce() -> Fut,
+        Fut: Future<Output = Result<V, E>>,
+    {
+        let mut entries = self.entries.lock().await;
+        if let Some(value) = entries.get(key).cloned() {
+            return Ok(value);
+        }
+
+        let value = create().await?;
```
liliu-z, line 33: `GlobalLruCache::get` holds the single global mutex across `create().await`, so while one cache-miss runs its provider construction (an STS/OIDC network call, up to ~1s) every other `get()` — including cache hits for unrelated keys — blocks on the same mutex. This head-of-line blocking causes process-wide latency spikes whenever a new key is first created (cold start or post-eviction). Note the single-flight-per-key behavior is intended (a test asserts one init per key), so the fix must be a per-key in-flight future / keyed lock, not a plain double-checked insert, which would permit duplicate concurrent creates for the same key and defeat single-flight.

cpp/src/format/bridge/rust/src/aliyun_oss_provider.rs
jiaqizho: ack

cpp/src/format/bridge/rust/src/aliyun_oss_provider.rs
```diff
@@ -872,49 +931,56 @@ impl StorageFactory for AliyunOssStorageFactory {
...
-            })?
-            .to_string();
-
-        let mut config_map = build_oss_config_from_iceberg_opts(bucket, &url, &self.props)
-            .map_err(|e| IcebergError::new(IcebergErrorKind::DataInvalid, e))?;
-
-        apply_ram_mode_if_requested(&mut config_map)
-            .await
-            .map_err(|e| {
+        let config_map = if let Some(store) = &self.store {
```
liliu-z, line 976: On the cached Iceberg path, create_operator takes config_map from store.current_config(), whose bucket was fixed to the table's metadata-location bucket when the factory was built in from_uri. Every create_operator call then targets that one bucket regardless of the requested path's bucket, so a data file living in a different bucket than the table metadata is read/written against the metadata bucket (wrong object or NotFound). This is a regression: the pre-PR create_operator rebuilt the operator per call from url.host_str(), targeting each path's own bucket. Fix: reuse the cached credentials but bind the operator to the requested path's bucket.

cpp/src/format/bridge/rust/src/aliyun_oss_provider.rs
```diff
@@ -925,16 +991,7 @@ impl AliyunOssStorage {
...
-        let relative = path
-            .strip_prefix(&prefix)
-            .ok_or_else(|| {
-                IcebergError::new(
-                    IcebergErrorKind::DataInvalid,
-                    format!("Invalid oss url: {path}, should start with {prefix}"),
-                )
-            })?
-            .to_string();
+        let relative = url.path().trim_start_matches('/').to_string();
```
liliu-z: create_operator now derives the object key with url.path().trim_start_matches('/'). Url::path() returns the percent-encoded, dot-segment-canonicalized path, so keys containing spaces, non-ASCII, '+', '%', or literal '..' no longer round-trip — e.g. a space becomes %20 and is then re-encoded by opendal, addressing the wrong object. The pre-PR code used path.strip_prefix("oss://{bucket}/") on the raw string, preserving the key verbatim. Derive the key from the raw path after the bucket rather than through Url::path().

buqian-zilliz: **[P1] Preserve the bucket on cached Aliyun OSS paths**

The cached factory/provider is initialized with the first URI's bucket, but this line discards the scheme and authority from every later absolute path. For Iceberg, `oss://other-bucket/key` is therefore executed as `key` against the metadata bucket; the same bucket-bound store is reused by Lance's `new_store`. This can return unrelated same-named data or fail valid cross-bucket references, whereas the previous implementation derived and validated the bucket per path. Cache credentials independently of the bucket or select an operator/store using each requested host.

cpp/src/format/bridge/rust/src/cloud_provider_cache.rs
buqian-zilliz, line 33: **[P2] Do not await provider creation while holding the global cache lock**

The `entries` mutex remains locked while `create().await` performs AWS credential warm-up or Aliyun metadata/OIDC/STS requests. A slow cold miss for one key therefore blocks every lookup in that entire Lance or Iceberg cache, including hits for unrelated tenants, serializing startup and propagating one tenant's STS latency to all others. Hold the LRU lock only for lookup and insertion, and use per-key in-flight state or a `OnceCell` to preserve same-key single-flight initialization.

cpp/src/format/bridge/rust/src/aliyun_oss_provider.rs
jiaqizho: ack

jiaqizho: ack

cpp/src/format/bridge/rust/src/aliyun_oss_provider.rs
jiaqizho: ack

cpp/src/format/bridge/rust/src/aliyun_oss_provider.rs
```diff
@@ -872,49 +955,57 @@ impl StorageFactory for AliyunOssStorageFactory {
...
-            .map_err(|e| {
+        let requested_bucket = url.host_str().ok_or_else(|| {
+            IcebergError::new(
+                IcebergErrorKind::DataInvalid,
+                format!("Invalid oss url: {path}, missing bucket"),
+            )
+        })?;
+        let config_map = if let Some(store) = &self.store {
+            store
+                .validate_bucket(requested_bucket)
```
weiliu1031, line 978: **[P1] Preserve cross-bucket access for Iceberg paths**

This factory is bound to the metadata bucket, and `validate_bucket` now rejects every subsequent absolute path targeting another bucket. Iceberg tables may legitimately reference manifests, data files, or delete files stored in a different bucket; before this change, the object store was selected from each path's bucket.

Please cache credentials independently of the bucket, or cache bucket-specific stores, so valid cross-bucket table paths remain accessible.

cpp/src/format/bridge/rust/src/aws_arn_provider.rs
weiliu1031, line 85: **[P1] Keep using cached credentials until they actually expire**

Entering the early-refresh window currently makes credential refresh synchronous and propagates any STS/OIDC error immediately, even though the cached credentials are still valid. A transient identity-service failure can therefore interrupt storage requests up to five minutes before AWS credentials expire. The Aliyun implementation has the same issue with its three-minute refresh window.

Please distinguish "needs refresh" from "expired": if refresh fails while the cached credentials remain valid, return them and retry the refresh with backoff. Requests should fail only after the credentials actually expire.

cpp/src/format/lance/lance_common.cpp
```diff
@@ -102,7 +104,7 @@ StorageOptions ToStorageOptions(const ArrowFileSystemConfig& config) {
         options["allow_http"] = "true";
     }
   } else if (provider == kCloudProviderGCP) {
-    if (!config.gcp_target_service_account.empty()) {
+    if (config.use_iam && !config.gcp_target_service_account.empty()) {
```
weiliu1031, line 107: **[P2] Avoid silently changing the GCP authentication identity**

Previously, configuring a target service account was sufficient to enable impersonation. Requiring `use_iam` here makes existing configurations that omit this new flag silently fall back to the VM/default identity, potentially changing permissions and data access without reporting a configuration error. The documented configuration examples also still describe the previous behavior.

Please preserve backward-compatible impersonation behavior, or explicitly reject ambiguous configurations and update the documentation. This unrelated authentication behavior change would also be safer as a separate change.

cpp/src/format/bridge/rust/src/aws_arn_provider.rs
```diff
@@ -0,0 +1,383 @@
...
+            let provider = provider.clone();
+            async move { provider.get_credential().await.unwrap() }
+        }))
+        .await;
+
+        assert!(credentials.iter().all(|credential| credential.key_id == "key-0"));
+        assert_eq!(calls.load(Ordering::SeqCst), 1);
+    }
+
+    #[tokio::test]
```
weiliu1031, line 342: **[P3] Replace duplicate typed-cache tests with provider-level coverage**

These provider-specific tests only verify that a generic `TypedCache` returns the same `Arc` for the same dummy key. They never construct the actual credential provider or filesystem factory, so they would still pass if the AWS or Aliyun cache wiring were broken. The equivalent generic behavior is repeated four times.

Please keep a single generic cache-contract test and use provider-specific tests to exercise the real builder path and verify that repeated construction actually reuses the cached provider or factory.

PR #615, opened 2026-08-06
enhance: share Lance scan schedulers across datasets by I/O domain

jiaqizho:
Storage callers can create thousands of Lance readers, and each Dataset open currently ends up with its own ScanScheduler. Large-projection take and range reads can then fan out through thousands of independent queues against the same bucket and credentials, so object-store concurrency grows with reader count instead of staying bounded. The goal here is to give active datasets in the same ObjectStore and credential domain one shared backpressure boundary without changing Lance or mixing unrelated stores into one process-wide queue.

This adds `fs.lance_io_parallelism` as the storage-level control, with a default of 64 and a valid range of 0 through 256. A positive value is forwarded through the common Lance storage options and captured when a domain creates its scheduler, while 0 keeps Lance's existing per-Dataset behavior. `LANCE_IO_THREADS` still wins when it is set, and the new property stays out of the filesystem and provider cache key so changing queue capacity does not create another credential or ObjectStore identity.

The Rust bridge builds each I/O domain from Lance's actual store_prefix plus a process-local fingerprint of the ObjectStore construction inputs. The fingerprint ignores the parallelism value and the independent provider-cache key, but keeps provider, endpoint, region, credentials, roles, and broker identities separate. This lets datasets with different paths share safely when they really use the same store, while preventing a later reader from sending I/O through another tenant's ObjectStore.

An active-domain registry stores weak ScanScheduler references and performs lookup or creation atomically. Datasets and fragment readers keep the scheduler alive while it is usable, but the registry itself does not retain it forever. Once the last strong reference goes away, a later miss can clean up the expired entry and create a fresh scheduler. This also avoids the duplicate-queue problem that a strong LRU could introduce when an entry is evicted while older readers are still active.

When the requested capacity differs from the ObjectStore's current value, the bridge creates a scheduler-only ObjectStore view that reuses the existing inner client, credential provider, block size, ordering behavior, and download retry settings, changing only the I/O parallelism seen by ScanScheduler. Fragment take, read-all, range reads, and metadata estimation reuse that scheduler through FragReadConfig, while Dataset open and manifest I/O, direct Dataset scan or take, legacy files, and base_id stores stay outside this limit.

The existing AWS and Aliyun credential-provider LRU remains separate and unchanged. The common C++ conversion carries the bridge-private option for Lance opens, while the test-only write path removes it because writer I/O does not participate in shared scan scheduling. The result is a concurrency cap per active ObjectStore and credential domain rather than one global process-wide IOPS limit, so separate buckets or tenant identities still keep independent queues.


jiaqizho:
base on !613

cpp/src/format/bridge/rust/src/cloud_provider_cache.rs
```diff
@@ -0,0 +1,151 @@
...
+    where
+        F: FnOnce() -> Fut,
+        Fut: Future<Output = Result<V, E>>,
+    {
+        let mut entries = self.entries.lock().await;
+        if let Some(value) = entries.get(key).cloned() {
+            return Ok(value);
+        }
+
+        let value = create().await?;
```
buqian-zilliz, line 33: **[P1] Avoid global head-of-line blocking during credential warm-up**

`entries` is still locked when `create().await` runs, and every current creator warms an AWS or Aliyun provider with STS, OIDC, or IMDS I/O. A miss for one key therefore blocks cache hits and unrelated keys in the same Lance or Iceberg cache, so one slow or failing tenant can stall all opens and distinct tenants initialize serially. Keep only the LRU mutation under the global lock, and use per-key single-flight state while performing the network await after releasing that lock.

cpp/src/format/bridge/rust/src/aliyun_oss_provider.rs
```diff
@@ -872,49 +955,57 @@ impl StorageFactory for AliyunOssStorageFactory {
...
-            .map_err(|e| {
+        let requested_bucket = url.host_str().ok_or_else(|| {
+            IcebergError::new(
+                IcebergErrorKind::DataInvalid,
+                format!("Invalid oss url: {path}, missing bucket"),
+            )
+        })?;
+        let config_map = if let Some(store) = &self.store {
+            store
+                .validate_bucket(requested_bucket)
```
buqian-zilliz, line 978: **[P1] Preserve cross-bucket Iceberg paths for cached Aliyun credentials**

The cached store is created from the metadata URI's bucket, so this validation rejects every later absolute Iceberg path in another bucket. Iceberg allows `write.data.path` and other manifest or delete locations to be independent absolute locations, and the previous implementation rebuilt the OSS configuration from each requested path; those valid tables now fail during planning or positional-delete reads. Cache the refreshable credentials independently of bucket, or maintain a per-bucket operator/store map.

cpp/src/format/bridge/rust/src/lance_bridgeimpl.rs
```diff
@@ -40,13 +42,144 @@ use lance_table::utils::stream::ReadBatchFutStream;
...
+    storage_options: &HashMap<String, String>,
+    parallelism: usize,
+) -> Result<Arc<ScanScheduler>> {
+    let mut schedulers = LANCE_IO_SCHEDULERS
+        .lock()
+        .map_err(|_| LanceError::Internal {
+            message: "Lance I/O scheduler registry mutex poisoned".into(),
+            location: snafu::location!(),
+        })?;
+    if let Some(scheduler) = schedulers.get(&key).and_then(Weak::upgrade) {
```
buqian-zilliz, line 212: **[P2] Preserve per-dataset I/O accounting when reusing schedulers**

Reusing this scheduler also reuses the `ObjectStore` captured from the dataset that created it. Each dataset has its own I/O tracker, while `BlockingDataset::io_stats_incremental()` reads that dataset-local tracker, so fragment reads from later datasets are charged to the first dataset instead. Later datasets report zero, the first reports combined traffic, and after the creator is dropped those statistics become inaccessible. Preserve accounting per dataset or explicitly expose domain-level scheduler metrics.

cpp/include/milvus-storage/ffi_c.h
```diff
@@ -116,6 +116,7 @@ FFI_EXPORT extern const char* loon_properties_fs_tls_min_version;
 FFI_EXPORT extern const char* loon_properties_fs_background_writes;
 FFI_EXPORT extern const char* loon_properties_fs_use_crc32c_checksum;
 FFI_EXPORT extern const char* loon_properties_fs_s3_crt_async_read;
+FFI_EXPORT extern const char* loon_properties_fs_lance_io_parallelism;
```
buqian-zilliz, line 119: **[P2] Expose the new property through the Python constants API**

This adds the native symbol, but `python/milvus_storage/_ffi.py` does not declare it and `python/milvus_storage/properties.py` does not add `FS_LANCE_IO_PARALLELISM` to `_PROPERTY_KEY_MAPPING`. Python callers therefore get `AttributeError` from `PropertyKeys.FS_LANCE_IO_PARALLELISM`, unlike the other exported filesystem settings. Add both Python bindings with the C API change.

cpp/src/format/bridge/rust/src/lance_bridgeimpl.rs
```diff
@@ -57,10 +59,128 @@ use crate::cloud_provider_cache::{
...
```
buqian-zilliz, line 212: **[P2] Preserve fairness when sharing schedulers across datasets**

`ScanScheduler` orders one pending heap by the combined operation, reader, and file-position priority, and Lance's supplied-scheduler path is intended to coordinate readers within one operation. The bridge always opens fragments with `FragReadConfig::default()`, so every dataset uses reader priority 0. Reusing the scheduler here therefore mixes unrelated datasets in one priority namespace: a large low-position read can keep all I/O slots occupied while a higher-position point read from another dataset remains queued until the lower-valued backlog drains. Share only the domain-wide capacity while retaining per-dataset scheduling or fairness, rather than reusing the scan-local priority queue itself.

cpp/src/format/bridge/rust/src/lance_bridgeimpl.rs
buqian-zilliz: This issue still exists: the shared ScanScheduler still owns the first domain ObjectStore, or a scheduler-only clone wrapping its tracked inner store, while io_stats_incremental() reads each BlockingDataset's own tracker. Fragment reads from later datasets are therefore reported as zero on those datasets and charged to the first dataset; once that creator handle is gone, the accumulated tracker is no longer reachable through the public per-dataset API.

jiaqizho: ACK, changed  API

PR #616, opened 2026-08-06
feat: support Apache Paimon non-direct-file reads

yanbinyang:
Part of #593

## What changed

Add native Apache Paimon DataSplit reads as a follow-up to #602.

- Persist and validate versioned native DataSplit descriptors using paimon-rust 0.3.0.
- Stream projected Arrow batches without materializing a complete split.
- Stream each DataSplit read operation without materializing the complete split, while keeping the forward-only adapter inside the Paimon reader.
- Add the `data-split` scan mode and automatic fallback when direct-file reads cannot preserve Paimon semantics.
- Add Loon fixtures for append-only, deletion-vector, and merge-on-read tables.

## Routing

- `auto` keeps safe raw-convertible Parquet and Vortex splits on the direct-file path.
- `auto` falls back to DataSplit for merge-on-read, row ranges, partition columns, schema or data evolution, sidecars, and unsupported direct-file formats.
- `direct-file` fails explicitly when a split cannot be read safely.
- `data-split` forces native Paimon execution.

The direct-file behavior introduced by #602 remains unchanged.

Compared with the earlier combined #594 implementation, this PR is limited to DataSplit execution, uses the released native split codec, avoids whole-split materialization, coalesces normal eager and lazy loads onto a forward stream, and isolates take, range, and backward reads from the shared load cursor.

## Test plan

- Rust unit tests
- Paimon C++ integration and error-classification tests
- Loon append-only, deletion-vector, and merge-on-read scenarios
- Milvus external-collection Go unit tests
- Milvus standalone client/v3 E2E with auto, direct-file, DataSplit, deletion-vector, merge-on-read, and pinned snapshots

Signed-off-by: YangYanbin <warlock.yyb@alibaba-inc.com>

### DIFF
--- a/cpp/include/milvus-storage/format/paimon/paimon_format_reader.h
+++ b/cpp/include/milvus-storage/format/paimon/paimon_format_reader.h
@@ -67,6 +67,7 @@ class PaimonFormatReader final : public FormatReader {
         const std::string& predicate);
   };
 
+  ~PaimonFormatReader() override;
   [[nodiscard]] arrow::Status open() override;
   [[nodiscard]] arrow::Result<std::vector<RowGroupInfo>> get_row_group_infos() override;
   [[nodiscard]] arrow::Result<std::vector<uint64_t>> get_rg_column_memsz(int64_t row_group_index) const override;
@@ -106,6 +107,9 @@ class PaimonFormatReader final : public FormatReader {
   std::shared_ptr<BlockingPaimonDataSplitReader> split_reader_;
   std::vector<std::string> split_columns_;
   std::shared_ptr<arrow::Schema> output_schema_;
+  // Retain only the forward cursor used by the non-thread-safe get_chunk path.
+  // Batch and range operations keep independent streams.
+  std::unique_ptr<DataSplitStreamCursor> data_split_cursor_;
 };
 
 }  // namespace milvus_storage::paimon
--- a/cpp/include/milvus-storage/format/paimon/paimon_format_reader.h
+++ b/cpp/include/milvus-storage/format/paimon/paimon_format_reader.h
@@ -23,10 +23,14 @@
 
 namespace milvus_storage::paimon {
 
+class BlockingPaimonDataSplitReader;
+class DataSplitStreamCursor;
+
 class PaimonFormatReader final : public FormatReader {
   public:
   struct MetaTrait {
     struct Payload {
+      std::string read_path;
       std::string data_format;
       // Direct-file reads delegate to the underlying Parquet/Vortex reader, so
       // the payload keeps that reader's parsed metadata (the alternative
@@ -38,6 +42,12 @@ class PaimonFormatReader final : public FormatReader {
           direct_file_metadata;
       std::vector<RowGroupInfo> direct_physical_row_groups;
       std::shared_ptr<const std::vector<uint64_t>> sorted_deletions;
+      // Shared projection-agnostic bridge handle for the data-split read
+      // path. Creating it resolves the table schema and the pinned snapshot;
+      // sharing it across every reader built from this cached metadata keeps
+      // that cost per-descriptor instead of per-reader. Streams opened from
+      // it are per-reader and per-projection.
+      std::shared_ptr<BlockingPaimonDataSplitReader> split_reader_handle;
       uint64_t record_count = 0;
       uint64_t physical_row_count = 0;
     };
@@ -70,23 +80,31 @@ class PaimonFormatReader final : public FormatReader {
   [[nodiscard]] std::shared_ptr<arrow::Schema> get_schema() const override;
 
   private:
+  [[nodiscard]] bool is_data_split() const;
   PaimonFormatReader(MetaTrait::MetadataPtr metadata,
                      api::ColumnGroupFile file,
                      std::shared_ptr<arrow::Schema> read_schema,
                      std::vector<std::string> needed_columns,
                      std::shared_ptr<FormatReader> direct_file_reader,
+                     std::shared_ptr<BlockingPaimonDataSplitReader> split_reader,
+                     std::vector<std::string> split_columns,
                      std::shared_ptr<arrow::Schema> output_schema);
 
   [[nodiscard]] arrow::Result<std::shared_ptr<arrow::RecordBatch>> filter_direct_batch(
       const std::shared_ptr<arrow::RecordBatch>& batch, uint64_t physical_start) const;
   [[nodiscard]] arrow::Result<std::vector<int64_t>> logical_to_physical(
       const std::vector<int64_t>& logical_offsets) const;
+  [[nodiscard]] arrow::Result<std::unique_ptr<DataSplitStreamCursor>> make_data_split_cursor() const;
 
   MetaTrait::MetadataPtr metadata_;
   api::ColumnGroupFile file_;
   std::shared_ptr<arrow::Schema> read_schema_;
   std::vector<std::string> needed_columns_;
   std::shared_ptr<FormatReader> direct_file_reader_;
+  // Shared with (and owned by) the cached metadata payload; streams opened
+  // from it carry this reader's own projection in split_columns_.
+  std::shared_ptr<BlockingPaimonDataSplitReader> split_reader_;
+  std::vector<std::string> split_columns_;
   std::shared_ptr<arrow::Schema> output_schema_;
 };
 
--- a/cpp/include/milvus-storage/format/paimon/paimon_format_reader.h
+++ b/cpp/include/milvus-storage/format/paimon/paimon_format_reader.h
@@ -80,7 +80,6 @@ class PaimonFormatReader final : public FormatReader {
   [[nodiscard]] std::shared_ptr<arrow::Schema> get_schema() const override;
 
   private:
-  [[nodiscard]] bool is_data_split() const;
   PaimonFormatReader(MetaTrait::MetadataPtr metadata,
                      api::ColumnGroupFile file,
                      std::shared_ptr<arrow::Schema> read_schema,
@@ -89,6 +88,7 @@ class PaimonFormatReader final : public FormatReader {
                      std::shared_ptr<BlockingPaimonDataSplitReader> split_reader,
                      std::vector<std::string> split_columns,
                      std::shared_ptr<arrow::Schema> output_schema);
+  [[nodiscard]] bool is_data_split() const;
 
   [[nodiscard]] arrow::Result<std::shared_ptr<arrow::RecordBatch>> filter_direct_batch(
       const std::shared_ptr<arrow::RecordBatch>& batch, uint64_t physical_start) const;
--- a/cpp/src/format/bridge/rust/include/paimon_bridge.h
+++ b/cpp/src/format/bridge/rust/include/paimon_bridge.h
@@ -35,6 +35,10 @@ struct PaimonFileInfo {
   std::string metadata_json;
 };
 
+struct PaimonTestTableInfo {
+  std::vector<int64_t> snapshot_ids;
+};
+
 arrow::Status MakePaimonBridgeErrorStatus(std::string_view message);
 
 arrow::Result<std::vector<PaimonFileInfo>> PlanFiles(const std::string& table_location,
@@ -47,6 +51,14 @@ arrow::Result<std::vector<uint64_t>> ReadDeletionVector(const std::string& path,
                                                         uint64_t length,
                                                         int64_t expected_cardinality,
                                                         const StorageOptions& storage_options);
+
+arrow::Result<PaimonTestTableInfo> CreateTestTableInfo(const std::string& table_location,
+                                                       uint64_t num_rows,
+                                                       const std::string& mode,
+                                                       const std::vector<int64_t>& deleted_positions = {},
+                                                       const StorageOptions& storage_options = {},
+                                                       const std::string& file_format = "parquet",
+                                                       uint32_t dimension = 0);
 
 arrow::Result<int64_t> CreateTestTable(const std::string& table_location,
                                        uint64_t num_rows,
--- a/cpp/src/format/bridge/rust/include/paimon_bridge.h
+++ b/cpp/src/format/bridge/rust/include/paimon_bridge.h
@@ -25,6 +25,11 @@
 #include "rust/cxx.h"
 #include "rust-bridge/lib.h"
 
+namespace arrow {
+class RecordBatchReader;
+class Schema;
+}  // namespace arrow
+
 namespace milvus_storage::paimon {
 
 using StorageOptions = std::unordered_map<std::string, std::string>;
@@ -40,6 +45,11 @@ struct PaimonTestTableInfo {
 };
 
 arrow::Status MakePaimonBridgeErrorStatus(std::string_view message);
+
+namespace internal {
+std::shared_ptr<arrow::RecordBatchReader> WrapPaimonRecordBatchReader(std::shared_ptr<arrow::RecordBatchReader> inner,
+                                                                      std::shared_ptr<arrow::Schema> output_schema);
+}  // namespace internal
 
 arrow::Result<std::vector<PaimonFileInfo>> PlanFiles(const std::string& table_location,
                                                      int64_t snapshot_id,
--- a/cpp/src/format/bridge/rust/include/paimon_bridge.h
+++ b/cpp/src/format/bridge/rust/include/paimon_bridge.h
@@ -12,13 +12,18 @@
 
 #pragma once
 
+#include <arrow/c/abi.h>
+
 #include <cstdint>
+#include <memory>
 #include <string>
 #include <string_view>
 #include <unordered_map>
 #include <vector>
 
 #include <arrow/result.h>
+#include "rust/cxx.h"
+#include "rust-bridge/lib.h"
 
 namespace milvus_storage::paimon {
 
@@ -49,5 +54,30 @@ arrow::Result<int64_t> CreateTestTable(const std::string& table_location,
                                        const std::vector<int64_t>& deleted_positions = {},
                                        const std::string& file_format = "parquet",
                                        uint32_t dimension = 0);
+
+/// Projection-agnostic handle over one decoded DataSplit. `Open` performs the
+/// expensive work (descriptor decode, schema resolution, pinned-snapshot time
+/// travel) once; `OpenStream` starts a merge-read stream with a per-call
+/// projection, so one handle can be shared by every reader created from the
+/// same cached metadata.
+class BlockingPaimonDataSplitReader {
+  public:
+  static arrow::Result<std::unique_ptr<BlockingPaimonDataSplitReader>> Open(const std::string& metadata_json,
+                                                                            const std::string& expected_table_location,
+                                                                            const StorageOptions& storage_options);
+
+  explicit BlockingPaimonDataSplitReader(rust::Box<ffi::BlockingPaimonDataSplitReader> impl) : impl_(std::move(impl)) {}
+
+  BlockingPaimonDataSplitReader(BlockingPaimonDataSplitReader&&) noexcept = default;
+  BlockingPaimonDataSplitReader& operator=(BlockingPaimonDataSplitReader&&) noexcept = default;
+  BlockingPaimonDataSplitReader(const BlockingPaimonDataSplitReader&) = delete;
+  BlockingPaimonDataSplitReader& operator=(const BlockingPaimonDataSplitReader&) = delete;
+
+  arrow::Status ExportSchema(ArrowSchema* schema) const;
+  arrow::Result<ArrowArrayStream> OpenStream(const std::vector<std::string>& projected_columns) const;
+
+  private:
+  rust::Box<ffi::BlockingPaimonDataSplitReader> impl_;
+};
 
 }  // namespace milvus_storage::paimon
--- a/cpp/src/format/bridge/rust/include/paimon_bridge.h
+++ b/cpp/src/format/bridge/rust/include/paimon_bridge.h
@@ -74,7 +74,7 @@ arrow::Result<int64_t> CreateTestTable(const std::string& table_location,
 /// same cached metadata.
 class BlockingPaimonDataSplitReader {
   public:
-  static arrow::Result<std::unique_ptr<BlockingPaimonDataSplitReader>> Open(const std::string& metadata_json,
+  static arrow::Result<std::shared_ptr<BlockingPaimonDataSplitReader>> Open(const std::string& metadata_json,
                                                                             const std::string& expected_table_location,
                                                                             const StorageOptions& storage_options);
 
--- a/cpp/src/format/bridge/rust/src/lib.rs
+++ b/cpp/src/format/bridge/rust/src/lib.rs
@@ -243,7 +243,7 @@ pub mod paimon_ffi {
 
         unsafe fn open_stream(
             self: &BlockingPaimonDataSplitReader,
-            projected_columns: &CxxVector<CxxString>,
+            projected_columns: Vec<String>,
             out_stream_ptr: *mut u8,
         ) -> Result<()>;
     }
--- a/cpp/src/format/bridge/rust/src/lib.rs
+++ b/cpp/src/format/bridge/rust/src/lib.rs
@@ -22,6 +22,7 @@ mod iceberg_testutil;
 mod lance_bridgeimpl;
 mod lance_memory_estimator;
 mod paimon_bridgeimpl;
+mod paimon_split_serde;
 mod paimon_testutil;
 mod predicate_parser;
 mod rust_runtime;
@@ -201,7 +202,7 @@ pub mod lance_ffi {
 
 #[cxx::bridge(namespace = "milvus_storage::paimon::ffi")]
 pub mod paimon_ffi {
-    /// A physical file that Paimon determined is safe to read directly.
+    /// A single physical file or atomic DataSplit returned by Paimon planning.
     struct PaimonFileInfo {
         path: String,
         file_size: u64,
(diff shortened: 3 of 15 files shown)


cpp/src/reader.cpp
```diff
@@ -726,9 +726,34 @@ folly::SemiFuture<arrow::Result<std::vector<std::shared_ptr<arrow::RecordBatch>>
   };
   // Natural tasks preserve contiguous file ranges. Optional splitting changes
   // task granularity only; it does not select or cap an executor.
-  auto all_tasks = ChunkTask::Build(unique_chunk_indices, get_chunk_info);
-  SplitAsyncTasks(all_tasks, std::max<size_t>(parallelism, 1), ChunkTask::SplitTraits{get_chunk_info},
-                  GetAsyncTaskSplitStrategy(properties_));
+  auto natural_tasks = ChunkTask::Build(unique_chunk_indices, get_chunk_info);
```
jiaqizho: Currently, Paimon does not implement the native async interface. It uses the generic fallback, which executes the blocking read and wraps the result in an async task/future. Given that, I think the sequential-read constraint should remain within Paimon’s own get_chunk semantics instead of changing the common ChunkTask splitting logic.

Also, for the non-direct-read path, the current approach does not fully solve the duplicated-read problem. I would prefer to keep this PR focused on functional correctness first, and address the performance issue later with a more complete design.

cpp/include/milvus-storage/format/format_reader.h
```diff
@@ -107,6 +107,11 @@ class FormatReader {
   [[nodiscard]] virtual arrow::Result<std::vector<std::shared_ptr<arrow::RecordBatch>>> get_chunks(
       const std::vector<int>& rg_indices_in_file) = 0;
 
+  // Forward-only readers must receive requested chunks for one file through a
+  // shared ascending cursor. This avoids reopening and replaying their source
+  // stream for every logical range.
+  [[nodiscard]] virtual bool requires_sequential_chunk_reads() const { return false; }
```
jiaqizho: I don’t think `requires_sequential_chunk_reads` is a good API for adapting a forward-only streaming reader to a random-access chunk interface.

jiaqizho:
change commit title to `feat: support Apache Paimon non-direct-file reads` ?

cpp/src/format/paimon/paimon_format.cpp
```diff
@@ -35,18 +35,20 @@ arrow::Result<std::vector<api::ColumnGroupFile>> PaimonFormat::explore(const std
...
+    std::string read_path;
     try {
       auto metadata = nlohmann::json::parse(info.metadata_json);
       if (!metadata.is_object()) {
         return arrow::Status::Invalid("Paimon planning metadata must be a JSON object");
       }
-      if (metadata.value("read_path", std::string{}) != "direct-file") {
-        return arrow::Status::Invalid("Paimon planner returned a non-direct metadata descriptor");
-      }
+      read_path = metadata.value("read_path", std::string{});
```
jiaqizho: constexpr "read_path"/`direct-file`/`data-split`

and check others

cpp/src/format/bridge/rust/src/paimon_bridgeimpl.rs
```diff
@@ -790,9 +989,154 @@ pub fn paimon_read_deletion_vector(
...
 }
 
+type PaimonBatchStream = BoxStream<'static, paimon::Result<RecordBatch>>;
+
+struct PaimonStreamReader {
+    stream: PaimonBatchStream,
+    schema: SchemaRef,
+}
+
+fn classify_stream_error(error: paimon::Error) -> ArrowError {
```
yhmo, line 1014: cpp/src/format/bridge/rust/src/paimon_bridgeimpl.rs line:999
Medium ---- This stream-specific mapping turns every non-temporary `IoUnexpected` (including NotFound and PermissionDenied) into `InvalidArgumentError`, while temporary failures become a generic Arrow I/O error and lose the `StorageTransientThrottling` / `StorageTransientService` details used by the rest of the bridge. A failure after stream open therefore has different retry and refresh semantics from planning/open. Please preserve the same not-found, transient, and plain-I/O classification across the Arrow stream boundary.

cpp/src/format/paimon/paimon_format_reader.cpp
```diff
@@ -241,83 +461,105 @@ arrow::Result<PaimonFormatReader::MetaTrait::MetadataPtr> PaimonFormatReader::Me
...
+                          api::GetValue<uint64_t>(properties, PROPERTY_READER_LOGICAL_CHUNK_ROWS));
+    ARROW_ASSIGN_OR_RAISE(
+        auto reader, BlockingPaimonDataSplitReader::Open(metadata_json, ToStandardUri(file.path), storage_options));
+    ArrowSchema c_schema{};
+    ARROW_RETURN_NOT_OK(reader->ExportSchema(&c_schema));
+    ARROW_ASSIGN_OR_RAISE(metadata->file_schema, arrow::ImportSchema(&c_schema));
+    // Keep the handle: every reader created from this cached metadata shares
+    // it, so schema and snapshot resolution happen once per descriptor.
+    metadata->payload.split_reader_handle = std::move(reader);
+    ARROW_ASSIGN_OR_RAISE(metadata->row_group_infos, MakeLogicalRowGroups(parsed.record_count, logical_chunk_rows));
```
yhmo, line 526: cpp/src/format/paimon/paimon_format_reader.cpp line:557
Medium ---- Planning filters out zero-row DataSplits, but the persisted descriptor reader accepts `record_count == 0` and `MakeLogicalRowGroups` publishes no chunks. Because no stream is then consumed, the EOF validation never compares the declared count with the actual split, so corrupted metadata can silently hide a non-empty split. Please reject zero for data-split descriptors, or otherwise force explicit count validation before exposing the metadata.

cpp/tools/loon.cpp
```diff
@@ -147,24 +158,91 @@ static int DoDemoTable(int argc, char** argv) {
...
+              << " [--scenario <scenario>] [--rows N] [--dim N]"
+              << " [--deletes pos1,pos2,...] [--prop key=value ...]" << std::endl;
     std::cerr << std::endl;
-    std::cerr << "Types: iceberg" << std::endl;
+    std::cerr << "Types: iceberg, paimon" << std::endl;
+    std::cerr << "Paimon scenarios: append-only (default), merge-on-read, deletion-vector" << std::endl;
     std::cerr << std::endl;
-    std::cerr << "Creates a demo table with schema (id int64, name string,"
-              << " value float64)." << std::endl;
+    std::cerr << "Scalar schema: id int64, name string, value float64." << std::endl;
```
yhmo, line 182: cpp/tools/loon.cpp line:167
Low ---- This help now applies to both Iceberg and Paimon, but the Paimon scalar fixture declares `id` as `Int32` in both `fixture_schema` and `scalar_fixture_batch`, not `int64`. A user or downstream E2E that follows this text can construct the wrong external schema. Please distinguish the two table types here or document Paimon's actual `id` type.

cpp/src/reader.cpp
yanbinyang: OK. I’ll simplify this PR to focus on functional correctness, remove the cross-call cursor reuse and related test hooks, and leave broader DataSplit read optimization for a separate change.

yanbinyang:
> change commit title to `feat: support Apache Paimon non-direct-file reads` ?

Updated the first commit title to `feat: support Apache Paimon non-direct-file reads.`

cpp/include/milvus-storage/format/format_reader.h
yanbinyang: Removed in [4283be8d](https://github.com/milvus-io/milvus-storage/pull/616/commits/4283be8d72449a07e7428bca77da39f29f8f16a3). The common reader and ChunkTask scheduling are unchanged; Paimon now creates a private forward cursor for each read operation.

cpp/src/format/paimon/paimon_format.cpp
yanbinyang: Updated in [4283be8d](https://github.com/milvus-io/milvus-storage/pull/616/commits/4283be8d72449a07e7428bca77da39f29f8f16a3). read_path, direct-file, and data-split now use shared constants from paimon_common.h.

cpp/src/format/bridge/rust/src/paimon_bridgeimpl.rs
yanbinyang: Fixed in [c70fa849](https://github.com/milvus-io/milvus-storage/pull/616/commits/c70fa849522a13eb82b40ca8ca7432ab2c370de6). Stream errors now preserve the existing Paimon markers through Arrow FFI, so not-found, transient, invalid, not-implemented, and plain I/O errors keep the same handling as planning and open failures. The FFI and C++ status mappings are covered by tests.

cpp/tools/loon.cpp
yanbinyang: Fixed in [6f3a9a5f](https://github.com/milvus-io/milvus-storage/pull/616/commits/6f3a9a5f8cef73bf1570349ba4cf79361ac8f716). The Paimon scalar fixture now uses BIGINT / Arrow Int64, so the shared Loon schema text is accurate. The dependent C++ and Loon tests were updated

cpp/src/format/paimon/paimon_format_reader.cpp
yanbinyang: Fixed in [e8758f0e](https://github.com/milvus-io/milvus-storage/commit/e8758f0eab2596d57268a490ef9a9690c2217ea3). Milvus planning already omits zero-row splits, so a persisted zero-row descriptor violates that contract. This commit rejects it during metadata parsing, preventing an empty chunk list from bypassing stream count validation. The new test covers a non-empty split whose declared count is corrupted to zero; the existing fully-deleted-table test verifies legitimate empty results are omitted during planning.

cpp/include/milvus-storage/format/paimon/paimon_format_reader.h
```diff
@@ -70,23 +80,31 @@ class PaimonFormatReader final : public FormatReader {
   [[nodiscard]] std::shared_ptr<arrow::Schema> get_schema() const override;
 
   private:
+  [[nodiscard]] bool is_data_split() const;
```
jiaqizho: move `is_data_split` below private constructor.

cpp/src/format/bridge/rust/src/lib.rs
```diff
@@ -225,11 +228,33 @@ pub mod paimon_ffi {
...
+        ) -> Result<Box<BlockingPaimonDataSplitReader>>;
+
+        unsafe fn export_schema(
+            self: &BlockingPaimonDataSplitReader,
+            out_schema_ptr: *mut u8,
+        ) -> Result<()>;
+
+        unsafe fn open_stream(
+            self: &BlockingPaimonDataSplitReader,
+            projected_columns: &CxxVector<CxxString>,
```
jiaqizho: `Vec<String>` ? 

should convert vector<string> to rust type in cpp.

cpp/src/format/paimon/paimon_format_reader.cpp
```diff
@@ -34,7 +37,38 @@
 namespace milvus_storage::paimon {
 namespace {
```
jiaqizho, line 37: `namespace {` No need empty namespace. This is a bad habit of GPT.

cpp/src/format/bridge/rust/src/lib.rs
```diff
@@ -225,11 +228,33 @@ pub mod paimon_ffi {
...
+            storage_options_keys: Vec<String>,
+            storage_options_values: Vec<String>,
+        ) -> Result<Box<BlockingPaimonDataSplitReader>>;
+
+        unsafe fn export_schema(
+            self: &BlockingPaimonDataSplitReader,
+            out_schema_ptr: *mut u8,
+        ) -> Result<()>;
+
+        unsafe fn open_stream(
```
jiaqizho, line 244: can `open_stream` pass with footer or some meta? 

Then different projection won't read footer in multi-times.

jiaqizho: notice that: The external table will projects each column individually.

cpp/src/format/bridge/rust/include/paimon_bridge.h
```diff
@@ -43,11 +52,44 @@ arrow::Result<std::vector<uint64_t>> ReadDeletionVector(const std::string& path,
...
                                        uint32_t dimension = 0);
 
+/// Projection-agnostic handle over one decoded DataSplit. `Open` performs the
+/// expensive work (descriptor decode, schema resolution, pinned-snapshot time
+/// travel) once; `OpenStream` starts a merge-read stream with a per-call
+/// projection, so one handle can be shared by every reader created from the
+/// same cached metadata.
+class BlockingPaimonDataSplitReader {
+  public:
+  static arrow::Result<std::unique_ptr<BlockingPaimonDataSplitReader>> Open(const std::string& metadata_json,
```
jiaqizho: direct return `std::shared_ptr`? Then u don't need warp it as shared

cpp/src/format/paimon/paimon_format_reader.cpp
```diff
@@ -421,6 +697,23 @@ arrow::Result<std::vector<std::shared_ptr<arrow::RecordBatch>>> PaimonFormatRead
...
+      if (index <= previous || index < 0 || static_cast<size_t>(index) >= metadata_->row_group_infos.size()) {
+        return arrow::Status::Invalid("Paimon get_chunks requires sorted unique row group indices");
+      }
+      previous = index;
+    }
+    ARROW_ASSIGN_OR_RAISE(auto cursor, make_data_split_cursor());
+    for (auto index : indices) {
+      const auto& group = metadata_->row_group_infos[index];
+      ARROW_ASSIGN_OR_RAISE(auto batches, cursor->ReadRange(group.start_offset, group.end_offset));
+      ARROW_ASSIGN_OR_RAISE(auto batch, CombineBatches(batches, output_schema_));
```
jiaqizho, line 697: It will make copy. 

In fact, the "logical chunk rows" value is merely an indicator rather than a hard requirement. You can observe the approach in vortex: it actually returns a `native split` to inform the caller how to partition the data so that no resulting `RecordBatch` exceeds the limit, thereby avoiding data copying.

jiaqizho: We can allow chunks to be smaller, rather than copying them.

cpp/src/format/paimon/paimon_format_reader.cpp
```diff
@@ -469,6 +762,11 @@ arrow::Result<std::shared_ptr<arrow::Table>> PaimonFormatReader::take(const std:
   if (indices.empty()) {
     return arrow::Table::MakeEmpty(output_schema_);
   }
+  if (is_data_split()) {
+    ARROW_ASSIGN_OR_RAISE(auto cursor, make_data_split_cursor());
+    ARROW_ASSIGN_OR_RAISE(auto selected_batches, cursor->TakeRows(indices));
+    return arrow::Table::FromRecordBatches(output_schema_, selected_batches);
```
liliu-z, line 753: On the data-split path `output_schema_` (which is simply `read_schema` whenever the caller supplies one, see `ProjectSchema` at :164) is bound to batches that carry Paimon's own stream schema, and nothing reconciles the two. `CombineBatches` returns the source batches untouched for any non-empty input, so `get_chunk`/`get_chunks` emit stream-schema batches and `DataSplitRangeReader::schema()` (:427) advertises a schema its `ReadNext` (:436) does not produce, while `take()` passes `output_schema_` to `arrow::Table::FromRecordBatches`, which rejects any batch whose schema differs in type or nullability. A caller whose `read_schema` differs from Paimon's Arrow schema only in nullability therefore gets `Invalid: Schema at index 0 was different` from `take()` and from `arrow::Table::FromRecordBatchReader` on the range reader, even though the Paimon direct-file path and the other format readers accept the same input. Pick one rule for this path — project the stream onto `output_schema_`, or derive `output_schema_` from the stream — and apply it to all three entry points. (Note: the original claim that `take()`/`read_with_range()` honor `read_schema` is not accurate; `read_with_range` uses the same `CombineBatches`, and `take()` only validates.)

cpp/test/format/paimon/paimon_integration_test.cpp
```diff
@@ -525,7 +570,151 @@ TEST_F(PaimonIntegrationTest, ExplicitDirectFileRejectsMergeOnRead) {
...
+  constexpr uint64_t kRows = 10000;
+  ASSERT_STATUS_OK(paimon::CreateTestTable(table_dir_, kRows, "mor").status());
+  ASSERT_AND_ASSIGN(auto files, Explore("auto"));
+  ASSERT_EQ(files.size(), 1);
+  ASSERT_EQ(api::SetValue(properties_, PROPERTY_READER_LOGICAL_CHUNK_ROWS, "8192"), std::nullopt);
+
+  auto column_groups = std::make_shared<api::ColumnGroups>();
+  column_groups->push_back(std::make_shared<api::ColumnGroup>(
+      api::ColumnGroup{.columns = {"id"}, .format = LOON_FORMAT_PAIMON_TABLE, .files = {files.front()}}));
+  auto schema = arrow::schema({arrow::field("id", arrow::int64())});
```
liliu-z, line 810: No new data-split test exercises `take()` or `read_with_range()` with a non-null `read_schema`: `DataSplitTakeCompactsSparseBatches` (:668) and `DataSplitSupportsRangeAndCloneReads` (:607) both build the reader with `FormatReader::create(nullptr, ...)`, so `output_schema_` is derived from Paimon's own stream schema and the schema binding cannot fail. The two tests that do pass a schema use `arrow::schema({arrow::field("id", arrow::int64())})`, which matches the fixture's nullable BIGINT exactly — and one of them only asserts `total_rows()` while the other goes through the chunk path. Add take/range cases with a `read_schema` that differs from Paimon's schema in nullability or type so the mismatch behaviour is pinned down rather than untested.

cpp/src/format/paimon/paimon_format_reader.cpp
```diff
@@ -421,6 +697,23 @@ arrow::Result<std::vector<std::shared_ptr<arrow::RecordBatch>>> PaimonFormatRead
...
   }
+  if (is_data_split()) {
+    int previous = -1;
+    for (auto index : indices) {
+      if (index <= previous || index < 0 || static_cast<size_t>(index) >= metadata_->row_group_infos.size()) {
+        return arrow::Status::Invalid("Paimon get_chunks requires sorted unique row group indices");
+      }
+      previous = index;
+    }
+    ARROW_ASSIGN_OR_RAISE(auto cursor, make_data_split_cursor());
```
liliu-z, line 693: Every public read on the data-split path opens a brand-new Paimon merge stream and forward-skips to its start, and the reader keeps no cursor between calls. `get_chunk(i)` delegates to `get_chunks({i})`, which builds a fresh cursor, so iterating a split chunk-by-chunk decodes O(N^2) rows: for a 10M-row split at the default 8192-row chunk size that is ~1220 streams and ~6 billion rows decoded instead of 10 million. This is the primary Milvus access pattern — `loon_get_chunk` (cpp/src/ffi/reader_c.cpp:111) is per-chunk, and `api::Reader` itself falls back to single-chunk reads whenever a predicate is set (cpp/src/reader.cpp:449-457). Hoisting the cursor to a reader member and reusing it while `start >= position_` (falling back to a fresh stream only for backward seeks) would keep the forward-only design and make sequential iteration linear; `get_chunks` already proves one cursor can serve many chunks.

cpp/src/format/bridge/rust/src/lib.rs
yanbinyang: open_stream consumes the projection synchronously and does not retain it, so I used a borrowed CxxVector<CxxString> to avoid copying the C++ strings into an owned Rust vector. Would you still prefer converting it on the C++ side to keep the Rust interface as Vec<String>? I can change it if so.

cpp/src/format/paimon/paimon_format_reader.cpp
yanbinyang: The anonymous namespace contains helpers used only in this translation unit and keeps their symbols file-local, consistent with the Iceberg and Vortex readers in this repository. Would you prefer using static helpers here instead? I can adjust it if that is the intended style.

cpp/src/format/bridge/rust/src/lib.rs
jiaqizho: for `projection field` copy, I do think that is fine.

cpp/src/format/paimon/paimon_format_reader.cpp
jiaqizho: ok, just let me clean up empty namespaces across the entire `storage` project in the future.

cpp/include/milvus-storage/format/paimon/paimon_format_reader.h
yanbinyang: Moved below the private constructor in [f2872880](https://github.com/milvus-io/milvus-storage/pull/616/commits/f28728802cd5d980d3a9bd16bd2cd7fc859f958e).

cpp/src/format/bridge/rust/include/paimon_bridge.h
yanbinyang: Updated in [400e0557](https://github.com/milvus-io/milvus-storage/pull/616/commits/400e05572e455d7967c9ce1e158fa863f1f81310).

cpp/src/format/paimon/paimon_format_reader.cpp
yanbinyang: Fixed in [6343a578](https://github.com/milvus-io/milvus-storage/pull/616/commits/6343a578b2bbed471cfde1f829d15e45863ebaf2). DataSplit batches are validated and rebound to output_schema_ for chunk, take, and range reads. Incompatible fields and nulls for non-nullable fields return Invalid.

cpp/test/format/paimon/paimon_integration_test.cpp
yanbinyang: Covered in [6343a578](https://github.com/milvus-io/milvus-storage/pull/616/commits/6343a578b2bbed471cfde1f829d15e45863ebaf2). The test uses a non-null read_schema with different nullability for take and range reads, and verifies that an incompatible type is rejected.

cpp/src/format/paimon/paimon_format_reader.cpp
yanbinyang: Fixed in 81e4d6fd. Sequential get_chunk calls reuse a reader-local forward cursor and reopen only for backward reads or after an error. Batch and range reads keep independent streams. The test covers both forward reuse and backward reopening.

cpp/src/format/bridge/rust/src/lib.rs
yanbinyang: Understood. I’ll switch the projection argument to an owned Vec<String> and convert it on the C++ side.

cpp/src/format/bridge/rust/src/lib.rs
yanbinyang: The current paimon-rust 0.3 API does not expose reusable file metadata across projected readers. I see two options:

1. Keep this PR on the released 0.3 API and accept repeated footer reads for now, then optimize it separately.

2. Add a scan-scoped read context with internal footer metadata caching to paimon-rust, and share it from BlockingPaimonDataSplitReader across projected streams. This is closer to paimon-cpp’s design, but requires upstream API work and likely waiting for paimon-rust 0.4.0, no earlier than the end of this month.

Would you prefer keeping this PR small, or waiting for the upstream API?

cpp/src/format/paimon/paimon_format_reader.cpp
yanbinyang: I checked Paimon Java and paimon-cpp. Neither exposes merged DataSplit output boundaries during planning. Both use a pull-based reader (readBatch / NextBatch) and allow each returned batch to have its actual size.
For DataSplit/MOR, output boundaries are determined during merge and deletion processing, so exposing physical file boundaries from paimon-rust would not describe the logical output. Removing CombineBatches cleanly would require a variable-sized streaming contract in milvus-storage, which is broader than this PR.

cpp/src/format/paimon/paimon_format_reader.cpp
yanbinyang: A clarification to my earlier reply: the common reader and ChunkTask changes remain removed. Removed in 4283be8d. After the later O(N²) issue was identified, 81e4d6fd restores only a Paimon reader-local forward cursor for sequential get_chunk calls. It reopens for backward reads or errors, while batch and range reads use independent streams, so no common scheduling capability was reintroduced.

cpp/src/format/bridge/rust/src/lib.rs
yanbinyang: Updated in [19d30d4](https://github.com/milvus-io/milvus-storage/pull/616/commits/19d30d42b74e3c7357a2a5469ed35398a4e6b597).

cpp/src/format/bridge/rust/src/lib.rs
jiaqizho: okay, So let's just leave a `FIXME(name)` then.

cpp/src/format/paimon/paimon_format_reader.cpp
jiaqizho: ACK

cpp/src/format/bridge/rust/src/lib.rs
yanbinyang: Updated in [19d30d4](https://github.com/milvus-io/milvus-storage/pull/616/commits/19d30d42b74e3c7357a2a5469ed35398a4e6b597).

yanbinyang:
Updated in 31c3bf17. read_schema is now treated as the parent schema and projected by needed_columns in the requested order. Added coverage for reordered subset projections across Parquet/Vortex direct-file and data-split paths, including both get_chunk and take.

PR #619, opened 2026-08-11
enhance: reuse Azure and GCP cloud credential providers across Lance and Iceberg

jiaqizho:
Azure broker-issued SAS credentials and GCP service account impersonation were still creating new credential state whenever Lance opened or wrote a dataset and whenever Iceberg built storage. That gets expensive quickly when callers create thousands of readers or writers, because the same filesystem identity repeatedly goes through provider setup, credential warm-up, and broker or IAM requests even though the underlying cloud configuration has not changed.

This change forwards the existing filesystem cache key for Azure broker and GCP impersonation configurations, then uses it to reuse typed credential providers and storage factories in the Rust bridge. New providers resolve an initial credential before they are published, concurrent initialization for the same key is coordinated, failed initialization is not retained, and the bounded LRU keeps process-level reuse from growing without limit. Lance and Iceberg still use separate caches, while Azure Iceberg entries also include the resolved storage scheme so incompatible endpoints cannot share a factory.

For Azure, Lance now reuses a warmed SAS provider and overlays refreshed tokens on top of the static storage options. Iceberg uses the same refreshable credential flow through a lightweight storage wrapper that injects the current SAS token before delegating each operation to OpenDAL. Refreshes use a fast cached path and coordinate concurrent callers to avoid broker request spikes, while a previously cached SAS token remains available when a refresh temporarily fails and requests fail closed when no credential has ever been resolved.

For GCP, the shared impersonation module now owns configuration parsing, token lifetime validation, provider construction, and refresh behavior for both storage stacks. Lance installs the cached provider into a fresh session for each dataset operation, while Iceberg obtains the current impersonated bearer token and creates fresh GCS storage for each operation. Ambient configuration and VM metadata credentials are disabled on this path so the requested target service account remains the authoritative identity.

The Lance read and write paths now share one object store setup flow, which keeps cloud routing, cache lookup, and credential initialization consistent instead of maintaining two copies of the same logic. Bridge-private routing and credential options are consumed before reaching upstream libraries, and runtime provider state is excluded from serialization so cached credentials and unrelated storage properties are not exposed.

Overall, this reduces repeated provider allocations and unnecessary Azure broker or GCP IAM traffic for high-volume reader and writer workloads, while keeping sessions, storage instances, and operation-specific state isolated.

### DIFF
--- a/cpp/src/format/bridge/rust/src/azure_sas_provider.rs
+++ b/cpp/src/format/bridge/rust/src/azure_sas_provider.rs
@@ -22,6 +22,12 @@ use std::time::Duration;
 use anyhow::{Result as AnyResult, anyhow, bail};
 use async_trait::async_trait;
 use chrono::{DateTime, Utc};
+use iceberg::io::{
+    ADLS_SAS_TOKEN, FileMetadata, FileRead, FileWrite, InputFile, OutputFile,
+    Storage as IcebergStorage, StorageConfig, StorageFactory,
+};
+use iceberg::{Error as IcebergError, ErrorKind as IcebergErrorKind, Result as IcebergResult};
+use iceberg_storage_opendal::OpenDalStorageFactory;
 use lance_core::error::{Error as LanceError, Result as LanceResult};
 use lance_io::object_store::StorageOptionsProvider;
 use serde::{Deserialize, Serialize};
@@ -314,18 +320,15 @@ impl AzureSasStorageOptionsProvider {
             "Azure SAS credential broker failure: {error}"
         ))))
     }
-}
 
-#[async_trait]
-impl StorageOptionsProvider for AzureSasStorageOptionsProvider {
-    async fn fetch_storage_options(&self) -> LanceResult<Option<HashMap<String, String>>> {
+    pub(crate) async fn current_credential(&self) -> AnyResult<AzureSasCredential> {
         let mut now = (self.clock)();
         {
             let cached = self.cached.read().await;
             if let Some(credential) = cached.as_ref()
                 && Self::is_fresh(credential, now)
             {
-                return Ok(Some(Self::to_options(credential)));
+                return Ok(credential.clone());
             }
         }
 
@@ -334,14 +337,13 @@ impl StorageOptionsProvider for AzureSasStorageOptionsProvider {
         if let Some(credential) = cached.as_ref()
             && Self::is_fresh(credential, now)
         {
-            return Ok(Some(Self::to_options(credential)));
+            return Ok(credential.clone());
         }
 
         match self.fetcher.fetch(now).await {
             Ok(credential) => {
-                let options = Self::to_options(&credential);
-                *cached = Some(credential);
-                Ok(Some(options))
+                *cached = Some(credential.clone());
+                Ok(credential)
             }
             Err(error) => {
                 let has_cached_sas = cached.is_some();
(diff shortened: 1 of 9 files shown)


cpp/src/format/bridge/rust/src/azure_sas_provider.rs
```diff
@@ -351,32 +353,190 @@ impl StorageOptionsProvider for AzureSasStorageOptionsProvider {
...
-                    has_cached_sas,
-                    cached_expired
+                    error, has_cached_sas, cached_expired
                 );
-                if let Some(credential) = cached.as_ref() {
-                    Ok(Some(Self::to_options(credential)))
-                } else {
-                    Err(Self::lance_error(&error))
-                }
+                cached.as_ref().cloned().ok_or(error)
```
buqian-zilliz: **[P1] Back off failed SAS refreshes before reusing the cache**

This fallback leaves the cached entry inside the refresh window after a failed fetch and returns it even when it is already expired. Because the provider is now shared and its write lock covers the broker request, every queued Lance or Iceberg caller can acquire the lock and repeat the same broker timeout serially; after expiry, storage is also invoked with a known-invalid SAS. Cache a failed-refresh backoff/result, fall back only while `expires_at > now`, and otherwise return the refresh error.

cpp/src/format/bridge/rust/src/gcp_impersonation.rs
```diff
@@ -460,24 +478,293 @@ impl ObjectStoreProvider for ImpersonatingGcsStoreProvider {
...
+impl GcpImpersonationStorage {
+    async fn current_storage(&self) -> IcebergResult<Arc<dyn IcebergStorage>> {
+        let provider = self.provider.as_ref().ok_or_else(|| {
+            IcebergError::new(
+                IcebergErrorKind::Unexpected,
+                "GCP impersonation runtime provider is unavailable",
+            )
+        })?;
+        let credential = provider
+            .get_credential()
```
buqian-zilliz, line 583: **[P2] Use the still-valid GCP token when refresh fails**

Once the cached token enters the 300-second refresh window, `get_credential()` propagates any metadata or IAM refresh error without checking whether that token is still unexpired. This new Iceberg call runs before every storage operation, so a transient refresh outage fails all operations up to five minutes early and concurrent callers can repeatedly retry IAM. Keep serving the cached bearer until its actual `expireTime`, with retry backoff, and fail only when no usable credential remains.

cpp/src/format/bridge/rust/src/iceberg_bridgeimpl.rs
```diff
@@ -145,6 +127,42 @@ async fn upstream_opendal_factory(
...
     match scheme {
+        "abfs" | "abfss" | "wasb" | "wasbs" if azure_broker.is_some() => {
+            let config = azure_broker.expect("guarded by is_some");
+            let inner = azdls_factory(scheme)?;
+            let scoped_cache_key = cache_key
+                .as_deref()
+                .map(|cache_key| format!("azure:{scheme}:{cache_key}"));
+            match scoped_cache_key.as_deref() {
+                Some(cache_key) => ICEBERG_FACTORY_CACHE
+                    .get(cache_key, || async {
```
buqian-zilliz, line 138: **[P2] Do not hold the global cache lock while warming credentials**

`GlobalLruCache::get` holds its single entries mutex while awaiting `create()`. These new Azure and GCP constructors perform credential-warming network calls before returning, so one slow cold miss blocks even cache hits for unrelated Iceberg identities and providers; the analogous Lance lookups have the same behavior within their caches. Use per-key single-flight initialization and hold the LRU mutex only for lookup and publication.

cpp/src/format/bridge/rust/src/gcp_impersonation.rs
```diff
@@ -460,24 +478,293 @@ impl ObjectStoreProvider for ImpersonatingGcsStoreProvider {
...
+            .await
+            .map_err(gcp_iceberg_credential_error)?;
+        let mut props = self.props.clone();
```
buqian-zilliz: **[P2] Reuse delegated Iceberg storage until the token changes**

Every Iceberg operation clones the full property map, allocates a new `StorageConfig`, and calls `OpenDalStorageFactory::build`; that factory clones and reparses the properties again. The Azure wrapper duplicates this pattern. Since a credential is normally stable for many minutes, large plans repeatedly rebuild identical delegated storage for every metadata, manifest, and delete-file I/O. Cache the delegated storage alongside the credential identity and rebuild it only when the token changes.

cpp/src/format/bridge/rust/src/azure_sas_provider.rs
```diff
@@ -351,32 +353,190 @@ impl StorageOptionsProvider for AzureSasStorageOptionsProvider {
...
+    IcebergError::new(
+        IcebergErrorKind::Unexpected,
+        format!("Azure SAS credential resolution failed: {error}"),
+    )
+}
+
+#[derive(Clone, Serialize, Deserialize)]
+pub(crate) struct AzureSasStorageFactory {
+    inner: OpenDalStorageFactory,
+    #[serde(skip)]
```
buqian-zilliz, line 388: **[P2] Keep typetag-deserialized storage usable**

The factory skips its only runtime provider, while the corresponding storage object skips both its provider and properties. Deserialization therefore produces `provider = None`, after which `build` or every I/O fails with "runtime provider is unavailable"; the GCP wrappers repeat the same pattern. This breaks Iceberg's documented typetag round-trip contract. Preserve enough non-secret configuration to reconstruct the provider lazily instead of serializing an unusable object.

cpp/src/format/bridge/rust/src/azure_sas_provider.rs
jiaqizho: ack

cpp/src/format/bridge/rust/src/gcp_impersonation.rs
jiaqizho: ack

cpp/src/format/bridge/rust/src/iceberg_bridgeimpl.rs
jiaqizho: expect, it's fine for now

cpp/src/format/bridge/rust/src/gcp_impersonation.rs
jiaqizho: ack

cpp/src/format/bridge/rust/src/azure_sas_provider.rs
jiaqizho: won;t use this path

cpp/src/format/bridge/rust/src/azure_sas_provider.rs
```diff
@@ -351,32 +353,221 @@ impl StorageOptionsProvider for AzureSasStorageOptionsProvider {
...
+    async fn delete(&self, path: &str) -> IcebergResult<()> {
+        self.current_storage().await?.delete(path).await
+    }
+
+    async fn delete_prefix(&self, path: &str) -> IcebergResult<()> {
+        self.current_storage().await?.delete_prefix(path).await
+    }
+
+    fn new_input(&self, path: &str) -> IcebergResult<InputFile> {
+        Ok(InputFile::new(Arc::new(self.clone()), path.to_string()))
```
buqian-zilliz, line 538: **[P2] Share immutable storage properties across Iceberg file handles**

`new_input()` wraps `self.clone()` in a new `Arc`, but cloning `AzureSasStorage` deep-copies the owned `props` map. Iceberg planning creates input handles for the table metadata, manifest list, every manifest, and each positional-delete file, so large plans repeatedly allocate and copy the same configuration; `new_output()` and `GcpImpersonationStorage::new_input()` repeat the pattern. Keep immutable properties behind an `Arc<HashMap<...>>` (as the Aliyun wrapper does) or otherwise make storage cloning shallow.

cpp/src/format/bridge/rust/src/azure_sas_provider.rs
buqian-zilliz: This issue still exists: the expired-token fallback is gone, but a transient refresh failure still discards a still-valid cached SAS and leaves every queued caller free to repeat the broker timeout; please retain valid credentials with bounded retry backoff and fail only after expiry.

cpp/src/format/bridge/rust/src/gcp_impersonation.rs
buqian-zilliz: This issue still exists: refresh failures are still propagated while the cached token is valid, so Iceberg operations can fail up to five minutes early and queued callers can repeatedly retry IAM.

cpp/src/format/bridge/rust/src/azure_sas_provider.rs
```diff
@@ -351,32 +353,221 @@ impl StorageOptionsProvider for AzureSasStorageOptionsProvider {
...
+    async fn metadata(&self, path: &str) -> IcebergResult<FileMetadata> {
+        self.current_storage().await?.metadata(path).await
+    }
+
+    async fn read(&self, path: &str) -> IcebergResult<bytes::Bytes> {
+        self.current_storage().await?.read(path).await
+    }
+
+    async fn reader(&self, path: &str) -> IcebergResult<Box<dyn FileRead>> {
+        self.current_storage().await?.reader(path).await
```
buqian-zilliz, line 518: **[P2] Keep Iceberg handles refresh-aware after creation**

`reader()` calls `current_storage()` only while creating the handle and then returns OpenDAL's `FileRead`, so every later range read bypasses this wrapper. A cached SAS can have only slightly more than the 60-second refresh margin remaining; if that continuous reader outlives the token, it keeps using the expired SAS even after the shared provider has refreshed. `GcpImpersonationStorage::reader()` has the same lifecycle for its bearer. Return a refresh-aware reader that reacquires current storage for each range read, and define an equivalent credential-lifetime strategy for `writer()`.

cpp/src/format/bridge/rust/src/azure_sas_provider.rs
buqian-zilliz: This issue still exists: refresh failures are still propagated while the cached SAS is valid, and there is still no retry backoff to prevent queued callers from repeating the broker timeout.

cpp/src/format/bridge/rust/src/gcp_impersonation.rs
buqian-zilliz: This issue still exists: refresh failures are still returned while the cached token remains valid, and queued callers can immediately repeat the metadata/IAM refresh without backoff.

cpp/src/format/bridge/rust/src/azure_sas_provider.rs
jiaqizho: ack

lentitude2tk (approved):
/lgtm

PR #620, opened 2026-08-11
feat: publish index artifacts through transactions

yuruiz:
## Summary

- publish completed index metadata through `Transaction::AddIndexInfo`
- persist typed artifact identity, index name, file keys, sizes, row count, and engine versions in manifest `Index`
- expose the same metadata through `LoonIndexInfo` and C FFI
- remove `ManifestMinorVersion`; index presence is represented by manifest `indexes`
- document the manifest-first, minimal-etcd QueryCoord/QueryNode load design

## Testing

- builder-container clang-format dry run
- builder-container build of `milvus_test` and `Test_FFI`
- focused manifest / transaction / FFI GTests (5 passed)
- `Test_FFI` (78 tests, 0 failed)

### DIFF
--- a/cpp/include/milvus-storage/ffi_c.h
+++ b/cpp/include/milvus-storage/ffi_c.h
@@ -284,14 +284,6 @@ typedef struct LoonIndexes {
 } LoonIndexes;
 
 /**
- * @brief Result of committing a manifest revision.
- */
-typedef struct LoonManifestCommitInfo {
-  int64_t manifest_version;
-  int32_t manifest_minor_version;
-} LoonManifestCommitInfo;
-
-/**
  * @brief C structure representing a Manifest
  */
 typedef struct LoonManifest {
@@ -757,16 +749,6 @@ FFI_EXPORT LoonFFIResult loon_transaction_get_read_version(LoonTransactionHandle
  * @return result of FFI
  */
 FFI_EXPORT LoonFFIResult loon_transaction_commit(LoonTransactionHandle handle, int64_t* out_committed_version);
-
-/**
- * @brief Commits a transaction and returns the complete manifest reference.
- *
- * @param handle Transaction handle
- * @param out_commit_info Output committed revision and manifest minor version
- * @return result of FFI
- */
-FFI_EXPORT LoonFFIResult loon_transaction_commit_with_info(LoonTransactionHandle handle,
-                                                           LoonManifestCommitInfo* out_commit_info);
 
 /**
  * @brief Destroys a Transaction
--- a/cpp/include/milvus-storage/ffi_c.h
+++ b/cpp/include/milvus-storage/ffi_c.h
@@ -267,12 +267,29 @@ typedef struct LoonLobFiles {
 } LoonLobFiles;
 
 /**
- * @brief Metadata for one already-written index file.
+ * @brief Metadata for one completed index artifact.
+ *
+ * Fields through index_file_keys are the typed metadata required to load an
+ * artifact. property_keys/property_values are reserved for index-specific
+ * parameters such as metric_type and Knowhere options.
  */
 typedef struct LoonIndexInfo {
   const char* column_name;
+  const char* index_name;
   const char* index_type;
-  const char* path;
+  const char* path;  ///< Artifact directory/prefix
+  int64_t field_id;
+  int64_t index_id;
+  int64_t build_id;
+  int64_t index_version;
+  int64_t num_rows;
+  int64_t serialized_size;
+  int64_t mem_size;
+  int32_t current_index_version;
+  int32_t current_scalar_index_version;
+  int32_t index_store_path_version;
+  const char** index_file_keys;
+  uint32_t num_index_file_keys;
   const char** property_keys;
   const char** property_values;
   uint32_t num_properties;
@@ -299,10 +316,9 @@ typedef struct LoonManifest {
   // LOB files (TEXT column metadata)
   LoonLobFiles lob_files;
 
-  // Index metadata and persistent feature marker. Appended to retain the
-  // layout of the existing prefix for older consumers.
+  // Index metadata. Appended to retain the layout of the existing prefix for
+  // older consumers.
   LoonIndexes indexes;
-  int32_t minor_version;
 } LoonManifest;
 
 /**
@@ -830,8 +846,7 @@ FFI_EXPORT LoonFFIResult loon_transaction_update_stat(LoonTransactionHandle hand
  * The transaction stores metadata only; it does not create or write the file
  * identified by index_info->path.
  */
-FFI_EXPORT LoonFFIResult loon_transaction_add_index_info(LoonTransactionHandle handle,
-                                                         const LoonIndexInfo* index_info);
+FFI_EXPORT LoonFFIResult loon_transaction_add_index_info(LoonTransactionHandle handle, const LoonIndexInfo* index_info);
 
 /**
  * @brief Add a LOB file info to the transaction updates
--- a/cpp/include/milvus-storage/ffi_c.h
+++ b/cpp/include/milvus-storage/ffi_c.h
@@ -267,6 +267,31 @@ typedef struct LoonLobFiles {
 } LoonLobFiles;
 
 /**
+ * @brief Metadata for one already-written index file.
+ */
+typedef struct LoonIndexInfo {
+  const char* column_name;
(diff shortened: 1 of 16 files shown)


cpp/include/milvus-storage/transaction/transaction.h
```diff
@@ -35,6 +35,14 @@ namespace milvus_storage::api::transaction {
  */
 static constexpr int64_t LATEST = -1;
 
+/**
+ * @brief Metadata needed to publish an immutable manifest revision.
+ */
+struct ManifestCommitInfo {
```
jiaqizho: I am not sure what `manifest_minor_version` for in this PR?

The `version` field in the commit represents the updated version number of the manifest.

In contrast, `constexpr int32_t MANIFEST_VERSION = 5;` represents the version number of the manifest itself (meaning this version number must be incremented whenever the payload is expanded).

These are two distinct version numbers.

cpp/include/milvus-storage/ffi_c.h
```diff
@@ -798,6 +824,15 @@ FFI_EXPORT LoonFFIResult loon_transaction_update_stat(LoonTransactionHandle hand
                                                       const char* const* metadata_values,
                                                       size_t metadata_len);
 
+/**
+ * @brief Record metadata for an index file that has already been written.
+ *
+ * The transaction stores metadata only; it does not create or write the file
+ * identified by index_info->path.
+ */
+FFI_EXPORT LoonFFIResult loon_transaction_add_index_info(LoonTransactionHandle handle,
```
jiaqizho: need export to `cpp/ffi_exports.map` and `cpp/ffi_exports_mac.map`.

cpp/include/milvus-storage/manifest.h
```diff
@@ -37,7 +37,20 @@ namespace milvus_storage::api {
...
+constexpr int32_t MANIFEST_VERSION = 6;
+
+/**
+ * @brief Persistent feature marker for a manifest revision.
+ *
+ * This is intentionally distinct from MANIFEST_VERSION, which denotes the
+ * serialization format.  The value is derived from the final manifest state:
+ * a manifest with index metadata is INDEX_INFO and one without it is NONE.
+ */
+enum class ManifestMinorVersion : int32_t {
```
jiaqizho: also no need? already updated `MANIFEST_VERSION`.

we can just check index payload exist to make sure index info exist or not.

cpp/src/ffi/bridge.cpp
```diff
@@ -312,6 +315,42 @@ arrow::Status manifest_export(const std::shared_ptr<milvus_storage::api::Manifes
...
+      // later allocation throws, the catch block below delegates cleanup of
+      // this partially constructed object to loon_manifest_destroy().
+      (*out_cmanifest)->indexes.indexes = new LoonIndexInfo[indexes.size()]{};
+      (*out_cmanifest)->indexes.num_indexes = static_cast<uint32_t>(indexes.size());
+
+      for (size_t i = 0; i < indexes.size(); ++i) {
+        const auto& index = indexes[i];
+        auto& out_index = (*out_cmanifest)->indexes.indexes[i];
+
+        auto copy_string = [](const std::string& value) {
```
jiaqizho: I just notice that, in `manifest_export` function, multi place use will do 
```
auto* result = new char[value.size() + 1];
std::memcpy(result, value.c_str(), value.size() + 1);
```

can we just replace it with `strdup` or move copy_string into the begin of this function and replace all string copy?

cpp/include/milvus-storage/ffi_c.h
```diff
@@ -266,6 +266,23 @@ typedef struct LoonLobFiles {
   uint32_t num_files;      ///< Number of LOB files
 } LoonLobFiles;
 
+/**
+ * @brief Metadata for one already-written index file.
+ */
+typedef struct LoonIndexInfo {
```
jiaqizho, line 276: is that enough? 


```
type SegmentIndex struct {
	SegmentID      int64
	CollectionID   int64
	PartitionID    int64
	NumRows        int64
	IndexID        int64
	BuildID        int64
	NodeID         int64
	IndexVersion   int64
	IndexState     commonpb.IndexState
	FailReason     string
	IsDeleted      bool
	CreatedUTCTime uint64
	IndexFileKeys  []string
	// The byte size of serialized index data at oos. (compressed)
	IndexSerializedSize uint64
	// The byte size of index data in memory. (uncompressed)
	// The IndexMemSize may not be stored at old milvus implementation, so it may be not accurate.
	// (generated by the IndexSerializedSize multiplied with a configured compress-ratio).
	IndexMemSize uint64
	// deprecated
	WriteHandoff              bool
	CurrentIndexVersion       int32
	FinishedUTCTime           uint64
	CurrentScalarIndexVersion int32
	IndexType                 string
	// Zero value is the legacy build-rooted layout and remains wire-compatible with old metadata.
	// IndexStorePathVersion is the requested path version while IndexState is
	// Unissued/InProgress, and reflects the actual on-disk layout only after
	// FinishTask. Readers that construct paths from this field must check
	// IndexState == Finished first.
	IndexStorePathVersion indexpb.IndexStorePathVersion
}
```

what about `IndexID`/`BuildID` and some field of `IndexVersion` ?

cpp/src/transaction/transaction.cpp
```diff
@@ -495,11 +495,15 @@ Transaction& Transaction::UpdateStat(const std::string& key, const Statistics& s
   return *this;
 }
 
-Transaction& Transaction::AddIndex(const Index& index) {
+Transaction& Transaction::AddIndexInfo(const Index& index) {
   updates_.AddIndex(index);
   return *this;
 }
 
+Transaction& Transaction::AddIndex(const Index& index) {
```
jiaqizho: no need dup define the same function with different function name.

cpp/include/milvus-storage/ffi_c.h
jiaqizho: Also 

```
cd cpp/
make fix-format
```

cpp/include/milvus-storage/manifest.h
```diff
@@ -84,14 +85,26 @@ struct DeltaLog {
...
+  int64_t build_id = 0;
+  int64_t index_version = 0;
+  int64_t num_rows = 0;
+  int64_t serialized_size = 0;  ///< Total object-storage bytes
+  int64_t mem_size = 0;         ///< Estimated loaded memory bytes
+  int32_t current_index_version = 0;
+  int32_t current_scalar_index_version = 0;
+  int32_t index_store_path_version = 0;
+  std::vector<std::string> index_file_keys;       ///< File names relative to path
   std::map<std::string, std::string> properties;  ///< Index-specific properties
```
jiaqizho: std::unordered_map ?

cpp/src/manifest.cpp
```diff
@@ -463,7 +499,20 @@ void Manifest::deserializeLegacy(std::istream& input_stream) {
   }
 
   if (version >= 2) {
```
jiaqizho: `version >= 6`？

jiaqizho: Please add comments; the index section of manifest version 2 has not been used by any caller.

cpp/include/milvus-storage/ffi_c.h
```diff
@@ -266,6 +266,40 @@ typedef struct LoonLobFiles {
...
 } LoonLobFiles;
 
+/**
+ * @brief Metadata for one completed index artifact.
+ *
+ * Fields through index_file_keys are the typed metadata required to load an
+ * artifact. property_keys/property_values are reserved for index-specific
+ * parameters such as metric_type and Knowhere options.
+ */
+typedef struct LoonIndexInfo {
```
jiaqizho, line 276: consider add these entry into `LoonIndexInfo`?

```
IsDeleted bool,
IndexState IndexState, // use the uint8_t
```

yuruiz:
Addressed the latest review comments in commit `5bafce4`:

- kept `Index::properties` as `std::unordered_map` for the common key-lookup path
- split legacy raw MILV index decoding by manifest version; v2-v5 use the four-field layout, while v6+ uses the typed layout
- added compatibility comments and kept `LoonIndexInfo` limited to completed artifact metadata; `IsDeleted`/`IndexState` remain transient build-task state

Validation: legacy v2/v3 tests passed and `Test_FFI` passed 78/78.

Issue #623, opened 2026-08-13
[bug]: GCP default port breaks credential lookup


xiaocai2333:
## Summary

GCP S3-compatible filesystem requests fail with a locally generated HTTP 403 `SignatureFailed` when the configured endpoint includes the default HTTPS port:

```text
storage.googleapis.com:443
```

The GCP credential provider is registered under an endpoint key containing `:443`, while the AWS SDK normalizes outgoing HTTPS request URIs and omits the default port. The registry lookup then cannot find the registered provider.

## Affected version

Observed with:

```text
Milvus image: milvus:3.0-20260808-1439dc7de-500a4a0
Milvus commit: 1439dc7de8b198a01c2afa0ae20c0c473e0e1abc
milvus-storage commit: 63c29c674bf8c75a84c49cca2c8ab088e771e57e
```

The issue is associated with the registry-backed GCP credential implementation introduced in milvus-storage commit `977044c` / PR #496. Older 2.6 builds using the previous GCP HTTP factory did not perform this endpoint-based provider lookup.

## Reproduction configuration

```yaml
address: storage.googleapis.com:443
bucket_name: <bucket>
cloud_provider: gcp
use_ssl: true
use_iam: true
```

A request generated by the AWS SDK has the following form:

```text
https://storage.googleapis.com/<bucket>/<object>
```

The worker logs then contain:

```text
GoogleHttpClientFactory: no GcpCredentialProvider registered for URI: https://storage.googleapis.com/<bucket>/<object>

HTTP response code: 403
Exception name: SignatureFailed
Error message: No GcpCredentialProvider registered for URI: https://storage.googleapis.com/<bucket>/<object>
```

The response includes no remote host IP, request ID, or response headers. This is the synthetic response created inside milvus-storage; the request does not reach GCS.

## Root cause

The provider is registered in `GcpFileSystemProducer::RegisterIdentity()` using the configured endpoint:

```cpp
GcpBucketKey key{
    NormalizeGcpEndpointHost(config.address),
    config.bucket_name
};
GcpCredentialRegistry::Instance().Register(std::move(key), std::move(provider));
```

In the affected version, `NormalizeGcpEndpointHost("storage.googleapis.com:443")` strips only the scheme/trailing slash and lowercases the string, so the registered endpoint remains:

```text
storage.googleapis.com:443
```

The lookup path uses `uri.GetAuthority()`. For an HTTPS URI using the default port, the AWS SDK authority is:

```text
[...]

PR #624, opened 2026-08-14
fix: normalize GCP credential registry endpoint matching

jiaqizho:
issue: #623

GCP credential providers were registered from filesystem configuration, while outgoing requests were matched from an AWS URI. The two paths could describe the same endpoint differently—for example, one side could include an explicit default port while the other omitted it—so the registry failed to find the provider even though the request was going to the configured bucket.

The registry now uses a structured endpoint key made up of the scheme, effective port, and lowercased host. Both registration and lookup build that key through the same URI-based normalization path, with an explicit HTTP or HTTPS scheme taking precedence and use_ssl filling it in when the address does not include one.

Path-style and virtual-host-style requests now reuse the same endpoint identity, including default ports, non-default ports, and IPv6 hosts. This makes equivalent endpoint forms resolve to the same credential provider while still keeping genuinely different schemes or ports isolated.

### DIFF
--- a/cpp/include/milvus-storage/filesystem/gcp/gcp_credential_registry.h
+++ b/cpp/include/milvus-storage/filesystem/gcp/gcp_credential_registry.h
@@ -14,6 +14,8 @@
 
 #pragma once
 
+#include <cstdint>
+#include <functional>
 #include <memory>
 #include <mutex>
 #include <string>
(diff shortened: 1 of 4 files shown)


yuruiz:
LGTM

yuruiz (approved):

PR #630, opened 2026-08-19
feat: expose cached filesystem metrics

jiaqizho:
Filesystem metrics could previously only be read through an individual filesystem handle, which made it difficult for callers to inspect every filesystem currently held by the global cache. Add a cache-wide listing API so callers can discover active filesystems and collect their latest metrics snapshots in one call.

Keep the existing hashed cache key as the internal LRU identity, while storing a separate human-readable display key alongside each cached filesystem. Local filesystems use file://{root_path}#{cache_key}, while remote filesystems use {address}/{bucket_name}#{cache_key}, with an empty address shown as <null>. The display key is created once when the filesystem enters the cache, and normal cache lookups continue to use the original hash without repeatedly copying the display string.

Expose the cached metrics list through the C FFI with explicit result structures and a matching release function for owned memory. Reuse the existing metrics snapshot conversion logic, clean up partially allocated results when an entry cannot be reported, and publish the new functions through both Linux and macOS symbol export lists.

### DIFF
--- a/cpp/include/milvus-storage/ffi_filesystem_metrics_c.h
+++ b/cpp/include/milvus-storage/ffi_filesystem_metrics_c.h
@@ -47,6 +47,22 @@ typedef struct LoonFilesystemMetricsSnapshot {  // NOLINT
 } LoonFilesystemMetricsSnapshot;  // NOLINT
 
 /**
+ * Metrics for one cached filesystem.
+ */
+typedef struct LoonFilesystemMetricsEntry {  // NOLINT
+  char* display_key;
+  LoonFilesystemMetricsSnapshot metrics;
+} LoonFilesystemMetricsEntry;  // NOLINT
+
+/**
+ * Metrics for all cached filesystems.
+ */
+typedef struct LoonFilesystemMetricsList {  // NOLINT
+  LoonFilesystemMetricsEntry* entries;
+  uint32_t count;
+} LoonFilesystemMetricsList;  // NOLINT
+
+/**
  * Get metrics from a filesystem handle.
  * Returns metrics if the filesystem is observable, otherwise returns error.
  *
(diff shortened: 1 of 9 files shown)


cpp/src/filesystem/fs.cpp
```diff
@@ -191,12 +191,29 @@ FilesystemCache& FilesystemCache::getInstance() {
 
 size_t FilesystemCache::size() const { return cache_.size(); }
 
+std::vector<std::pair<std::string, ArrowFileSystemPtr>> FilesystemCache::list() const {
```
weiliu1031, line 194: Could we decouple metrics enumeration from LRU membership? A `FileSystemWrapper` holds its own `shared_ptr`, so eviction or `clean()` can remove a filesystem from `cache_` while an existing handle continues performing I/O and incrementing its metrics. This method will no longer report that live instance. If the same configuration is later recreated, it also receives the same display key with counters reset to zero. Please keep a metrics registry/accumulator independent of the LRU lifecycle and add a capacity-1 test that retains A, inserts B to evict A, then performs I/O through A and verifies that the new traffic remains observable.

cpp/src/filesystem/fs.cpp
```diff
@@ -191,12 +191,29 @@ FilesystemCache& FilesystemCache::getInstance() {
...
 
 void FilesystemCache::clean() { cache_.clean(); }
 
 void FilesystemCache::set_capacity(size_t capacity) { cache_.set_capacity(capacity); }
 
+std::string FilesystemCache::MakeDisplayKey(const ArrowFileSystemConfig& config, const std::string& cache_key) {
+  if (config.storage_type == "local") {
+    return "file://" + config.root_path + "#" + cache_key;
+  }
+  return (config.address.empty() ? "<null>" : config.address) + "/" + config.bucket_name + "#" + cache_key;
```
weiliu1031, line 214: Please avoid exposing the internal cache key as part of the public display key. `GetCacheKey()` uses `std::hash` over credential-related fields, including `access_key_value`, `external_id`, and credential identities. Returning that value through the FFI exposes a deterministic credential fingerprint and turns an implementation-specific hash into a public metrics-label contract. Please use an opaque per-process identifier or a keyed fingerprint that does not reveal a reusable hash of the credential-bearing configuration.

cpp/include/milvus-storage/ffi_filesystem_metrics_c.h
```diff
@@ -57,6 +73,22 @@ typedef struct LoonFilesystemMetricsSnapshot {  // NOLINT
...
                                                      LoonFilesystemMetricsSnapshot* out_metrics);
 
+/**
+ * List metrics for all cached filesystems.
+ *
+ * @param out_list The output metrics list. Must be released with
+ *                 loon_filesystem_free_metrics_list.
+ * @return result of FFI
+ */
+FFI_EXPORT LoonFFIResult loon_filesystem_list_metrics(LoonFilesystemMetricsList* out_list);
```
weiliu1031, line 83: The Python CFFI declarations manually mirror this metrics interface, but `python/milvus_storage/_ffi.py` still declares only `LoonFilesystemMetricsSnapshot`, `loon_filesystem_get_metrics`, and `loon_filesystem_reset_metrics`. As a result, Python callers cannot access the new list structures or functions. Please add the corresponding CFFI declarations and a wrapper/free-path test, or explicitly document that this API is intentionally limited to native C/cgo consumers.

weiliu1031 (commented):
Please see the inline comments.

Issue #633, opened 2026-08-24
[2.6] Parquet writer creates undersized row groups for zero-copy sliced RecordBatches


presburger:
## Problem

The Parquet writer may create undersized row groups when writing an Arrow `RecordBatch` produced by `Array::Slice()` or `RecordBatch::Slice()`.

Arrow slicing is zero-copy. A sliced array continues to reference the complete buffers of the original array, even when the slice contains only a small number of rows.

`ParquetFileWriter` currently calculates the batch size using:

```cpp
auto batch_size = GetRecordBatchMemorySize(record);
```

`GetRecordBatchMemorySize()` counts the complete underlying buffers. As a result, a small slice can be reported as consuming the same amount of memory as its much larger parent array.

`ParquetFileWriter::Flush()` then uses this value to estimate the average row size:

```cpp
double avg_row_size = static_cast<double>(batch_size) / total_rows;
auto max_rows = static_cast<int64_t>(remain_size / avg_row_size);
```

This overestimates `avg_row_size`, reduces `max_rows`, and causes the writer to create row groups containing very little actual data.

## Example

Consider a `FixedSizeBinaryArray` with the following properties:

- 4,096 rows
- 512 bytes per row
- Approximately 2 MiB of underlying data

A slice containing 64 rows logically contains approximately 32 KiB:

```cpp
auto sliced = array->Slice(1024, 64);
```

However, because the slice still references the original buffer, `GetRecordBatchMemorySize()` reports approximately 2 MiB.

With a 1 MiB target row-group size, the writer estimates:

```text
average row size = 2 MiB / 64 = 32 KiB
rows per row group = 1 MiB / 32 KiB = 32
```

The 64-row slice can therefore be split into two row groups. Each row group contains only 32 rows and approximately 16 KiB of actual data.

The same problem occurs with a zero-offset prefix slice:

```cpp
auto sliced = array->Slice(0, 64);
```

This case cannot be reliably detected using `Array::offset()`, because the offset remains zero even though the slice retains the original buffers.

## Expected Behavior

Row-group sizing should be based on the data represented by the sliced batch, rather than the complete buffers retained from its parent array.

Small slices should be combined with other batches until the target row-group size is approached, instead of producing very small row groups.

## Proposed Solution

Materialize the arrays in the incoming `RecordBatch` before caching the batch and calculating its memory size:

```cpp
std::vector<std::shared_ptr<arrow::Array>> columns;
columns.reserve(record->num_columns());

[...]

PR #635, opened 2026-08-24
fix: [2.6] materialize batches before row group sizing

presburger:
issue: #633

### DIFF
--- a/cpp/src/format/parquet/parquet_writer.cpp
+++ b/cpp/src/format/parquet/parquet_writer.cpp
@@ -26,9 +26,21 @@
 #include <boost/filesystem/operations.hpp>
 #include <memory>
 
+#include <arrow/array/concatenate.h>
 #include <arrow/io/buffered.h>
 
 namespace milvus_storage::parquet {
+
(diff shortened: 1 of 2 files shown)


xiaofan-luan:
I think `arrow::util::ReferencedBufferSize(*record)` is a better fit for row-group sizing here:

```cpp
ARROW_ASSIGN_OR_RAISE(auto referenced_size,
                      arrow::util::ReferencedBufferSize(*record));
auto batch_size = static_cast<size_t>(referenced_size);
cached_batches_.push_back(record);
```

It measures the buffers referenced by the slice instead of charging the full parent backing buffers, so it fixes the row-group sizing issue while avoiding the extra allocation and full-batch copy introduced by `materialize_batch()` / `Concatenate()` on every write. The existing sliced-batch test should cover this implementation as well.

I would keep `GetRecordBatchMemorySize` unchanged in `PackedRecordBatchWriter` / `ColumnGroup`, because those call sites are retained-memory backpressure rather than logical row-group sizing.

presburger:
@xiaofan-luan That would be a better approach. Materializing the batch introduces an extra full-batch copy and is less efficient. Thanks for the suggestion!

PR #636, opened 2026-08-24
enhance: replace Lance ObjectStore with filesystem-backed storage

jiaqizho:
The Lance bridge previously used one object-store setup for both reads and writes, even though the two paths have different ownership, credential, and lifecycle requirements. Readers need to reuse the filesystem that milvus-storage has already resolved and cached, while writers still need Lance’s native mutation-capable object stores. Keeping those paths together duplicated cloud-provider logic, made metrics and retry behavior inconsistent, and created unnecessary provider state when callers opened large numbers of readers.

This change moves Lance reads onto a read-only ObjectStore backed by the existing C++ Arrow FileSystem and passes that filesystem across the CXX boundary with shared ownership. The adapter handles object metadata, directory listing, range reads, local and remote path mapping, CRT asynchronous reads, bounded metadata caching, and shared scan scheduling. Reader instances can now reuse the same filesystem and cloud-provider resources instead of building another Rust-side storage stack for every dataset.

Lance writes now follow a separate writer configuration and continue using native Lance object stores. Dynamic credential modes that cannot be supported safely by the writer path fail explicitly instead of silently falling back to a different identity. The C++ reader constructors, storage option conversion, benchmarks, and tools are updated to make the split visible and keep filesystem lifetimes valid for as long as Rust-backed readers need them.

The filesystem FFI now exposes richer object metadata and preserves detailed storage error codes across the C++ and Rust boundary. Not-found, permission, precondition, and unsupported errors are mapped to their matching object_store variants. Throttling errors remain visible to Lance AIMD, and range reads complete inside the AIMD retry boundary so throttling can actually reduce the request rate and trigger the expected retry behavior.

Common Arrow C Data ownership and error translation are also shared across the Lance, Vortex, and Paimon bridges. Exported Arrow structures are released correctly when exceptions occur, and detailed storage errors survive the round trip back to C++ without changing the existing format-level behavior.


cpp/src/format/bridge/rust/src/lance_object_store.rs
```diff
@@ -0,0 +1,1666 @@
...
+    fs_cache_key: String,
+    object_store: &Arc<ObjectStore>,
+) -> Result<Arc<ScanScheduler>> {
+    let mut schedulers = LANCE_IO_SCHEDULERS
+        .lock()
+        .map_err(|_| LanceError::Internal {
+            message: "Lance I/O scheduler registry mutex poisoned".into(),
+            location: snafu::location!(),
+        })?;
+    if let Some(scheduler) = schedulers.get(&fs_cache_key).and_then(Weak::upgrade) {
```
liliu-z, line 416: shared_scan_scheduler looks the registry up by fs_cache_key alone, and ArrowFileSystemConfig::GetCacheKey() does not hash iops_initial_rate/iops_max_rate. Two readers whose configs differ only in those rates therefore share one scheduler, and with it the first reader's AimdThrottledStore, so the later reader's throttling settings are silently dropped. The base revision hashed every storage option except parallelism and the cache key and had a test asserting a different fingerprint produced a different scheduler; both were removed here. The doc comment above the function does acknowledge first-reader-wins, but folding the AIMD options back into the key would keep it from being silent.

cpp/src/format/bridge/rust/src/lance_bridgeimpl.rs
```diff
@@ -510,201 +310,47 @@ impl BlockingDataset {
...
-            storage_options,
-            parallelism,
-        )?;
-        dataset
-            .scan_scheduler
-            .set(scheduler)
-            .expect("a newly opened BlockingDataset has no scan scheduler");
-    }
+
+    let scheduler = shared_scan_scheduler(
```
liliu-z, line 346: open_dataset always installs the scheduler returned by shared_scan_scheduler(read_options.object_store_prefix(), ...), and that registry is keyed on the filesystem cache key only. FilesystemCache is a 16-entry LRU, so after an eviction a later get() builds a new arrow::fs::FileSystem for the same cache key; while any earlier reader still holds the scheduler alive, the new reader's fragment I/O is routed through the scheduler's captured ObjectStore — i.e. through the old filesystem instance and its old credential provider — rather than the one it just resolved. Retaining filesystem_ in the reader is fine; the fix is to fold the filesystem instance identity into the scheduler domain key.

cpp/src/format/bridge/rust/src/lance_object_store.rs
```diff
@@ -0,0 +1,1666 @@
...
+    }
+
+    fn list(
+        &self,
+        prefix: Option<&ObjectPath>,
+    ) -> BoxStream<'static, object_store::Result<ObjectMeta>> {
+        let store = self.clone();
+        let prefix = prefix.cloned().unwrap_or(ObjectPath::ROOT);
+        async_stream::try_stream! {
+            let result = store.list_impl(prefix, true).await?;
```
liliu-z: list() awaits list_impl(prefix, true), which materializes every entry — one FFI string copy per path plus a full sort — before the first yield. Lance's resolve_version_from_listing is written against a paginated stream and on the lexically-ordered path needs only the first entry plus take(999), so opening a dataset pays for the whole _versions/ directory regardless of commit count. Note the underlying loon_filesystem_list_dir FFI has no pagination, so fixing this means adding a paged listing call, not just restructuring the stream.

cpp/src/format/bridge/rust/src/bridge_util.cpp
```diff
@@ -0,0 +1,126 @@
...
+  if (status.ok()) {
+    return arrow::Status::OK();
+  }
+  if (ExtendStatusDetail::UnwrapStatus(status)) {
+    return MakeExtendErrorWithContext(context, status);
+  }
+  if (arrow::internal::ErrnoFromStatus(status) == ENOENT) {
+    return MakeIOErrorWithContext(context, status);
+  }
+  auto parsed_status = MakeBridgeErrorStatus(status.message());
```
liliu-z: When `LanceErrorTranslatingReader::ReadNext()` returns an ordinary `Invalid`/`TypeError` without an FFI marker (e.g. a corrupt batch producing a Rust non-I/O Arrow error), `MakeBridgeErrorStatus(status.message())` reparses the message and rebuilds it as `IOError`, so callers classify corrupt/invalid data as storage failure and may apply inappropriate retry handling. This path is new to Lance in this PR (base returned `ImportRecordBatchReader` unwrapped). Parse markers when present but otherwise preserve `status.code()`.

cpp/src/format/bridge/rust/src/lance_object_store.rs
```diff
@@ -0,0 +1,1666 @@
...
+    // reading provider environment variables.
+    let constructor_location =
+        Url::parse("file-object-store:///").expect("constant file-object-store URL is valid");
+    let mut store = ObjectStore::new(
+        inner.clone(),
+        constructor_location,
+        Some(block_size),
+        None,
+        false,
+        true,
```
liliu-z, line 316: build_unwrapped_ffi_store constructs the store from a file-object-store:// URL with list_is_lexically_ordered = true (positional arg 6), which turns off both of Lance's manifest fast paths: is_local() only matches file/file+uring, and uses_version_hint() requires !list_is_lexically_ordered. Every open therefore falls into resolve_version_from_listing, replacing Lance's single std::fs::read_dir for local datasets with a full recursive FFI listing that stats and strdups each entry. The flag itself is honest — list_impl sorts by location — and switching to the `file` scheme would make Lance read the local directory directly and bypass the C++ filesystem, so the fix is to cut the listing cost rather than to adopt a local scheme.

cpp/src/format/bridge/rust/src/lance_object_store.rs
jiaqizho: This is intentional. The scheduler registry represents one I/O domain per milvus_fs_cache_key, so matching readers share the same ScanScheduler, ObjectStore, and AIMD controller. The first active reader defines AIMD and parallelism until the last strong scheduler reference is dropped. Adding the AIMD options to the key would create independent limiters for the same filesystem and could let the aggregate request rate exceed the intended backend limit. I have updated the public property documentation and the in-code comments to make this first-reader-wins contract explicit.

cpp/src/format/bridge/rust/src/lance_bridgeimpl.rs
jiaqizho: This is also intentional. fs_cache_key is the logical filesystem identity. LRU eviction removes only the cache-owned reference; it does not invalidate an active filesystem lease. The shared scheduler keeps the first ObjectStore and C++ filesystem alive so existing and later same-key readers remain in one I/O and credential domain. A different configured credential identity produces a different cache key, while temporary credentials refresh inside the provider. Once the last active scheduler is dropped, the weak registry entry expires and a subsequent open uses the currently cached or newly created filesystem. Adding pointer identity would split one logical filesystem domain solely because of cache churn and recreate provider/scheduler state.

cpp/src/ffi/filesystem_c.cpp
```diff
@@ -233,6 +233,51 @@ LoonFFIResult loon_filesystem_get_file_info(FileSystemHandle handle,
...
+                         "] details: ", info_result.status().ToString());
+    }
+
+    const auto& info = info_result.ValueOrDie();
+    if (info.type() == arrow::fs::FileType::NotFound) {
+      RETURN_ERROR(LOON_FILE_NOT_FOUND, "File not found: ", path);
+    }
+
+    *out_is_dir = info.type() == arrow::fs::FileType::Directory;
+    *out_size = *out_is_dir ? 0 : static_cast<uint64_t>(info.size());
```
liliu-z, line 271: Arrow returns `kNoSize` (-1) for any FileInfo that is neither a regular file nor a directory (LocalFileSystem sets `FileType::Unknown` with `kNoSize` for fifos, sockets, and device nodes). The `is_dir` guard doesn't cover that case, so `static_cast<uint64_t>(-1)` yields `u64::MAX`, which the Rust adapter stores verbatim in `ObjectMeta.size`; a subsequent full-object `get_opts` computes `0..u64::MAX` and calls `BytesMut::with_capacity(usize::MAX)`, aborting the process. Treat `info.size() < 0` the same as a directory (or fail with a clear error) rather than casting. The same unguarded cast feeds `LoonFileInfo::size` in `loon_filesystem_list_dir`, which now also becomes `ObjectMeta.size` through the new listing path.

cpp/src/format/bridge/rust/src/filesystem_object_store.rs
```diff
@@ -0,0 +1,1074 @@
...
+    }
+
+    fn list(
+        &self,
+        prefix: Option<&ObjectPath>,
+    ) -> BoxStream<'static, object_store::Result<ObjectMeta>> {
+        let store = self.clone();
+        let prefix = prefix.cloned().unwrap_or(ObjectPath::ROOT);
+        async_stream::try_stream! {
+            let result = store.list_impl(prefix, true).await?;
```
liliu-z, line 858: `list()` awaits the full recursive listing before yielding anything, and the underlying `loon_filesystem_list_dir` materializes the entire result in one Arrow `GetFileInfo(selector)` call. Lance's `resolve_version_from_listing` only needs the first entry (plus up to 999 for a sanity check) because the store advertises `list_is_lexically_ordered = true`, so a `_versions` directory with many versions costs a full multi-page LIST on every dataset open where object_store's paged stream would have stopped after one page. Making `list()` lazy alone won't fix it — the pagination has to be pushed into the C listing FFI.

cpp/ffi_exports.map
```diff
@@ -19,6 +19,7 @@
     loon_errcode_fault_inject;
     loon_errcode_not_support;
     loon_errcode_file_not_found;
+    loon_errcode_permission_denied;
```
liliu-z, line 22: The new `loon_filesystem_get_object_info` is declared `FFI_EXPORT` in the public header but was never added to either export map, so under `WITH_PYTHON_BINDING` (which compiles with `-fvisibility=hidden` plus `local: *;`) it is hidden while every other public `loon_filesystem_*` symbol stays visible. Add it next to `loon_filesystem_get_file_info` in both ffi_exports.map and ffi_exports_mac.map.

cpp/src/format/bridge/rust/src/lance_object_store.rs
```diff
@@ -0,0 +1,685 @@
...
+impl std::fmt::Debug for FFIObjectStoreProvider {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str("FFIObjectStoreProvider(filesystem_c)")
+    }
+}
+
+#[async_trait::async_trait]
+impl ObjectStoreProvider for FFIObjectStoreProvider {
+    async fn new_store(&self, base_path: Url, _params: &ObjectStoreParams) -> Result<ObjectStore> {
+        self.mapper
```
liliu-z, line 335: A manifest whose `base_paths` point at a different bucket or scheme still opens successfully, because the identity check only runs when Lance lazily resolves the base store — `Dataset::base_object_store` calls `ObjectStore::from_uri_and_params(session.store_registry(), &base_path.path, ...)` at first fragment read. Reads then fail with `URI storage identity does not match the bound filesystem` (or an unknown-scheme registry error) long after open and metadata succeeded, which is hard to diagnose. Validate `manifest.base_paths` against the mapper during `open_dataset` and reject with a message that names the offending base URI.

cpp/src/format/bridge/rust/src/filesystem_object_store.rs
```diff
@@ -0,0 +1,1074 @@
...
+
+    let (read_result, completed_reader) =
+        receiver.await.map_err(|_| object_store::Error::Generic {
+            store: "filesystem_c",
+            source: Box::new(io::Error::new(
+                io::ErrorKind::BrokenPipe,
+                "async object read completion channel closed",
+            )),
+        })?;
+    drop(completed_reader);
```
liliu-z, line 170: `drop(completed_reader)` runs on the Tokio async worker that awaited the completion, and dropping the last `Arc<ReaderHandle>` invokes `loon_filesystem_reader_close` — an FFI call into `arrow::io::RandomAccessFile::Close()`, which is genuinely blocking for remote filesystems. The same function already defers this drop to `spawn_blocking` on both cancellation paths, so the success path should do the same (or move the reader into the blocking task that consumes it).

cpp/src/format/bridge/rust/src/filesystem_object_store.rs
jiaqizho: The early-stop behavior here only applies to V2 manifest naming. In resolve_version_from_listing, the V1 branch drains the stream to find the maximum version. milvus-storage explicitly sets enable_v2_manifest_paths = false in both the base revision and this PR, so datasets written by this project already caused the native object_store implementation to consume every LIST page.

The current adapter does add eager materialization, FFI string copies, and a full sort, and external V2 datasets lose the native first-page early stop. Those are valid overheads, but this is not a general increase in LIST page count for the V1 datasets we currently write. I would scope this concern to V2/external datasets and the additional materialization cost for V1.

cpp/src/format/bridge/rust/src/filesystem_object_store.rs
jiaqizho: The async branch is only entered when reader_supports_async sees NonBlockingReadAtFile. Today that is ObjectCrtInputFile, whose Close implementation only resets holder_ and sets closed_; it performs no remote request or wait. Readers without async support execute the read in spawn_blocking, and the moved ReaderHandle is dropped at the end of that blocking task.

The cancellation path uses spawn_blocking to keep FFI destruction off the CRT callback thread, not because this Close is known to block. Adding another blocking-pool dispatch to every successful CRT range read would add scheduling and queueing overhead without offloading blocking work.

Vortex uses the same ReaderHandle and assumption: ObjectStoreReadSourceCpp retains the owning Arc while async callbacks release temporary clones, and its eventual Close is the same implementation. If a future NonBlockingReadAtFile introduces a blocking Close, that should be handled at that implementation or lifecycle boundary. For the current CRT reader, I do not think the success-path drop needs spawn_blocking.

cpp/src/ffi/filesystem_c.cpp
jiaqizho: ACK

somfornot:
## Code review

Traced the FFI path/lifetime plumbing end to end — path mappers vs. `FileSystemProxy`'s `SubTreeFileSystem` rooting, `FileInfoListGuard` vs. `loon_filesystem_free_file_info_list` re-entry, `ArrowCDataReleaseGuard` arm/disarm against each Rust `from_raw` ordering, `ReaderHandle` drop paths, and AIMD option clamping in `create_file_system_config` — and those all look correct. Five findings below.

---

**1. New FFI symbol missing from the export maps** — [`cpp/ffi_exports.map:174`](https://github.com/milvus-io/milvus-storage/blob/028347e0930050d25562f14407695817bcccc609/cpp/ffi_exports.map#L174)

`loon_filesystem_get_object_info` is declared `FFI_EXPORT` in the public header but was never added to either export map. `local: *` hides it in `WITH_PYTHON_BINDING` builds, so the symbol won't resolve there.

**2. Status code lost when recovering an embedded error marker** — [`cpp/src/format/bridge/rust/src/bridge_util.cpp:133`](https://github.com/milvus-io/milvus-storage/blob/028347e0930050d25562f14407695817bcccc609/cpp/src/format/bridge/rust/src/bridge_util.cpp#L133)

A `NotImplemented` recovered from an embedded `__LOON_FFI_ERRCODE__=9` marker is rebuilt as `IOError` by `MakeIOErrorWithContext`, so the status code is dropped on the record-batch-reader translation path. ENOENT/EACCES survive only via the errno detail; callers that switch on `StatusCode` will misclassify.

**3. `shared_scan_scheduler` ignores the `filesystem` argument** — [`cpp/src/format/bridge/rust/src/lance_object_store.rs:404`](https://github.com/milvus-io/milvus-storage/blob/028347e0930050d25562f14407695817bcccc609/cpp/src/format/bridge/rust/src/lance_object_store.rs#L404)

The scheduler is keyed only on `fs_cache_key`, so the `filesystem` passed to later `BlockingDataset::Open` calls is silently discarded. After a 16-entry `FilesystemCache` eviction, fragment I/O keeps flowing through the evicted instance (which the scheduler retains forever) and bypasses the live filesystem's metrics.

**4. `list_dir` fails wholesale on one non-regular entry** — [`cpp/src/ffi/filesystem_c.cpp:988`](https://github.com/milvus-io/milvus-storage/blob/028347e0930050d25562f14407695817bcccc609/cpp/src/ffi/filesystem_c.cpp#L988)

The new size/type guard errors out the entire listing when a single entry isn't a regular file — e.g. a concurrently deleted file or a broken symlink — instead of skipping it and returning the rest.

**5. `LanceTableWriter::Close()` hard-errors on zero new fragments** — [`cpp/src/format/lance/lance_table_writer.cpp:107`](https://github.com/milvus-io/milvus-storage/blob/028347e0930050d25562f14407695817bcccc609/cpp/src/format/lance/lance_table_writer.cpp#L107)

`Close()` now returns an error when no new fragments were written (previously it returned a valid `ColumnGroupFile`), and also errors after a successful multi-fragment commit. Scoped to test-only code (`#ifdef BUILD_GTEST`), so low impact, but the empty-write case is a behavior regression.

jiaqizho:
> ## Code review
> Traced the FFI path/lifetime plumbing end to end — path mappers vs. `FileSystemProxy`'s `SubTreeFileSystem` rooting, `FileInfoListGuard` vs. `loon_filesystem_free_file_info_list` re-entry, `ArrowCDataReleaseGuard` arm/disarm against each Rust `from_raw` ordering, `ReaderHandle` drop paths, and AIMD option clamping in `create_file_system_config` — and those all look correct. Five findings below.
> 
> **1. New FFI symbol missing from the export maps** — [`cpp/ffi_exports.map:174`](https://github.com/milvus-io/milvus-storage/blob/028347e0930050d25562f14407695817bcccc609/cpp/ffi_exports.map#L174)
> 
> `loon_filesystem_get_object_info` is declared `FFI_EXPORT` in the public header but was never added to either export map. `local: *` hides it in `WITH_PYTHON_BINDING` builds, so the symbol won't resolve there.
> 
> **2. Status code lost when recovering an embedded error marker** — [`cpp/src/format/bridge/rust/src/bridge_util.cpp:133`](https://github.com/milvus-io/milvus-storage/blob/028347e0930050d25562f14407695817bcccc609/cpp/src/format/bridge/rust/src/bridge_util.cpp#L133)
> 
> A `NotImplemented` recovered from an embedded `__LOON_FFI_ERRCODE__=9` marker is rebuilt as `IOError` by `MakeIOErrorWithContext`, so the status code is dropped on the record-batch-reader translation path. ENOENT/EACCES survive only via the errno detail; callers that switch on `StatusCode` will misclassify.
> 
> **3. `shared_scan_scheduler` ignores the `filesystem` argument** — [`cpp/src/format/bridge/rust/src/lance_object_store.rs:404`](https://github.com/milvus-io/milvus-storage/blob/028347e0930050d25562f14407695817bcccc609/cpp/src/format/bridge/rust/src/lance_object_store.rs#L404)
> 
> The scheduler is keyed only on `fs_cache_key`, so the `filesystem` passed to later `BlockingDataset::Open` calls is silently discarded. After a 16-entry `FilesystemCache` eviction, fragment I/O keeps flowing through the evicted instance (which the scheduler retains forever) and bypasses the live filesystem's metrics.
> 
> **4. `list_dir` fails wholesale on one non-regular entry** — [`cpp/src/ffi/filesystem_c.cpp:988`](https://github.com/milvus-io/milvus-storage/blob/028347e0930050d25562f14407695817bcccc609/cpp/src/ffi/filesystem_c.cpp#L988)
> 
> The new size/type guard errors out the entire listing when a single entry isn't a regular file — e.g. a concurrently deleted file or a broken symlink — instead of skipping it and returning the rest.
> 
> **5. `LanceTableWriter::Close()` hard-errors on zero new fragments** — [`cpp/src/format/lance/lance_table_writer.cpp:107`](https://github.com/milvus-io/milvus-storage/blob/028347e0930050d25562f14407695817bcccc609/cpp/src/format/lance/lance_table_writer.cpp#L107)
> 
> `Close()` now returns an error when no new fragments were written (previously it returned a valid `ColumnGroupFile`), and also errors after a successful multi-fragment commit. Scoped to test-only code (`#ifdef BUILD_GTEST`), so low impact, but the empty-write case is a behavior regression.
> 
> 🤖 Generated with [Claude Code](https://claude.com/claude-code)

1,2 ACK 

3. Not applying. fs_cache_key is the logical filesystem identity, and same-key readers intentionally share one filesystem, ObjectStore, scheduler, and credential domain. Keying by instance identity would split that domain after cache churn.
4. Not applying. list_dir should not silently return partial results when it encounters an unsupported or unknown entry; failing explicitly is safer than hiding entries from object-store and version resolution.
5. Not applying. LanceTableWriter is compiled only under BUILD_GTEST, and this test-only write path supports a non-empty, single-fragment write. Zero- or multi-fragment outputs are outside its supported contract.

PR #652, opened 2026-08-27
enhance: add FlatObjectStorage primitives for object-key-native filesystem ops

somfornot:
## What

Adds a `FlatObjectStorage` extension interface to the filesystem layer, exposing three object-key-native primitives that bypass Arrow's directory emulation:

- **`ListObjectsByPrefix`** — raw object-key prefix listing (S3 `ListObjectsV2` with `Prefix=` and no delimiter, paginated). Cost is proportional to the number of matching keys, and it matches raw *mid-segment* prefixes that a directory-scoped `GetFileInfo(FileSelector)` cannot.
- **`DeleteObject`** — a single idempotent `DeleteObject`, with no preceding `HeadObject` and no parent directory-marker maintenance (unlike `DeleteFile`, which does `HeadObject` → `DeleteObject` → a `PutObject` that recreates the parent `dir/` marker).
- **`ObjectExists`** — exact-key `HeadObject` semantics: a key that only has descendants (or a zero-byte `key/` marker) is reported as **absent**, not present.

`S3FileSystem` implements the interface natively. `FileSystemProxy` delegates to its wrapped base filesystem (prepending/stripping the subtree base) and reports `NotImplemented` when the base cannot serve it. Free helpers dispatch to the native implementation when present and otherwise fall back to Arrow's directory semantics, so callers stay backend-agnostic on non-object-store filesystems (local, Azure).

A second, independent commit **fixes sub-second S3 request timeouts**: `CreateS3Options` derived `request_timeout` (a `double`, in seconds) from the int64 `request_timeout_ms` via integer division, truncating every sub-second component (1500ms → 1s; anything below 1000ms → 0, which the client builder then silently drops in favor of the SDK default). Dividing by `1000.0` preserves the fractional value; the consumer already applies `ceil(request_timeout * 1000)` to recover the exact millisecond value. Applied in both the S3 and GCP producers.

## Why

Callers that treat the store as a flat key/value namespace — notably Milvus's `ChunkManager` (`ListWithPrefix` / `Remove` / `Exist`) — get wrong or costly behavior from Arrow's directory emulation on a real object store: a directory-scoped listing misses raw mid-segment prefix matches, delete is non-idempotent and rewrites marker objects, and existence returns `true` for a path that only has descendants.

These primitives are the milvus-storage half of the fix for review feedback on the ArrowFileSystemChunkManager work in milvus-io/milvus#52805; the Milvus-side consumer changes will follow once this lands.

## Tests

- New `flat_object_storage_test.cpp`: **4 local** (free-helper fallback + `FileSystemProxy` `NotImplemented` delegation over a local base) + **3 remote/minio** (native S3 raw-prefix listing incl. the mid-segment vs. boundary-prefix distinction, exact-key existence, idempotent delete). All 7 pass against minio.
- Regression: full S3 + filesystem suites (`S3UnitTest`, `S3ClientTest`, `S3ProviderTest`, `ExternalFilesystemTest`, `LocalFsTest`, `CloudFsTest`) — **128 passed, 1 skipped** (`CloudFsTest.ConditionalWriteAzure`, no Azure endpoint), **0 failed**. `S3UnitTest.TestCreateS3Options` covers the timeout change.

DCO signed off on both commits.

### DIFF
--- a/cpp/include/milvus-storage/filesystem/flat_object_storage.h
+++ b/cpp/include/milvus-storage/filesystem/flat_object_storage.h
@@ -1,0 +1,96 @@
+// Copyright 2024 Zilliz
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include <memory>
+#include <string>
+#include <vector>
+
+#include <arrow/filesystem/filesystem.h>
+#include <arrow/result.h>
+#include <arrow/status.h>
+
+namespace milvus_storage {
+
+/// \brief Object-key-native operations for filesystems backed by a flat object
+/// store (AWS S3, S3-compatible services, and GCS via the S3 interoperability
+/// layer).
+///
+/// Arrow's FileSystem models a hierarchical namespace, which forces the S3
+/// backend to emulate directories:
+///   - GetFileInfo(FileSelector) enumerates a whole directory subtree;
+///   - DeleteFile is directory-aware (HeadObject, then DeleteObject, then a
+///     PutObject to recreate the parent "dir/" marker);
+///   - GetFileInfo(path) reports a key that only has descendants (or a "key/"
+///     marker) as a Directory rather than NotFound.
+///
+/// On a real object store those behaviours cost extra requests, mutate marker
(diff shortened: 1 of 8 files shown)


cpp/include/milvus-storage/filesystem/flat_object_storage.h
```diff
@@ -0,0 +1,96 @@
...
+///   - GetFileInfo(path) reports a key that only has descendants (or a "key/"
+///     marker) as a Directory rather than NotFound.
+///
+/// On a real object store those behaviours cost extra requests, mutate marker
+/// objects, and break exact-object semantics. A caller that treats the store as
+/// a flat key/value namespace (e.g. Milvus's ChunkManager) wants the raw
+/// equivalents. Filesystems that can serve them natively implement this
+/// interface; the free helpers below dispatch to it and otherwise fall back to
+/// Arrow's directory semantics, so callers stay backend-agnostic.
+class FlatObjectStorage {
```
jiaqizho, line 45: direct change the impl of `S3FileSystem::GetFileInfo` ?

PR #653, opened 2026-08-28
enhance: replace opendal with filesystem-backed storage

jiaqizho:
Iceberg planning used to build its own Rust cloud providers for S3, Azure, GCP, and OSS, which duplicated the filesystem clients and credential refresh logic already owned by C++. This change makes every production metadata, manifest, and positional-delete read reuse the cached filesystem selected by the caller.

PlanFiles now receives that filesystem together with credential-free reader identity options and exposes failures as Arrow results. The C++ bridge keeps the filesystem leased for the entire Rust planning operation, while the OpenDAL adapter forwards stat, range-read, and list calls through the filesystem FFI without constructing another cloud backend.

The Rust implementation is split into a generic, read-only filesystem-to-OpenDAL bridge and an Iceberg-specific StorageFactory and Storage adapter. The Iceberg layer binds each table to its scheme family and bucket, container, or local root, validates every absolute URI before converting it into a relative filesystem key, and rejects mutations instead of falling back to a native provider.

The old AWS AssumeRole, Azure broker SAS, GCP impersonation, Aliyun role factories, and Iceberg-side provider cache are removed. Native table creation remains separate for benchmarks and tooling and follows the same writer policy as Lance: static AK/SK or default IAM is supported, while role, broker, impersonation, missing credential, and unknown provider configurations return NotImplemented.

URI conversion continues to preserve the existing Milvus external-table contract, including Azure endpoint-bearing paths and delete metadata. CRT asynchronous reads also transfer reader ownership back to the Tokio task, or defer cleanup when that task is cancelled, so reader close and destruction never happen inside the CRT completion callback.

External-table exploration, benchmarks, and tools now pass the resolved filesystem and use the updated result-based bridge APIs. Obsolete exception baseline entries and direct provider dependencies are removed together with the old read path.


cpp/src/format/bridge/rust/src/iceberg_bridgeimpl.rs
```diff
@@ -298,7 +159,7 @@ pub(crate) fn denormalize_uri(uri: &str) -> String {
     let authority_start = scheme_end + 3;
     let rest = &uri[authority_start..];
     match &uri[..scheme_end] {
-        "abfss" | "abfs" => {
+        "azure" | "abfs" | "abfss" | "wasb" | "wasbs" => {
```
buqian-zilliz, line 56: **[P1] Canonicalize the newly accepted Azure aliases in the delete reader too**

This branch now strips the recorded account endpoint from `azure`, `wasb`, and `wasbs` data/delete paths before returning them to C++, but `ReadPositionalDeleteFile` still canonicalizes the `file_path` values stored inside a positional-delete Parquet file with `StripAbfssEndpoint`, which only handles `abfs` and `abfss`. For a table recorded with one of the newly accepted aliases, the row value remains, for example, `wasbs://container@account.../data.parquet` while `normalized_data_uri` is `wasbs://container/data.parquet`; none of the comparisons match, so the delete positions are silently ignored and deleted rows can be returned. Please canonicalize the same alias set on the C++ comparison path, ideally through one shared URI-normalization contract.

cpp/src/format/bridge/rust/src/filesystem_opendal.rs
```diff
@@ -0,0 +1,1013 @@
...
+        let info = self.object_info(path).await?;
+        Ok(opendal::raw::RpStat::new(info.into_metadata()?))
+    }
+
+    async fn read(
+        &self,
+        path: &str,
+        args: opendal::raw::OpRead,
+    ) -> opendal::Result<(opendal::raw::RpRead, Self::Reader)> {
+        let info = self.object_info(path).await?;
```
buqian-zilliz, line 792: **[P2] Avoid a metadata RPC before every filesystem-backed read**

This unconditional `object_info` call adds a remote `GetFileInfo` or `HEAD` request before every data read. Iceberg loads the table metadata, manifest list, and each manifest through `Storage::read`, so a snapshot with N manifests now incurs at least N+2 extra metadata RPCs compared with the native OpenDAL read path; repeated range reads also pay another stat because `FilesystemAccess` retains no size metadata. At production manifest counts this can roughly double the remote request count on the planning path. Please provide full-read and bounded-read FFI paths that do not require a preceding stat, and retain already-known sizes for range readers.

buqian-zilliz: This issue still exists: `FilesystemAccess::read` still performs an unconditional `object_info()` call before every filesystem-backed read, so production Iceberg planning continues to add one metadata RPC per metadata or manifest object and repeated range reads do not reuse a known size.

jiaqizho: The concern is valid in the general N-manifest case, but that is not the workload this path currently targets. In the current planning path, a snapshot has one data manifest, so a normal PlanFiles call reads the table metadata, one manifest list, and one manifest. The additional cost is therefore a fixed three metadata lookups per refresh, not a production-scale N-manifest fan-out. The repeated-range concern also does not apply to the normal no-delete planning path because these objects are each read once in full.

Fixing this correctly is broader than removing object_info(). Range reads would need to retain the existing C++ ReaderHandle so they can reuse the input stream content-length cache, and unknown-size full reads need new semantics while preserving the CRT asynchronous lifetime. Adding a metadata LRU would not remove the cold lookup for each distinct object.

Given the small bounded impact in the current workload and the scope of this provider-replacement PR, I am going to defer this optimization. We can revisit the persistent-reader path if this planning flow starts handling snapshots with many manifests.

cpp/src/format/iceberg/iceberg_format.cpp
```diff
@@ -23,19 +23,23 @@ namespace milvus_storage {
...
 
-  // Convert Milvus URI (scheme://address/bucket/path) to scheme://bucket/path.
-  // For S3 this is the final format; for Azure ABFSS, the Rust bridge further
-  // expands it to container@account.dfs.endpoint format that opendal requires.
-  ARROW_ASSIGN_OR_RAISE(auto parsed_uri, StorageUri::Parse(explore_dir));
-  ARROW_ASSIGN_OR_RAISE(auto iceberg_uri, StorageUri::Make(parsed_uri, false));
+  // Convert remote Milvus URI (scheme://address/bucket/path) to
+  // scheme://bucket/path. Local absolute paths are already in the form
+  // accepted by the filesystem-backed Iceberg adapter.
+  auto iceberg_uri = explore_dir;
```
buqian-zilliz, line 36: **[P1] Convert local planned paths back to filesystem-relative keys**

The new local branch plans from an absolute metadata location, but Rust returns each planned data-file and positional-delete path in that same absolute form. Because `ToMilvusUri` is a no-op for local storage, `create_reader` later passes the absolute path as `uri.key` to the same `FileSystemProxy`, which is already rooted at `fs.root_path`; `SubTreeFileSystem::PrependBase` removes the leading slash and prepends the root again, so a table under `/root/table` is opened as `/root/root/table`. Convert local planned paths to filesystem-relative keys before C++ I/O, while retaining the original Iceberg URI separately for positional-delete matching.

cpp/src/format/bridge/rust/src/iceberg_bridgeimpl.rs
buqian-zilliz, line 56: **[P2] Canonicalize accepted Azure schemes before denormalizing**

`parse_remote_uri` lowercases the scheme before validation, so valid mixed- or upper-case Azure URIs such as `ABFSS://container@account.dfs.../data.parquet` are accepted and read through the bound filesystem. `denormalize_uri` then matches the original scheme case-sensitively and leaves the `@endpoint` authority intact; when C++ reattaches the Milvus address, `ToMilvusUri` interprets the account host as the bucket, and the case-sensitive C++ delete-path normalization can also miss positional deletes. Canonicalize the scheme once, or use the same case-insensitive Azure-family check at every conversion boundary.

cpp/include/milvus-storage/format/iceberg/iceberg_common.h
```diff
@@ -19,13 +19,19 @@
...
 
+#include <arrow/result.h>
+
 #include "milvus-storage/filesystem/fs.h"
 
 namespace milvus_storage::iceberg {
 
-/// Convert ArrowFileSystemConfig to Iceberg storage options.
-/// @throws std::runtime_error for unsupported providers (Tencent, Huawei)
-std::unordered_map<std::string, std::string> ToStorageOptions(const ArrowFileSystemConfig& config);
```
buqian-zilliz, line 28: **[P2] Preserve the installed ToStorageOptions API**

This header is installed from the public `include/milvus-storage` tree, and this patch removes both the `ToStorageOptions` declaration and its shared-library definition. Downstream clients compiled against the previous release will either fail to compile after updating the headers or fail to resolve the old symbol at runtime; no compatibility shim or versioned API break is provided. Keep a deprecated transition entry point with explicit reader/writer semantics, or stage this as a versioned API break.

cpp/src/format/iceberg/iceberg_format.cpp
jiaqizho: Local Iceberg data access is not a supported production configuration. The production external-table flow does not support, and is not intended to support, reading actual Iceberg data from local storage; this local path exists only for test utilities and integration tests, which deliberately use `root_path="/"` for the absolute paths recorded in Iceberg metadata. This limitation also predates the provider-replacement change, so we will not add production path-rewriting semantics in this PR. I added a `FIXME(jiaqizho)` at the Rust handoff documenting that local data/delete paths must be normalized before production local data access could ever be enabled.

cpp/src/format/bridge/rust/src/iceberg_bridgeimpl.rs
jiaqizho: The supported Iceberg/OpenDAL URI schemes in this integration are the canonical lowercase spellings. `parse_remote_uri` lowercases only for scheme-family validation; it is not an API promise that this bridge rewrites URIs recorded in user Iceberg metadata. We should preserve that input rather than silently mutate it during denormalization. We therefore will not add mixed- or uppercase-scheme rewriting here; callers must use the lowercase schemes required by iceberg-rust/OpenDAL. If mixed-case support is desired, it should be normalized consistently upstream in iceberg-rust/OpenDAL rather than only at this bridge boundary.

somfornot (commented):
Reviewed the full diff (Rust bridge, C++ iceberg layer, FFI boundary, and the removed provider modules); every inline comment below was verified against the PR head and the merge-base before posting.

**Overall:** the layering is sound — the generic read-only filesystem→OpenDAL bridge vs. the Iceberg-specific URI binding is clean, single-bucket binding / mutation rejection / no-credentials-in-Rust are all documented and tested, and the FFI contract (signatures, ownership, null checks, error recovery via `ExtendStatusDetail`) checks out end to end. The two main concerns are one behavior regression around percent-encoding and a large amount of copied unsafe FFI plumbing that has already started to drift — see inline comments.

Also verified and deliberately **not** flagged: the CRT-callback deferred-drop pattern (runtime is a never-shutdown static; the shutdown race is unreachable), single-bucket binding (documented, tested tenant-isolation design), empty-`bucket_name` configs (unreachable through `resolve_config`), and OSS `use_iam` (opendal's reqsign sweeps the same env vars the deleted factory did). Two pre-existing latent issues outside this PR's hunks may deserve follow-ups: unsigned wraparound in `record_count - num_deleted_rows` (`iceberg_format.cpp:51`) when a malformed manifest omits `record_count`, and unguarded `std::stoll` on `PROPERTY_ICEBERG_SNAPSHOT_ID` (`iceberg_format.cpp:31`).

cpp/src/format/bridge/rust/src/iceberg_opendal.rs
```diff
@@ -0,0 +1,802 @@
...
+        if raw_authority.contains('@') {
+            return Err(invalid(
+                "non-Azure Iceberg URI must not contain user information",
+            ));
+        }
+        raw_authority
+    };
+
+    validate_raw_uri_path(uri)?;
+    let path = object_store::path::Path::from_url_path(url.path())
```
somfornot, line 202: **[correctness, medium] Percent-decoding regression for URL-encoded object keys**

`object_store::path::Path::from_url_path` percent-decodes (`%3A` → `:`), whereas the deleted `iceberg-storage-opendal` path computed the relative key by raw string slicing (`&path[prefix.len()..]`) with no decoding. Java/Spark Iceberg writers URL-escape partition values, and the escaped string is both the physical key and the manifest `file_path`.

So a table with e.g. `.../ts_hour=2024-01-01T00%3A00/delete.parquet` breaks: `count_positional_deletes` → `Storage::reader` → `relative_path` fetches the *decoded* key and gets NotFound, failing planning for a previously readable table.

Note the C++ side (`StorageUri::Parse` via `arrow::util::Uri`) does **not** decode, so Rust planning reads and C++ data reads now disagree about the same URI within this PR's own path.

cpp/src/format/bridge/rust/src/filesystem_opendal.rs
```diff
@@ -0,0 +1,1018 @@
...
+/// undifferentiated bridge failure.
+fn into_opendal_error(
+    error: LoonFFIError,
+    operation: &'static str,
+    path: &str,
+) -> opendal::Error {
+    let code = error.err_code;
+    let kind = if is_not_found_error_code(code) {
+        opendal::ErrorKind::NotFound
+    } else if code == unsafe { loon_errcode_aws_access_denied } {
```
somfornot: **[correctness, low] Missing `loon_errcode_permission_denied` mapping**

`into_opendal_error` maps only `loon_errcode_aws_access_denied` to `PermissionDenied`; the Lance twin `into_object_store_error` (`filesystem_object_store.rs:377`) maps both permission codes. C++ does emit the generic one — `FFIErrorCodeFromExtendStatus` converts EACCES/EPERM → `LOON_PERMISSION_DENIED` — so a local/non-AWS permission failure surfaces as `ErrorKind::Unexpected` on the Iceberg path but `PermissionDenied` on the Lance path. One-line fix mirroring the object_store mapping.

cpp/src/format/bridge/rust/src/filesystem_opendal.rs
```diff
@@ -0,0 +1,1018 @@
...
+// -----------------------------------------------------------------------------
+
+#[cfg(feature = "s3-crt-async")]
+/// State whose ownership is transferred to the C callback.
+///
+/// Keeping the reader and buffer in the same box guarantees that the native
+/// reader and the buffer allocation outlive the asynchronous write. A successful
+/// submission transfers this box to exactly one callback; an immediate submit
+/// failure reclaims it synchronously in `read_async_via_ffi`.
+struct FilesystemAsyncReadCallbackState {
```
somfornot, line 174: **[reuse] ~450 lines of unsafe FFI core copied from the Lance bridge**

The CRT async-read callback state machine, `is_not_found_error_code` (byte-identical), `FileInfoListGuard`, and the stat/open-reader/readat/list FFI sequences duplicate `filesystem_object_store.rs` nearly verbatim, differing only in the error type. Drift has already happened twice within this PR:

1. the permission-denied mapping gap (see comment at line 74);
2. `list_entries` lacks the ENOTDIR/`loon_errcode_arrow` fallback probe that `list_impl` has (`filesystem_object_store.rs:562-586`), so list-on-a-file answers differently per adapter.

This is exactly the code where a memory-safety fix applied to one copy and not the other is worst. Both files already import from `filesystem_c.rs`; suggest extracting the FFI cores there returning `Result<_, LoonFFIError>` with thin per-adapter error conversion.

cpp/src/format/iceberg/iceberg_common.cpp
```diff
@@ -43,13 +68,19 @@ std::string ConvertDeleteMetadataPaths(const std::vector<uint8_t>& json_bytes, c
...
   }
   return folly::toJson(parsed);
 }
 
-std::unordered_map<std::string, std::string> ToStorageOptions(const ArrowFileSystemConfig& config) {
+// This path is used only by CreateTestTable and related tests/benchmarks.
+// Production planning reads through the bound C++ filesystem instead. Keep the
+// native writer policy aligned with Lance: allow static AK/SK or default IAM,
+// and reject provider-specific role, broker, or impersonation flows.
+arrow::Result<std::unordered_map<std::string, std::string>> ToWriterOptions(const ArrowFileSystemConfig& config) {
```
somfornot, line 83: **[reuse] Writer credential policy duplicated from Lance with nothing enforcing alignment**

`ToWriterOptions` duplicates `lance_common.cpp:43-113`'s skeleton — char-identical `set_endpoint` lambda, identical `use_ak_sk`/`use_supported_credentials` computation and per-provider NotImplemented gates — and the comment above promises to "keep the native writer policy aligned with Lance", which nothing enforces. When the policy changes (e.g. allowing AssumeRole writers), one format will accept a config the other rejects.

`ToReaderOptions` (line 25) likewise shares the `milvus_fs_*` emission + root normalization with Lance's — a cross-language contract with the Rust bridge where drift means the two formats bind different filesystem identities from the same config.

Suggest a shared credential-policy helper plus a shared `milvus_fs` identity emitter, keeping only per-format key mapping (full dedup of the Azure branches isn't possible — endpoints differ — but the policy skeleton is).

cpp/src/format/bridge/rust/src/iceberg_opendal.rs
```diff
@@ -0,0 +1,802 @@
...
+    bucket: Option<String>,
+}
+
+impl FilesystemReadOptions {
+    /// Parse the complete option map and reject every unconsumed key.
+    ///
+    /// Strict consumption prevents a future caller from accidentally restoring
+    /// the old Rust credential path. Error messages name only the key, never its
+    /// value, so a rejected credential cannot be logged.
+    pub(crate) fn parse(mut options: HashMap<String, String>) -> Result<Self> {
```
somfornot, line 67: **[reuse] Security-relevant path validation implemented twice**

`FilesystemReadOptions::parse` re-implements `FFIReadOptions::parse` (`lance_object_store.rs:101-159`) for the same `milvus_fs_is_local`/`milvus_fs_root_path` keys, and `local_relative_path` (line 278) re-implements `FFIPathMapper::from_local_path` (`filesystem_object_store.rs:225-270`): ParentDir rejection, `strip_prefix` root, separator normalization.

The local-root escape validation now has two independent implementations with already-different semantics (iceberg rejects `CurDir|ParentDir` in root, Lance only `ParentDir`; Lance requires absolute input paths, iceberg accepts relative). Hardening one copy (symlink/UNC handling) leaves the other escapable. Suggest a shared identity parser + local-root mapper with a neutral error type, layering the Iceberg `file://` URI handling on top.

cpp/src/format/bridge/rust/src/iceberg_opendal.rs
```diff
@@ -0,0 +1,802 @@
...
+/// bound from S3 may reference `s3a`, for example, but may never switch to GCS.
+#[derive(Clone, Debug, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+pub(crate) enum SchemeFamily {
+    S3,
+    Gcs,
+    Azure,
+    Oss,
+}
+
+fn scheme_family(scheme: &str) -> Option<SchemeFamily> {
```
somfornot, line 128: **[reuse] Scheme-alias lists now live in three places**

`scheme_family` here (s3|s3a, gs|gcs, azure|abfs|abfss|wasb|wasbs, oss), `normalize_uri`/`denormalize_uri` in `iceberg_bridgeimpl.rs` (whose `normalize_uri` azure arm still misses `wasb`/`wasbs`), and Lance's `FFIPathMapper::Remote` (exact-match, no aliases).

This PR itself had to patch `denormalize_uri`'s drifted alias list (`abfss|abfs` → `azure|abfs|abfss|wasb|wasbs`) — evidence the split already bites. The next alias addition must touch three lists; missing one forks behavior between the two bridges. One shared scheme-family module should own alias grouping.

cpp/src/format/bridge/rust/src/filesystem_opendal.rs
```diff
@@ -0,0 +1,1018 @@
...
+            .with_context("path", path)
+            .set_source(source)
+        })?;
+
+        let filesystem = self.filesystem.clone();
+        let ffi_path = path.to_string();
+        let error_path = ffi_path.clone();
+        // Opening an Arrow RandomAccessFile is synchronous even when the
+        // concrete provider later supports CRT's asynchronous ReadAt.
+        let reader = TOKIO_RT
```
somfornot: **[efficiency] Non-CRT reads pay two sequential `spawn_blocking` round trips**

One blocking task opens the reader + checks `reader_supports_async` (509-564), then a second runs the blocking `ReadAt` (577-602) — on every local/MinIO/GCS/Azure read. One closure can do open → check → ReadAt, returning early with the `ReaderHandle` only when the CRT async path applies; all inputs are already moved into the closure. (The TODO below covers only the pre-read stat, not this.)

cpp/src/format/bridge/rust/src/filesystem_opendal.rs
```diff
@@ -0,0 +1,1018 @@
...
+                .set_source(source)
+            })?
+    }
+}
+
+/// Raw OpenDAL surface exposed by the bridge. Mutation methods are not
+/// overridden, so OpenDAL's `Access` defaults return Unsupported. The `()`
+/// associated types and false capability flags reflect that same contract.
+impl opendal::raw::Access for FilesystemAccess {
+    type Reader = opendal::Buffer;
```
somfornot, line 761: **[efficiency, latent] `type Reader = opendal::Buffer` repeats stat+open+close per ranged read**

Each `Reader::read(range)` via `Storage::reader` re-drives the full `Access::read` cycle (stat + open reader + readat + close) on the same file; the TODO at ~757 acknowledges only the single pre-read stat, not this per-range amplification.

Currently unreachable in production (iceberg-rust 0.9's planning path reads metadata/manifest-list/manifests as whole-file `Storage::read` calls; puffin/parquet `FileRead` consumers aren't exercised by the bridge), but the first feature reading in ranges pays N×(HEAD+open+GET+close) per file. A stateful `oio::Read` that opens and stats once would close the trap — worth a note next to the existing TODO at minimum.

cpp/src/format/iceberg/iceberg_common.cpp
```diff
@@ -67,154 +98,93 @@ std::unordered_map<std::string, std::string> ToStorageOptions(const ArrowFileSys
...
-      // factory LRU, and reuses it while creating a fresh GCS operator for
-      // each Iceberg storage operation.
-      set("gcs.service-account", config.gcp_target_service_account);
-      if (config.load_frequency > 0) {
-        options["gcp_credential_refresh_secs"] = std::to_string(config.load_frequency);
+  }
+  if (provider == kCloudProviderAzure) {
+    // Broker-issued SAS credentials belong to the C++ filesystem read path and
+    // are not reconstructed by the native writer.
+    if (use_supported_credentials && !config.IsAzureCredentialBrokerEnabled()) {
```
somfornot, line 122: **[correctness, low] Azure native test writer loses user-assigned managed identity**

The `use_iam` branch no longer forwards `AZURE_CLIENT_ID` as `adls.client-id` (old `ToStorageOptions` did), and opendal 0.55's azdls builder never sweeps env vars. reqsign appends `client_id` to the IMDS token request to select a *user-assigned* managed identity, so VMs with only a user-assigned MI now fail auth on this path.

Narrow scope: test/benchmark-only, system-assigned MI still works, and federated workload identity never worked here (nothing ever set `federated_token_file`) — but worth either restoring the forward or documenting the narrowed support.

cpp/src/format/bridge/rust/Cargo.toml
```diff
@@ -40,67 +40,46 @@ vortex = { git = "https://github.com/zilliztech/vortex", branch = "v0.75.0-patch
...
 serde_json = "1"
 chrono = "0.4"
 uuid = { version = "1", features = ["v7"] }
 
-# AWS SDK (for Lance AssumeRole credential provider)
-aws-config = "1"
-aws-credential-types = "1"
+# Shared ObjectStore and path types used by the filesystem-backed Lance reader
+# adapter and the Iceberg URI binding.
 object_store = { version = "0.13", features = ["aws", "gcp"] }
```
somfornot, line 75: **[cleanup] Stale `aws`/`gcp` features**

Their only users (`aws_arn_provider.rs`, `gcp_impersonation.rs`, `cloud_provider_cache.rs`) are deleted by this PR; remaining usage is `object_store::path`, core traits, and test-only `InMemory` — none needs a feature. The declared surface contradicts the comment right above ("no native Rust cloud clients for reads"). `object_store = "0.13"` suffices; lance transitively enables its own features for the writer path, so build output is unchanged.

cpp/src/format/bridge/rust/src/iceberg_bridgeimpl.rs
```diff
@@ -17,28 +17,15 @@ use crate::TOKIO_RT;
...
-};
 use crate::iceberg_ffi::IcebergFileInfo;
-
-const CLOUD_PROVIDER_KEY: &str = "cloud_provider";
-
-static ICEBERG_FACTORY_CACHE: LazyLock<GlobalLruCache<Arc<dyn StorageFactory>>> =
-    LazyLock::new(|| GlobalLruCache::new(CACHE_CAPACITY));
+use crate::iceberg_opendal::build_filesystem_file_io;
 
 fn azdls_factory(scheme: &str) -> anyhow::Result<OpenDalStorageFactory> {
```
somfornot: **[cleanup] Test-writer-only factories left in the production planning module**

After this PR, `azdls_factory`, `upstream_opendal_factory`, `build_file_io`, and `normalize_uri` are called only from `iceberg_testutil.rs`'s `CreateTestTable` writer; the production plan path uses only `iceberg_opendal::build_filesystem_file_io` + `denormalize_uri`. Readers of this module will see `OpenDalStorageFactory::S3/Gcs/Oss/Azdls` construction and conclude production still builds native cloud providers — exactly the confusion the Cargo.toml comments try to dispel. Suggest moving the four functions (and their unit tests) into `iceberg_testutil.rs`; `denormalize_uri` stays.

cpp/src/format/bridge/rust/src/iceberg_opendal.rs
```diff
@@ -0,0 +1,802 @@
...
+// Credential-free reader options and URI identity binding
+// -----------------------------------------------------------------------------
+
+/// Filesystem identity passed by C++ for one planning call.
+///
+/// These options describe only the already-selected filesystem. Provider
+/// endpoints and credentials are intentionally invalid here because the Rust
+/// adapter must not create a second cloud client or credential lifecycle.
+#[derive(Clone, Debug)]
+pub(crate) struct FilesystemReadOptions {
```
somfornot: **[cleanup] Tri-state struct with two unreachable error branches**

`parse()` guarantees exactly one of `root_path`/`bucket` is `Some`, yet the sole caller `IcebergUriBinding::new` must handle the impossible `None` states with two `ok_or_else` branches (~242, ~249) that can never fire, and a future refactor can desynchronize `is_local` without the type system noticing. `IcebergUriBinding` already has the `Local(PathBuf) | Remote(String)` shape — having `parse()` return that enum deletes the flag and both dead branches.

cpp/src/format/bridge/rust/src/filesystem_opendal.rs
```diff
@@ -0,0 +1,1018 @@
...
+        if filesystem.is_null() {
+            return Err(opendal::Error::new(
+                opendal::ErrorKind::ConfigInvalid,
+                "filesystem-backed OpenDAL access requires a filesystem",
+            ));
+        }
+        Ok(Self::from_filesystem(filesystem))
+    }
+
+    #[cfg(test)]
```
somfornot: **[cleanup] `new_for_test` only bypasses the null check**

Its two users are tests over `SharedPtr::null()` that never touch FFI. The tests sit in a child module with `use super::*`, so they can call the private `from_filesystem(SharedPtr::null())` directly — this constructor can be deleted with zero behavior change while keeping the null-pointer proof that mutations are rejected before any FFI call.

cpp/src/format/iceberg/iceberg_format.cpp
```diff
@@ -23,19 +23,23 @@ namespace milvus_storage {
 
 arrow::Result<std::vector<api::ColumnGroupFile>> IcebergFormat::explore(const std::string& explore_dir,
                                                                         const api::Properties& properties) {
+  ARROW_ASSIGN_OR_RAISE(auto filesystem, FilesystemCache::getInstance().get(properties, explore_dir));
```
somfornot, line 26: **[efficiency, minor] Config resolved twice**

`FilesystemCache::get` already runs `resolve_config` internally (`fs.cpp:361`); the explicit `resolve_config` on the next line re-parses the storage URI and re-extracts all `extfs.*` properties to rebuild the config the cache lookup just computed. Pure computation so the cost is modest, but a `get` overload returning fs+config would remove the redundancy (and could also cover the identical pre-existing pattern in `lance_format.cpp`).

cpp/src/format/bridge/rust/src/iceberg_opendal.rs
```diff
@@ -0,0 +1,802 @@
...
+}
+
+/// Parse a remote Iceberg URI into its scheme family, logical authority, and
+/// relative object path.
+///
+/// The logical Azure authority is the container before `@`; any recorded
+/// account endpoint is metadata, not permission to escape the C++ filesystem.
+/// Non-Azure user information, queries, and fragments are rejected to keep path
+/// mapping unambiguous.
+fn parse_remote_uri(uri: &str) -> Result<(SchemeFamily, String, String)> {
```
somfornot, line 172: **[cleanup, careful] `parse_remote_uri` tokenizes the same URI four ways**

Manual scheme/authority slicing, `url::Url::parse`, `validate_raw_uri_path` (its own `"://"` scan that builds and discards an `object_store` `Path`), then `from_url_path(url.path())` recomputing the discarded path — per `exists`/`metadata`/`read` call and per planned data-file URI.

Caution if simplifying: the raw-string `validate_raw_uri_path` is what rejects `s3://bucket/table/../x` *before* WHATWG dot-segment normalization silently resolves it (the traversal tests depend on this), and the manual authority slicing preserves Azure `container@account` casing that `url.host_str()` would lowercase. Only the discarded `Path::parse` inside the validator is safely removable (e.g. have it return the validated path for the clean-path case).

cpp/src/format/bridge/rust/src/iceberg_opendal.rs
jiaqizho: Leaving this unchanged after checking the Arrow 17 implementation used here. `StorageUri::Parse` calls `arrow::util::Uri::path()`, and that method applies `UriUnescape` to every path segment. The C++ side therefore decodes `%3A` as well, so this PR does not introduce the claimed Rust/C++ disagreement. Supporting object keys whose physical names retain literal percent escapes would require a cross-layer raw-path contract and is outside this PR.

cpp/src/format/bridge/rust/src/filesystem_opendal.rs
jiaqizho: ACK. Fixed in `0e60bc4`. `into_opendal_error` now maps both `loon_errcode_permission_denied` and `loon_errcode_aws_access_denied` to `PermissionDenied`. A regression test covers the generic permission code.

cpp/src/format/bridge/rust/src/filesystem_opendal.rs
jiaqizho: ACK on the concrete issues called out here. Addressed the drift in `0e60bc4`: generic permission mapping and the Arrow/ENOTDIR list-on-file fallback now match the ObjectStore adapter. I am not moving the entire adapter implementation in this PR because their public error, metadata, and listing contracts differ. The raw FFI declarations, `LoonFFIError`, result checking, and `ReaderHandle` are already shared in `filesystem_c.rs`. I added a `TODO(jiaqizho)` to move backend-neutral stat/read/list and CRT lifecycle plumbing in a separate PR while keeping adapter-specific conversion at each boundary.

cpp/src/format/iceberg/iceberg_common.cpp
jiaqizho: Not extracting a shared writer policy in this PR. The Iceberg native writer is used only by `CreateTestTable`, tests, and benchmarks; production Iceberg planning reads through the bound C++ filesystem. The reader option sets are also not identical: Lance carries scheduler/cache settings, while Iceberg carries the bucket/root binding and rejects every extra key. Keeping these mappings explicit avoids coupling production format code to a test-only writer.

cpp/src/format/bridge/rust/src/iceberg_opendal.rs
jiaqizho: Not sharing this mapper in this PR. Production local external-table data reads are not supported; the existing FIXME documents that local planned data/delete paths remain test-only. The contracts also intentionally differ: Iceberg handles `file://` and relative paths, while Lance requires absolute dataset paths. There is no concrete escape bug here, and a neutral helper would need policy switches for those differences.

cpp/src/format/bridge/rust/src/iceberg_opendal.rs
jiaqizho: Keeping alias handling at the format boundaries. The production Iceberg binding accepts `s3/s3a`, `gs/gcs`, and all Azure aliases, while endpoint stripping covers `azure/abfs/abfss/wasb/wasbs`. The remaining normalization code is test-writer-only, and Lance uses exact storage identity by design. The supported production alias behavior is covered without introducing a shared module whose callers would still need different policies.

cpp/src/format/bridge/rust/src/filesystem_opendal.rs
jiaqizho: ACK. Fixed in `0e60bc4` for both adapters. Non-CRT open, capability detection, blocking `ReadAt`, and reader destruction now stay in one `spawn_blocking` task. An async-capable CRT reader returns its handle to the existing callback path.

cpp/src/format/bridge/rust/src/filesystem_opendal.rs
jiaqizho: ACK. Added an explicit `TODO(jiaqizho)` next to the read path. The current iceberg-rust 0.9 production planning flow performs whole-file reads, so the `Buffer` implementation remains for now. A stateful `oio::Read` should be introduced before enabling a ranged `FileRead` consumer.

cpp/src/format/iceberg/iceberg_common.cpp
jiaqizho: Not restoring this in the current PR. The affected path is the native `CreateTestTable` writer used by tests and benchmarks; production reads use the C++ filesystem. The test writer scope is intentionally limited to static AK/SK or default IAM, with provider-specific broker, role, and impersonation flows rejected. User-assigned managed identity support for the test writer can be handled separately if it becomes necessary.

cpp/src/format/bridge/rust/Cargo.toml
jiaqizho: Keeping these direct features. The Lance native writer remains in this crate and uses the `object_store` AWS/GCP integrations. Transitive feature unification currently makes the build output similar, but relying on Lance to enable features required by this crate makes the dependency contract fragile. The nearby comment is scoped to production reads, not every writer path.

cpp/src/format/bridge/rust/src/iceberg_bridgeimpl.rs
jiaqizho: ACK. Fixed in `0e60bc4`. `azdls_factory`, `upstream_opendal_factory`, `build_file_io`, `normalize_uri`, and their tests now live in `iceberg_testutil.rs`. The production planning module retains only the planning and URI denormalization pieces it uses.

cpp/src/format/bridge/rust/src/iceberg_opendal.rs
jiaqizho: ACK. Fixed in `0e60bc4`. `FilesystemReadOptions` is now `Local { root } | Remote { bucket }`, so parsing constructs only a valid identity and `IcebergUriBinding::new` no longer handles unreachable `Option` states.

cpp/src/format/bridge/rust/src/filesystem_opendal.rs
jiaqizho: ACK. Fixed in `0e60bc4`. `new_for_test` was removed, and the child test module calls the private `from_filesystem` constructor directly.

cpp/src/format/iceberg/iceberg_format.cpp
jiaqizho: Leaving this unchanged. We considered returning the resolved config from `FilesystemCache::get`, but that would mix cache lookup with exposing configuration and create an easy-to-misuse API. The same pre-existing pattern exists in Lance, and the duplicated work is only pure configuration parsing, so this PR does not expand the cache interface.

cpp/src/format/bridge/rust/src/iceberg_opendal.rs
jiaqizho: Leaving this unchanged. The different views preserve separate safety properties: raw path inspection rejects traversal before WHATWG normalization, manual authority parsing preserves Azure `container@account` casing, and `Url` handles query/fragment parsing. The safely removable parse is only a micro-optimization in security-sensitive code, so I am not changing it without a broader path-representation design.

PR #657, opened 2026-09-01
fix: emit per-file ranges in v2 ffi column group builder

congqixia:
BuildLoonColumnGroups was writing cumulative row offsets per file, but the packed reader intersects start/end indices against each file's own zero-based row groups. For a group spanning multiple files, every file after the first got an empty overlap and its rows were silently dropped without any error.

Write per-file ranges ([0, rowCounts[f])) to match the reader contract and the parquet writer's per-file bookkeeping, and add a regression test pinning total rows for multi-file groups.


jiaqizho:
change title `column group builder ` to `v2 ffi column group builder`.

congqixia:
> change title `column group builder ` to `v2 ffi column group builder`.

done

PR #658, opened 2026-09-01
fix: reuse Lance datasets across fragments

jiaqizho:
Format reader metadata is cached by ColumnGroupFile, and each Lance ColumnGroupFile identifies one fragment. A fragment URI key therefore only reuses repeated opens of the same fragment. Different fragments of one dataset miss independently, even though each metadata load opens the same dataset and decodes the same table manifest.

Key Lance metadata by base URI and treat the cached payload as dataset-level metadata. Keep the opened BlockingDataset in that payload and cache immutable fragment-specific schema, row groups, deletion state, and memory estimates by fragment ID. BlockingFragmentReader remains private to each reader so projection and mutable read state stay isolated.

Because FormatReaderMetadataCache is owned by the top-level MetadataCache, dataset reuse is scoped to one top-level reader. Fragments within that reader share one dataset snapshot, while independent readers open independent snapshots without process-global state.

### DIFF
--- a/cpp/include/milvus-storage/format/lance/lance_table_reader.h
+++ b/cpp/include/milvus-storage/format/lance/lance_table_reader.h
@@ -42,16 +42,18 @@ class LanceTableReader final : public FormatReader, public std::enable_shared_fr
                    const std::vector<std::string>& needed_columns = {});
 
   struct MetaTrait {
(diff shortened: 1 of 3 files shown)


jiaqizho:
/unhold

change to segment level

PR #659, opened 2026-09-01
doc: add CLAUDE.md agent guide

somfornot:
Adds a `CLAUDE.md` describing this repository for coding agents, modeled on the one in the Milvus repo, plus an `AGENTS.md` symlink so agents looking for either filename find the same guide.

Everything in it was derived from the current tree — build files, CI workflows and headers — rather than from prose already in `README.md`:

- **Layout and core subsystems** — what `ffi/`, `transaction/`, `manifest`, `segment/`, `packed/`, `format/` and `filesystem/` each own, and an index of the design docs under `docs/`.
- **Build** — the `cpp/Makefile` options are environment variables (`USE_ASAN`, `WITH_UT`, `USE_JNI`, `USE_PYTHON_BINDING`, …), not the `WITH_JNI=ON` style names the README's option table lists; also records that toggling one against a populated `build/` does not reconfigure, so `make clean` is required first.
- **Reusing the Milvus builder container** for C++ builds, to keep the toolchain aligned and share the Conan cache.
- **Test** — the C++ / Python / Java / Rust / `tests/` command matrix.
- **Error handling** — the two layers of `error_handling_ratchet.sh` (the committed baseline must match the tree exactly, *and* CI compares per-category totals against the base branch, so raising the baseline inside a PR cannot pass), and the `ExtendStatusCode` ↔ `LOON_*` taxonomy where transient-vs-permanent is a retry contract that must survive to the C ABI.
- **Verification gate** — G1–G4 adapted to this repo: audit the construction sites of a value rather than the transform that forwards it, and reach the failure modes through `WITH_FIU` / minio / azurite instead of relying on green happy-path runs.
- **Conventions** — clang-format 18, `LOG_STORAGE_*`, the three places a new property must be registered, DCO sign-off.

`.gitignore` previously listed `AGENTS.md`; that entry is dropped so the symlink can be tracked. `.claude/` stays ignored.

Not addressed here: the Build Options table in `README.md` still documents the old `WITH_JNI=ON` / `WITH_PYTHON_BINDING=ON` names. `CLAUDE.md` notes the discrepancy and points at the Makefile variables that actually take effect; fixing the README is left to a separate `doc:` PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,4 +35,3 @@ python/venv/
 python/env/
 .claude/
 .worktrees/
-AGENTS.md
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,1 @@
+CLAUDE.md
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,257 @@
+# Milvus Storage
+
+Columnar storage engine for Milvus, built on Apache Arrow. C++20 core (`cpp/`) with
+Python, Java/Scala (JNI) and Rust bindings layered on a stable C ABI.
+
+The C++ core is the only place logic lives. Every binding is a thin shim over
+`extern "C"` entry points in `cpp/src/ffi/` — when you change behavior, change it in
+the core and check whether the C ABI surface (`cpp/include/milvus-storage/ffi_*.h`)
+needs to follow.
+
+## Layout
+
+| Path | Contents |
+|------|----------|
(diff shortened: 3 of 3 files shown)


jiaqizho:
I have discussed similar issues in other communities before.

checkout: https://github.com/apache/cloudberry/pull/1740

maybe we should consider use `AGENTS.md.template`.

PR #668, opened 2026-09-09
feat: add host-controlled OpenTelemetry tracing for storage reads

somfornot:
Superseded by #684, which contains the identical commit `c5afcec` and forms native GitHub stack #685 with #683. This PR is retained as the archive of prior reviews and discussion; continue review on #684.

---

Storage reads cross reader, metadata-cache, filesystem and Rust async boundaries, making it difficult for a host request trace to account for the complete operation. This change adds host-controlled OpenTelemetry tracing and preserves parent context across those boundaries.

## Changes

- Add C++ provider configuration and scoped parent injection without installing a global provider. Trace reader, format, metadata-cache and file I/O operations, including asynchronous S3 requests, with span budgets and sanitized error classes. Expose generic RAII scopes with initial attributes, links and span kind, mutable attributes, explicit status completion, and parent construction from an OpenTelemetry SpanContext. C++ hosts must rebuild against the updated header.
- Observe returned errors, early returns and exception unwinding while preserving original business results and exceptions across synchronous, Folly and native asynchronous calls. Own deferred attributes and links until span creation, and provide initial attributes to the host sampler. Return configuration failures as Status without replacing the previous configuration; contain tracing attachment failures and keep metadata in-flight loads progressing.
- Propagate context through Folly, Arrow and Rust async work. Start lazy spans when work executes, preserve per-operation configuration, and keep shared readers and file handles independent of request parents.
- Reduce context checks, repeated span-start locking and Rust bridge handle allocations. Null-provider operations retain a lightweight disabled snapshot without span state, budgets, or I/O completion accounting. Remove duplicate Vortex context binding and task boxing while preserving Tokio scheduling, profiling labels and cancellation. Cover mixed traced/untraced cache callers, concurrent lazy starts, host sampler decisions and context restoration.
- Add tracing and complete-read benchmarks, a paired comparison runner, and design, integration and performance documentation.

## Rebase and stack validation (2026-09-17)

Rebased this layer onto `main` at `7f8e0b9`; the new head is `c5afcec`. Conflict resolution preserves the upstream Rust module relocation, Lance dataset version cache and current Talon fallback behavior. Talon tracing is the separate dependent layer #683, not part of this PR's incremental scope. The branch chain is `main <- feature/opentel <- feature/opentel-talon`. GitHub rejected native stack grouping because this existing PR comes from a fork; #683 targets an upstream mirror of this PR's head and must be retargeted to main after this PR merges.

At the combined Talon adapter head `5fbbe7a`, development-container validation with `WITH_TALON=ON` passed: 394 selected Storage/Talon tests (including all 40 tracing tests), 77 conditional skips, 83 C FFI tests, and 7 new Talon tracing integration tests. Tracked-source clang-format 18 and the unchanged error ratchet passed. The Folly ThreadedExecutor ABI case remains excluded. The validation and performance history below predates this rebase; benchmarks were not rerun.

## Previous validation

All builds, formatting and tests ran in the development container with the required Conan cache mount. Current C++ validation uses the existing ASAN/coverage build with leak detection disabled:

- Tracing suite: 40 passed. Coverage includes real Reader early failures, original synchronous and asynchronous exceptions, scope completion and context restoration, sampler-visible initial attributes, deferred attribute/link ownership, configuration failure, and metadata in-flight progress under persistent tracing fault injection.
- Selected reader, format, metadata-cache, executor, local-filesystem and filesystem-metrics regression suite: 308 passed, 69 conditionally skipped (377 total), separately from the tracing suite.
- C FFI: 83 passed, 0 failed.
- Built `milvus_test`, `Test_FFI`, `tracing_benchmark` and `telemetry_read_benchmark`.
- Changed C++ files passed clang-format 18 checks; `git diff --check` and the error-handling ratchet passed (21 throw sites, with no baseline or gate changes). Unwind observers record tracing errors without catching and replacing business exceptions.
- Earlier validation of the unchanged Rust adapter used a standalone probe linked to the actual C++/Folly bridge: parent restoration across Pending/Ready/unwind, CPU versus blocking-pool placement, profiling labels, cancellation and subsequent work passed. That probe was not rerun for this C++ change.
- The known Folly `ThreadedExecutor` dependency/compiler ABI mismatch test remains excluded; this is not a full-suite pass. Remote storage and deployed end-to-end/exporter behavior have not been validated. Performance measurements below are historical and were not rerun for the current scope and exception changes.

## Sample: actual local trace output

Captured on commit `80cccd0`, before the disabled-path follow-up, in the development container, using the existing ASAN/coverage test build and a standalone local harness. The harness writes 100 rows in each of Parquet and Vortex, then calls `reader->take_async({0, 3, 7}, 2)` twice on the same Reader with metadata caching enabled. All four reads verified the returned IDs `[0, 3, 7]`; the test passed.

The host injects a parent context, an `AlwaysOnSampler`, and an `InMemorySpanExporter`, with `io_spans=true` and `max_spans_per_operation=256`. The tree below is reconstructed from actual exported span IDs and parent IDs; durations and attributes are measured values. This is a formatted SDK span dump, not OTLP wire-format output or a network-exporter test.

Actual Parquet cold-metadata read (head spans and some attributes omitted for readability):

```text
storage.read                           54.633 ms
| storage.operation=take_async
| storage.io.reads=3
| storage.io.requested_bytes=5103
| storage.io.returned_bytes=5103
| storage.spans.dropped=0
|
+-- storage.read_task                  53.153 ms
|   +-- storage.metadata.lookup         0.078 ms  storage.cache=miss
|   +-- storage.metadata.load           3.922 ms
|   |   +-- storage.fs.read             0.742 ms
|   |       storage.backend=local, storage.offset=2346
|   |       storage.requested_bytes=611, storage.returned_bytes=611
|   +-- storage.format.read            48.185 ms  storage.format=parquet
|       +-- storage.fs.read             0.834 ms
|           storage.offset=4, storage.requested_bytes=2246, storage.returned_bytes=2246
|
+-- storage.read_task                  53.314 ms
|   +-- storage.metadata.lookup         0.035 ms  storage.cache=in_flight
|   +-- storage.metadata.wait           3.268 ms
|   +-- storage.format.read            48.550 ms  storage.format=parquet
|       +-- storage.fs.read             1.376 ms
|           storage.offset=4, storage.requested_bytes=2246, storage.returned_bytes=2246
|
+-- storage.assemble                    0.075 ms  storage.operation=build_take_table
```

The two read tasks overlap: one loads metadata while the other waits for the in-flight load. Three logical file reads return 5,103 bytes in total. The subsequent read reports `storage.cache=hit` in both tasks.

| Local operation | Root span duration | Exported spans | Logical file reads | Returned bytes |
| --- | ---: | ---: | ---: | ---: |
| Parquet, cold metadata | 54.633 ms | 16 | 3 | 5,103 |
| Parquet, warm metadata | 4.663 ms | 12 | 2 | 4,492 |
| Vortex, cold metadata | 9.791 ms | 11 | 1 | 10,956 |
| Vortex, warm metadata | 3.290 ms | 8 | 0 | 0 |

These are individual trace examples from an ASAN/coverage build, not performance comparisons or acceptance results. Cold/warm refers to metadata caching, not disk caching; the Vortex warm read's zero file reads is specific to this small-file case. Format spans include I/O wait, and overlapping child durations must not be added to infer end-to-end latency. A parent/child tree alone does not depict overlap; span start times and durations provide the timeline. This sample has one file-read child per format-read span. Separately, the current `ReadManyAsync` wrapper emits one aggregate logical span with `storage.range_count`, rather than an individual span for each range.

## Performance status — pending acceptance

The optimization comparison uses commit `6966d49` as its baseline, with matching Release builds and jemalloc. Null-provider root-plus-four-child allocations fall from 27 to 23, and Rust capture-plus-attach allocations from 8 to 6. Holding 50,000 captured Rust context handles reduces jemalloc allocated bytes from 3.2 MB to 1.6 MB; this does not represent whole-service RSS savings.

Six representative local read comparisons detected no statistically significant regression against that baseline, but three confidence intervals remain inconclusive. The 2% measurement sensitivity is not an accepted regression allowance. No-regression acceptance against the version before Telemetry was introduced is still outstanding.

The recorded performance comparison predates the upstream merge; performance acceptance also remains pending for the merged head.

The disabled-path follow-up also compares the C++ runtime sources from `80cccd0` with `4b45ee7` using matching Release tooling and dependencies. Each operation includes parent attachment (empty tracestate), one root and four synchronous children; the asynchronous root returns a ready SemiFuture. After 100 warmup iterations and 1,000 counted iterations:

| Null-provider runtime probe | C++ new calls per operation | Requested bytes per operation |
| --- | ---: | ---: |
| Parent attached, synchronous root | 33 → 31 | 2,232 → 1,992 |
| Parent attached, ready asynchronous root | 48 → 32 | 5,408 → 2,152 |
| No parent, synchronous root | 0 → 0 | 0 → 0 |
| No parent, ready asynchronous root | 1 → 1 | 160 → 160 |

These counts include attachment and business-Future allocations and have a different denominator from the historical probe above. They measure C++ runtime allocations only, not complete-read latency, Rust allocation totals or RSS. These measurements describe `4b45ee7`. The current implementation preserves original business exceptions and returns disabled futures without adding an exception-conversion continuation; allocation and latency measurements have not been rerun for this change. Folly still propagates RequestContext, so these results do not establish zero overhead. Full no-regression acceptance against the pre-tracing version remains pending.

See [Storage OpenTelemetry](docs/opentelemetry-storage.md) for the API contract, propagation boundaries, validation limits, benchmark methodology and optimization measurements.


cpp/benchmark/CMakeLists.txt
```diff
@@ -50,3 +50,29 @@ target_include_directories(benchmark PUBLIC
...
+endif()
+if (NOT APPLE)
+  target_link_options(tracing_benchmark PRIVATE "LINKER:--disable-new-dtags")
+endif()
+
+# Run the same end-to-end harness against pre-tracing and current libraries.
+add_executable(telemetry_read_benchmark benchmark_telemetry_read.cpp)
+target_link_libraries(telemetry_read_benchmark PRIVATE milvus-storage benchmark::benchmark
+  opentelemetry-cpp::opentelemetry_trace)
+target_compile_definitions(telemetry_read_benchmark PRIVATE STORAGE_TELEMETRY_BENCHMARK)
```
jiaqizho, line 71: Is this necessary?

cpp/include/milvus-storage/tracing.h
```diff
@@ -0,0 +1,55 @@
+// Copyright 2026 Zilliz
```
jiaqizho, line 1: move to the `cpp/include/milvus-storage/common`. And please setup right license.

Check other license pls.

cpp/src/filesystem/s3/s3_filesystem.cpp
```diff
@@ -822,32 +824,40 @@ class ObjectCrtInputFile final : public arrow::io::RandomAccessFile, public NonB
...
-        ctx->request, [ctx](const Aws::S3Crt::S3CrtClient*, const S3CrtModel::HeadObjectRequest&,
-                            const S3CrtModel::HeadObjectOutcome& outcome,
-                            const std::shared_ptr<const Aws::Client::AsyncCallerContext>&) mutable {
-          if (!outcome.IsSuccess()) {
-            if (outcome.GetError().GetResponseCode() == Aws::Http::HttpResponseCode::NOT_FOUND) {
-              ctx->future.MarkFinished(arrow::Result<int64_t>(PathNotFound(ctx->path)));
+    ctx->trace.Attribute("storage.backend", "s3-crt");
+    ctx->trace.Attribute("storage.operation", "head");
+    (void)tracing::Observe(ctx->future, ctx->trace);
+    try {
```
jiaqizho, line 830: why we need do `try...catch` here?  Was it introduced by OpenTelemetry?

cpp/include/milvus-storage/tracing.h
```diff
@@ -0,0 +1,55 @@
...
+
+struct TraceOptions {
+  bool io_spans = true;
+  // Includes the operation span. Excess children are aggregated on the operation.
+  uint32_t max_spans_per_operation = 256;
+};
+// Changes apply to subsequent operations; active operations retain their snapshot.
+void SetTraceOptions(const TraceOptions& options);
+
+class TraceScope {
```
jiaqizho, line 35: We do defined tracing.h and runtime.h.

In storage we used `runtime.h` without `tracing.h` ?

jiaqizho, line 35: How about we just provider a single usaged API(which like the defined in arrow):
```
#  define START_SPAN(target_span, ...)
#  define START_SCOPED_SPAN(target_span, ...)
#  define START_SCOPED_SPAN_SV(target_span, name)
#  define START_COMPUTE_SPAN(target_span, ...)
#  define ACTIVATE_SPAN(target_span)
#  define MARK_SPAN(target_span, status)
#  define EVENT(target_span, ...)
#  define EVENT_ON_CURRENT_SPAN(...)
#  define END_SPAN(target_span)
#  define END_SPAN_ON_FUTURE_COMPLETION(target_span, target_future)
#  define PROPAGATE_SPAN_TO_GENERATOR(generator)
#  define WRAP_ASYNC_GENERATOR(generator)
#  define WRAP_ASYNC_GENERATOR_WITH_CHILD_SPAN(generator, name)
```

The user side: 
```
START_SPAN(span, "A",
               ...);
// running
if (...) EVENT(span, "abort")

// running
END_SPAN(span);
```

cpp/src/tracing/runtime.h
```diff
@@ -0,0 +1,170 @@
...
+#include <mutex>
+#include <optional>
+#include <utility>
+#include <folly/io/async/Request.h>
+#include <folly/futures/Future.h>
+#include <arrow/status.h>
+#include <arrow/util/future.h>
+#include <opentelemetry/trace/tracer.h>
+
+namespace milvus_storage::tracing {
```
jiaqizho, line 16: I think we should consider which processes here require instrumentation. so as to avoid excessive tracking.

checkout arrow: 
```
class ARROW_ACERO_EXPORT TracedNode {

void TracedNode::NoteStartProducing(std::string extra_details) const {
  std::string node_kind(node_->kind_name());
  EVENT_ON_CURRENT_SPAN(node_kind + "::StartProducing", {{"node.details", extra_details},
                                                         {"node.label", node_->label()}});
}
}
```

Storage does not have an architecture similar to an executor. However, as I understand it, it can be broadly divided into three parts:
1. Cloud access, s3 read,write ... 
2. The interface between the format writer/reader and the file system.
3. The interface between the top-level writer/reader and the format writer/reader.

So, I don't think there's a need for such intrusive data tracking; placing tracking points at just a few select entry and exit points should suffice.

cpp/src/tracing/runtime.cpp
```diff
@@ -0,0 +1,267 @@
...
+  configuration = std::move(next);
+}
+OperationTrace::OperationTrace(
+    const char* name, bool lazy, bool io, ot::SpanContext link, const char* operation, const char* format) {
+  context_ = Capture();
+  if (!context_ || context_->disabled)
+    return;
+  // The cache leader already owns the physical metadata load. A format open
+  // below that leader must not create another span for the same work.
+  if (context_->operation && std::string_view(name) == "storage.metadata.load" &&
```
jiaqizho, line 164: Coupled business logic.

PR #670, opened 2026-09-09
feat: support Volcengine TOS object storage provider

wu-xingying:
Add Volcengine (火山引擎) TOS support alongside the existing Aliyun/Tencent/Huawei S3-compatible providers, using the OIDC WebIdentity -> STS AssumeRole credential flow. This covers per-role OIDC credentials, cross-account OIDC role chaining for external tables, and hardened STS expiry parsing.

Credential providers:
- VolcengineSTSCredentialsClient calls AssumeRoleWithOIDC against sts.volcengineapi.com (overridable via VOLCENGINE_OIDC_STS_ENDPOINT) and parses the temporary AK/SK/token/expiration.
- VolcengineSTSAssumeRoleWebIdentityCredentialsProvider reads VOLCENGINE_OIDC_ROLE_TRN / _TOKEN_FILE / _ROLE_SESSION_NAME from the environment, with lazy load and a 30-minute early refresh under a reader/writer lock. It is parameterized so an external-table role_arn can drive the AssumeRoleWithOIDC RoleTrn while the VKE-injected OIDC token file continues to come from VOLCENGINE_OIDC_TOKEN_FILE.
- Cross-account OIDC role chaining mirrors the Aliyun provider: (1) AssumeRoleWithOIDC into the machine-identity role (VOLCENGINE_OIDC_ROLE_TRN + VOLCENGINE_OIDC_TOKEN_FILE), then (2) sts:AssumeRole into the customer target role (extfs.role_arn), signed with Volcengine V4 (HMAC-SHA256) using the step-1 STS creds as caller. The chain degrades to step-1 credentials when the target role is empty or equals the machine role, so single-account role_arn deployments keep working.

STS expiry parsing:
- A ParseStsExpiry helper hardens expiry parsing against format drift: it normalizes a numeric timezone offset to UTC Z via NormalizeToUtcZ, falls back to the SDK's ISO-8601 parser for fractional-second or colon-less offset forms, and if both fail keeps the credentials while assigning a conservative 1h TTL to force an early refresh instead of a hard failure. The expiry only drives the local refresh clock and is never signed or sent upstream, so an unparseable value must not invalidate an otherwise good AK/SK/token.
- common/Util.h NormalizeToUtcZ normalizes RFC3339 timestamps carrying a numeric timezone offset (e.g. +08:00) to UTC "...Z" so the AWS SDK's ISO_8601 parser accepts them.

Provider wiring:
- Wire VOLCENGINE into the CloudProviderType enum/map, the S3 dispatch switch, the credential provider factory, and every "non-AWS S3-compatible" branch: force_virtual_addressing, MakeClientConfiguration (disable IMDS), checksum WHEN_REQUIRED, DeleteObjects Content-MD5 injection, and the fs.cloud_provider property enum.
- Route the Volcengine role_arn path through S3FileSystemProducer with fail-fast validation of the OIDC token file and warnings for the unsupported external_id / load_frequency options.
- Add a volcengine target to the Makefile and a volcengine provider entry to run_cloud_test.sh.

Tests:
- Golden-vector unit test locking the hand-rolled V4 signer.
- Provider tests covering env-only construction, parameterized construction, missing-token cases, empty STS responses, and argument precedence over the environment RoleTrn.
- End-to-end tests for the two-step chain, single-account short-circuit, and inner/outer failure semantics.
- Regression test for the unparseable-expiry fallback (bare Unix epoch), verifying credentials survive with a bounded future TTL.


cpp/src/common/Util.cpp
```diff
@@ -0,0 +1,51 @@
...
+#include <stdexcept>
+#include <string>
+#include <regex>
+#include <sstream>
+#include <ctime>
+#include <iomanip>
+
+namespace milvus_storage {
+std::string
+NormalizeToUtcZ(const std::string& s) {
```
jiaqizho, line 16: no need add a `Util.cpp` as a helper class. just move it into `cpp/src/filesystem/s3/provider/VolcengineSTSClient.cpp`

cpp/src/filesystem/s3/provider/VolcengineSTSClient.cpp
```diff
@@ -0,0 +1,436 @@
+#include "milvus-storage/filesystem/s3/provider/VolcengineSTSClient.h"
```
jiaqizho, line 1: Licence and check other files.

cpp/src/filesystem/s3/provider/VolcengineCredentialsProvider.cpp
```diff
@@ -0,0 +1,183 @@
...
+#include <limits.h>
+#include <mutex>
+#include <sstream>
+#include <random>
+#include <iostream>
+#include <fstream>
+
+namespace milvus_storage {
+
+static const char STS_ASSUME_ROLE_WEB_IDENTITY_LOG_TAG[] =
```
jiaqizho, line 26: Is `Volcengine` support condition write?

cpp/test/packed/run_cloud_test.sh
```diff
@@ -81,8 +81,20 @@ huawei() {
     export TEST_ENV_USE_SSL=true
 }
 
+volcengine() {
```
jiaqizho, line 84: no need change this file. Will remove this file later.

cpp/Makefile
```diff
@@ -2,7 +2,7 @@
 	python-lib java-lib package \
 	fix-format check-format fix-tidy check-tidy \
 	check-error-ratchet update-error-ratchet \
-	aws gcp azure aliyun tencent huawei
+	aws gcp azure aliyun tencent huawei volcengine
```
jiaqizho, line 5: also no need change this target and `test-cloud-storage` target.

Will remove this logical later

cpp/test/packed/run_cloud_test.sh
jiaqizho: And another queustion is, does `volcengine` have have a backend emulator similar to MinIO or Azurite?

jiaqizho:
Could you spin up an actual Volcengine service. And use the environment variables from `cpp/test/include/test_env.h` to fully verify that Volcengine works correctly with both AK/SK and IAM authentication methods?

Also, please post the test results in this PR. Thanks.

xiaofan-luan:
can we access TOS with S3 compatible or OSS compatible API

jiaqizho:
> can we access TOS with S3 compatible or OSS compatible API

The current PR is implemented: only a credential provider has been added, while the actual request path still goes through the S3 filesystem.

jiaqizho:
Hi @wu-xingying 

Are you planning to continue working on this PR? Storage will need to support TOS after October.  If this PR is no longer being updated, I’ll submit a new PR with a new TOS implementation and close this one.

That said, I’d prefer to move forward with merging this PR, provided you have time to address the outstanding review comments.

PR #675, opened 2026-09-11
fix: preserve binary encryption keys across storage interfaces

XuanYang-cn:
Binary AES keys can be truncated at embedded NUL bytes when crossing a string-based FFI. A 32-byte key truncated to 16 or 24 bytes is still a valid AES key, so writing can succeed while reading with the original key fails authentication.

Require standard padded Base64 for `writer.enc.key` for all properties-based writers, including native C++. Decode and validate a local copy in `convert_write_properties` when constructing the Parquet encryption configuration; `loon_writer_new` preserves the supplied properties. Reject noncanonical Base64 and keys other than 16, 24, or 32 bytes without including key material in errors. Invalid writer keys fail when write/close creates the Parquet writer, before opening the data file.

Define the callback registered by `loon_reader_set_keyretriever` to return Base64 and decode it in the C adapter. Native C++ `Reader::set_keyretriever` / `KeyRetriever::GetKey` continues to return raw binary bytes, as used by Milvus. The Python binding accepts raw `bytes` (or UTF-8 `str` for compatibility) and encodes inside its C callback. A null or invalid C callback key makes the actual read fail.

Function signatures and struct layouts remain unchanged. The writer property and C callback value contracts change, so callers must supply Base64 at those interfaces. This prevents new incorrectly encrypted files; it does not rewrite existing files encrypted with a truncated key.

Validation on the updated implementation:

- Encryption and metadata-reopen C++ tests: 6 passed; 4 existing non-Parquet parameter cases skipped.
- Real C FFI writes and reads cover 10 AES-length/NUL-offset combinations, including high bytes. An independent Parquet reader verifies exact values using the original binary key and rejects valid-length 16/24-byte truncated keys. Native C++ coverage verifies raw callback keys and unchanged Base64 properties.
- Six malformed writer inputs and six unavailable/malformed C callback variants fail as expected; callback execution is asserted with metadata caching disabled.
- Full C FFI suite: 83 passed, with the newly built storage library verified at runtime.
- Full clang-format 18 check, error-handling ratchet against `7cc50449245b2a924ca2fc18f48fd1b0bb3b1720`, and the full Python lint CI (Black, isort, flake8) passed after wrapping the property comment.
- clang-tidy 18 completed without errors using a temporary config that removes the obsolete `AnalyzeTemporaryDtors` option and scopes header diagnostics to this checkout. It reports nonfatal modernization/style suggestions, including C arrays and braces in the FFI tests; this is not a warning-free lint result.
- Standards review: no findings. Spec review: no findings.
- No new local Python binding runtime tests were added or run, as requested. The existing Python 3.11 CI suite passed; storage tests directly exercise the C callback path.
- Milvus local CMEK IT branch, commit `b4404c7a724319a9e37cdefa0c7bb99683e6c963`, pins storage `6c12ad2ce75070da2f937f8bfb4bb51b3d98de7a` and includes the matching Base64 producer. Its final C++/Go incremental build passed.
- Milvus race-enabled IT: all 4 NUL-offset cases passed through both an independent Parquet decoder and production `NewFFIPackedReader`, including exact values, EOF, and rejection of truncated keys. V2 scalar/vector/StructArray (3 cases) and V3 cold-read/scalar/vector/StructArray (4 cases) passed with the fixture DEK NUL offset forced to 24.
- Milvus race-enabled plaintext packed writer/reader and manifest regressions: 15 passed, including partial-column and multiple-batch reads.
- The Milvus dependency update also supplies NULL for the optional `file_properties` argument introduced by upstream #672. Fresh compilation of the packed test package exposed the missing argument; after adapting it, the final build and all Milvus regressions above were rerun. This interface change belongs to the PR base, not this encryption fix.

Local C++ validation uses cached Milvus Conan dependencies, Release/FIU/PIC enabled, and ASAN/Talon disabled. This revision ran the focused C++ encryption tests and the full C FFI suite; the standard storage dependency and service matrices remain covered by PR CI.


cpp/src/ffi/writer_c.cpp
```diff
@@ -39,6 +40,24 @@ LoonFFIResult loon_writer_new(const char* base_path,
       RETURN_ERROR(LOON_INVALID_PROPERTIES, "Failed to parse properties [", opt->c_str(), "]");
     }
 
+    if (GetValueNoError<bool>(properties_map, PROPERTY_WRITER_ENC_ENABLE)) {
```
jiaqizho: pls checkout cpp/src/format/parquet/parquet_writer.cpp:L71.

We should direct decode base64 on use side without change the `properties`. Semantically the `properties` is unmutable.

jiaqizho: Should we also change the `cpp/include/milvus-storage/format/parquet/key_retriever.h` 

Or did you wrap the logic inside the `key_retriever_callback_`?

PR #681, opened 2026-09-16
feat: extend storage JNI for connector reads and writes

santiago-wjq:
The Spark connector's Storage V3 read/write path needs filesystem access, owned manifests and independent row-take results that the Java binding does not expose.

- Expose filesystem metadata, byte/range I/O, manifest/transaction operations and shared column-group metadata accessors.
- Keep the metadata accessors in a separate JNI source and obtain V3 reader layouts from manifests. Preserve the existing Storage V2 interfaces for Milvus 2.6 backfill, including their original three-argument column-group builder.
- Preserve native error codes and pending Java exceptions, and release temporary JNI resources on failures.
- Give row-take results independent ownership and expose Arrow copy statistics.
- Extract native resources from their containing JAR.
- Configure shared AWS CRT dependencies and jemalloc TLS for loading into a running JVM.

The core library, JNI, tests and benchmarks use the base branch's original link settings, including PUBLIC dependency propagation. The additional Linux noexecstack option has also been removed; cpp/CMakeLists.txt matches the PR base.

Validation on Linux x86_64 / JDK 21 at `8ef20b0`, before restoring the original link settings:

- Java/Scala binding tests: 18 passed, 0 failed or canceled, including a V3 writer → transaction commit → manifest reopen → reader round trip and owned/borrowed lifetime checks.
- Existing V2 column-group builder tests: 6 passed.
- Recompiled all 13 JNI source files and the restored V2 builder; strict JNI linking and immediate native-library loading passed. The test process verified the resource paths and SHA-256 of the libraries actually loaded.
- C++ formatting and both layers of the error-handling check passed.

Local runtime validation reused the remaining native-core objects and dependencies from the prior build based on `cb22dfaa01175c429edbd0fb6794251f1a58f400` with the earlier JNI work. This is targeted validation, not a full rebuild of the current PR base; the complete build remains for CI. With the repository's Arrow Java 14.0.1 dependencies, `-Xcheck:jni` reports five warnings in Arrow's JNI loader while all functional tests pass.

For the link-setting rollback in `4308ecc` and removal of noexecstack in `5689301`, `git diff --check` passed. Native builds and runtime tests have not been rerun locally after these changes; CI validation is pending.

Scope: this PR updates the storage bindings for the V3 path. Removal of existing V2 call paths in the Spark connector repository is separate work.


java/src/main/java/io/milvus/storage/NativeLibraryLoader.java
```diff
@@ -1,241 +1,147 @@
...
-
-        if (os.contains("windows")) {
-            libExtension = "dll";
-            platform = "windows-" + (arch.contains("64") ? "x86_64" : "x86");
-        } else if (os.contains("mac")) {
-            libExtension = "dylib";
-            platform = "darwin-" + (arch.contains("aarch64") || arch.contains("arm64") ? "aarch64" : "x86_64");
+        if (os.contains("linux")) {
+            platform = "linux";
+        } else if (os.contains("mac") || os.contains("darwin")) {
```
jiaqizho, line 62: darwin?

java/src/main/java/io/milvus/storage/NativeLibraryLoader.java
```diff
@@ -1,241 +1,147 @@
...
-            return;
-        } catch (Exception e) {
-            System.out.println("Could not load from resources: " + e.getMessage());
-        }
-
-        try {
-            // Fall back to system library path
+        String platform = currentPlatform();
+        String prefix = "native/" + platform + "/";
+        String libraryName = System.mapLibraryName(JNI_LIBRARY_NAME);
```
jiaqizho, line 34: why we need platform details for load libs?

jiaqizho (commented):
Could u

cpp/include/milvus-storage/ffi_internal/v2_column_groups_builder.h
```diff
@@ -37,13 +37,17 @@ namespace milvus_storage {
...
+//   - an empty format
 //   - mismatched outer lengths between the three vectors
 //   - zero column groups
 //   - a group with empty columns or empty files
 //   - a group where rowCounts.size() != files.size()
+//   - a negative per-file row count
 LoonColumnGroups* BuildLoonColumnGroups(const std::vector<std::vector<std::string>>& columns_per_group,
                                         const std::vector<std::vector<std::string>>& files_per_group,
-                                        const std::vector<std::vector<int64_t>>& row_counts_per_group);
+                                        const std::vector<std::vector<int64_t>>& row_counts_per_group,
```
jiaqizho: Storage v2 no need `format`. It have not manifest and `format` options.

cpp/src/jni/v2_column_groups_jni.cpp
```diff
@@ -23,37 +23,49 @@
...
 
-namespace {
+using namespace milvus_storage::jni;
 
-void ThrowJava(JNIEnv* env, const char* cls_name, const std::string& msg) {
-  jclass exc = env->FindClass(cls_name);
-  if (exc != nullptr) {
-    env->ThrowNew(exc, msg.c_str());
+namespace {
+const LoonColumnGroup* Group(JNIEnv* env, jlong pointer, jint index) {
```
jiaqizho: I don't think Spark should continue down the "Storage V2" path. The V2 interfaces there were originally introduced specifically to facilitate backfilling in Milvus 2.6.

cpp/CMakeLists.txt
```diff
@@ -163,10 +163,29 @@ if (WITH_FIU)
 endif()
 
 
-target_link_libraries(milvus-storage PUBLIC ${LINK_LIBS})
+# Keep implementation dependencies out of consumers such as JNI. In particular,
+# linking the static gRPC closure again registers its process-wide flags twice.
+# Arrow, Folly and fmt are part of the public C++ header/API requirements.
+target_link_libraries(milvus-storage
+  PRIVATE ${LINK_LIBS}
```
jiaqizho: private link ?

cpp/src/jni/filesystem_jni.cpp
```diff
@@ -0,0 +1,254 @@
...
+                                                                                    jlong handle) {
+  GuardVoid(env, [&] { loon_filesystem_destroy(static_cast<FileSystemHandle>(handle)); });
+}
+
+JNIEXPORT jbyteArray JNICALL Java_io_milvus_storage_MilvusStorageFileSystemNative_readFileAll(JNIEnv* env,
+                                                                                              jobject,
+                                                                                              jlong handle,
+                                                                                              jstring path) {
+  return Guard<jbyteArray>(env, nullptr, [&]() -> jbyteArray {
+    ScopedUtf8 p(env, path);
```
jiaqizho, line 73: what for? why we need convert it into utf8?

cpp/src/jni/reader_jni.cpp
```diff
@@ -125,7 +201,10 @@ extern "C" LoonFFIResult loon_record_batch_reader_read_next(LoonRecordBatchReade
             if (!concat_result.ok()) {
               RETURN_ERROR(LOON_ARROW_ERROR, concat_result.status().ToString());
             }
-            fresh_cols.push_back(concat_result.ValueOrDie());
+            auto copied = concat_result.ValueOrDie();
+            holder->copies += 1;
+            holder->copied_bytes += BufferBytes(*copied);
```
jiaqizho, line 215: what for?

cpp/src/jni/v2_column_groups_jni.cpp
santiago-wjq: Agreed. In 8ef20b081bfe07d4811c3aae9907faf7aedc0014, I removed this PR's changes to the V2 column-group builder/JNI and packed writer, preserving the existing Milvus 2.6 backfill interfaces.

The shared column-group metadata accessors now live in `column_groups_jni.cpp`. The regression tests exercise V3 writer → transaction commit → manifest reopen → reader, including ownership and lifetime checks.

Validation: 18 Java/Scala tests and 6 existing C++ builder tests passed. All current JNI sources and the builder were rebuilt; the remaining native-core objects and dependencies were reused, so this is targeted validation rather than a full rebuild. Arrow's JNI loader still emits five strict-JNI warnings.

This change covers the storage bindings in this PR. Removing the Spark connector's existing V2 call paths is separate work in that repository.

cpp/include/milvus-storage/ffi_internal/v2_column_groups_builder.h
santiago-wjq: Removed in 8ef20b081bfe07d4811c3aae9907faf7aedc0014. The C++ builder and Scala/JNI `createFromGroups` are back to the original three-argument API, with parquet fixed by the V2 implementation. The legacy builder, V2 JNI sources and existing C++ tests now match the PR base.

The shared `format` metadata accessor is retained for column groups returned by V3 writers/manifests. It does not introduce a V2 format option. The six existing V2 builder tests pass.

cpp/CMakeLists.txt
santiago-wjq: Reverted the dependency-visibility change in 4308ecc01f7e52e9daec7621f548b1817031550b. `milvus-storage` again uses `PUBLIC ${LINK_LIBS}`, and the JNI, test and benchmark link settings match the base branch.

The separate `-z noexecstack` block and its comments have also been removed and pushed in 5689301241df37f0ed52bc76eb040406e542b3de. `cpp/CMakeLists.txt` now matches the PR base. `git diff --check` passed; native build/runtime validation after the rollback is pending CI.

cpp/src/jni/reader_jni.cpp
santiago-wjq: These are performance counters for the existing Arrow slice materialization in the JNI reader. `copies` counts column arrays copied by `arrow::Concatenate` to normalize nonzero offsets, and `copied_bytes` sums the resulting arrays' top-level buffer sizes.

The Spark connector exposes them as `milvus.copies` and `milvus.copied.bytes` on the SQL scan node to help diagnose copy overhead when transferring Arrow batches from C++ to the JVM. They do not affect read results or execution decisions, and the counters introduce no additional data copies. They cover this slice-normalization step, not all copies in the read pipeline.

cpp/src/jni/filesystem_jni.cpp
santiago-wjq: `path` is a Java `jstring`, while `loon_filesystem_read_file_all` takes a `const char*` and a byte length. `ScopedUtf8` obtains the native string bytes and automatically calls `ReleaseStringUTFChars` when leaving the scope, including on early returns. This conversion is needed to pass the path across the JNI boundary, so we are keeping it.

One encoding detail will be addressed separately: `GetStringUTFChars` returns JNI modified UTF-8 rather than standard UTF-8. Correct handling of supplementary Unicode characters in paths, including the reverse conversion back to Java, is deferred to a follow-up and is outside this PR's scope.

java/src/main/java/io/milvus/storage/NativeLibraryLoader.java
santiago-wjq: Yes, we need to support loading on macOS so engineers can run Spark tests locally on their Macs. `darwin` is the existing macOS resource-directory identifier (`native/darwin-x86_64/` or `native/darwin-aarch64/`), which this loader preserves.

java/src/main/java/io/milvus/storage/NativeLibraryLoader.java
santiago-wjq: The intent is to distribute one cross-platform JAR containing native libraries for all supported platforms, following the usual native-library fat-JAR packaging pattern. The Java classes are portable, but the bundled native binaries are specific to the OS and CPU architecture (`.so` on Linux, `.dylib` on macOS).

The loader therefore uses the current OS and architecture to select `native/<os>-<arch>/` and extract only the matching JNI library and its dependencies. This prevents libraries for different platforms from being mixed or overwriting each other during extraction.

cpp/src/jni/reader_jni.cpp
jiaqizho: try with arrow type `FixedSizeList<float32, 4>` or other embed type. It will be a problem.

santiago-wjq: Yes, `FixedSizeList<float32, 4>` stores its float values in child arrays, so this top-level buffer sum can substantially undercount embeddings. These counters are intended only as rough diagnostics for the existing concat workaround, not precise copy or allocation accounting. The calculation is unchanged in this PR; I have added comments documenting this limitation in both the native implementation and the Java API.

Concat is still needed because the reader can return slices with nonzero offsets, which the Arrow Java C Data import path used here does not honor. Exporting those slices directly can repeat earlier rows. Once JVM consumers handle offsets correctly and the concat workaround is removed, these temporary counters and their JNI API should be removed together.

PR #682, opened 2026-09-16
enhance: add async Azure input reads using the Arrow IO executor

jiaqizho:
Azure input files only exposed Arrow’s regular random-access interface, so asynchronous callers had to fall back to the generic path even though the blocking Azure SDK operations could safely run on the filesystem’s IO executor. This made Azure behave differently from other providers that expose NonBlockingRandomAccessFile and prevented callers from selecting the optimized async read path at runtime.

Make the Azure input file implement NonBlockingRandomAccessFile and add asynchronous size lookup and range reads. Cached sizes return immediately, while uncached metadata requests and range downloads are submitted to the executor from the file’s IOContext. Submitted operations keep the input file alive until the blocking Azure SDK call finishes, and validation or submission failures are returned through completed Arrow futures.

Move Azure read metrics into the input file and return it directly from both OpenInputStream and OpenInputFile. This keeps read counts and byte totals unchanged while avoiding the existing metrics wrappers hiding the new runtime capability from dynamic_cast<NonBlockingRandomAccessFile*>.


cpp/src/filesystem/azure/azurefs.cc
```diff
@@ -1012,24 +1148,79 @@ class ObjectInputFile final : public io::RandomAccessFile {
...
+
+    // Azure SDK reads are blocking, so they run on the filesystem's Arrow IO pool.
+    // Parquet CPU work and continuations run on the caller's separate Folly pool.
+    // With this separation, even a synchronous footer wait on the CPU side does
+    // not occupy an IO worker, so this submission cannot cause a same-pool deadlock.
+    // Keep blocking Parquet opens off this IO pool if supplying a custom IOContext.
+    // The caller owns `out`; the file owns itself until the submitted read completes.
+    auto self = std::shared_ptr<ObjectInputFile>(shared_from_this(), this);
+    auto submitted =
+        io_context_.executor()->Submit([self = std::move(self), position, nbytes, out]() -> Result<int64_t> {
```
buqian-zilliz, line 1213: **[P2] Avoid blocking the same executor that services Azure reads**

Under `WITH_CRT`, the synchronous Parquet footer path now casts this file to `NonBlockingRandomAccessFile` and immediately waits on `ReadAtAsyncInto(...).result()`. Because this implementation always queues the actual read back to the file's `IOContext` executor, invoking synchronous open from that executor, or filling all its workers with such opens, deadlocks while the queued reads have no available worker. `AzureFileSystem` accepts caller-supplied `IOContext` executors, so this is reachable. Keep synchronous footer reads on `ReadAt`, or otherwise ensure this implementation cannot enqueue and wait on the same executor.

cpp/src/filesystem/azure/azurefs.cc
```diff
@@ -436,16 +447,43 @@ Status ExceptionToStatus(const Azure::Core::RequestFailedException& exception,
...
+Status ExceptionToStatus(const Azure::Core::Credentials::AuthenticationException& exception,
+                         PrefixArgs&&... prefix_args) {
   std::stringstream ss;
   (ss << ... << std::forward<PrefixArgs>(prefix_args));
-  ss << " Azure Error: [" << exception.ErrorCode << "] " << exception.what();
-  auto message = ss.str();
-  return milvus_storage::MakeExtendError(*code, message, message);
+  ss << " Azure authentication failed: " << exception.what();
+  const auto message = ss.str();
+  return milvus_storage::MakeExtendError(milvus_storage::ExtendStatusCode::AwsErrorAccessDenied, message, message);
```
buqian-zilliz, line 480: **[P2] Do not equate every token acquisition failure with access denied**

`AuthenticationException` is not limited to a storage 401 or 403. Azure Identity also uses it when token acquisition cannot execute; for example, its token pipeline wraps transport, timeout, and process failures in this exception. Mapping every instance to the non-retryable `AwsErrorAccessDenied` detail misclassifies temporary identity-provider failures as permanent credential or permission errors, so callers suppress retries. Preserve an unknown status unless the underlying failure can be positively classified, or classify recoverable token-acquisition failures separately.

cpp/src/filesystem/azure/azurefs.cc
```diff
@@ -458,24 +496,35 @@ std::string AzureOptions::AccountDfsUrl(const std::string& account_name) const {
...
 
 Status AzureOptions::ConfigureDefaultCredential() {
-  credential_kind_ = CredentialKind::kDefault;
-  token_credential_ = std::make_shared<Azure::Identity::DefaultAzureCredential>();
-  return Status::OK();
+  const auto on_error = [&](const auto& exception) -> Status {
+    return ExceptionToStatus(exception, "ConfigureDefaultCredential failed for account '", account_name, "'.");
+  };
+  try {
+    credential_kind_ = CredentialKind::kDefault;
```
buqian-zilliz: **[P2] Commit credential state only after construction succeeds**

If credential construction throws, this new status-returning path leaves `credential_kind_` set to `kDefault` while `token_credential_` still contains the previous credential. `MakeBlobServiceClient` then skips default construction because the pointer is non-null and can silently authenticate with that stale credential if the options object is reused after the error. Construct into a local first, then assign the pointer and credential kind together only after construction succeeds; apply the same ordering to the other token-credential setters.

cpp/src/filesystem/azure/azurefs.cc
```diff
@@ -2796,15 +3191,24 @@ class AzureFileSystem::Impl {
         std::optional<Status> first_failure_status;
         for (auto& [blob_name_view, deferred_response] : deferred_responses) {
           bool success = true;
+          const auto on_blob_error = [&, blob_name_view = blob_name_view](const auto& exception) {
+            success = false;
+            if (!first_failure_status.has_value()) {
```
buqian-zilliz, line 3197: **[P2] Do not classify a mixed batch by the first failed blob**

A batch can return independent failures for different blobs, but this records only the first exception and uses that status detail for the entire operation. For example, a 503 followed by a 403 is reported as transient, whereas the same failures in the opposite blob order are reported as permanent. Retry behavior therefore depends on blob enumeration order and can retry-storm a batch that also has a permanent failure. Aggregate all subresponse classifications using explicit deterministic precedence instead of copying the first one.

cpp/src/filesystem/azure/azurefs.cc
```diff
@@ -1012,24 +1148,79 @@ class ObjectInputFile final : public io::RandomAccessFile {
...
+        return Status::IOError("Unexpected DownloadTo Content-Range length ", bytes_read, " for '", location_.all,
+                               "' at position ", position, " for range read of ", nbytes, " bytes");
       }
       RETURN_NOT_OK(CacheContentLengthFromRead(result.Value));
+      if (bytes_read > 0) {
+        metrics_->IncrementReadCount();
+        metrics_->IncrementReadBytes(bytes_read);
+      }
       return bytes_read;
+    } catch (const Storage::StorageException& exception) {
```
buqian-zilliz, line 1172: **[P3] Collapse identical request-failure catch arms**

`StorageException` and `TransportException` both derive from `RequestFailedException`, and all three branches here immediately call the same `on_error` handler. Replace them with one `catch (const Azure::Core::RequestFailedException&)`, retaining the separate authentication and generic handlers; apply the same simplification to other catch ladders that perform no subtype-specific handling. This removes duplicated control flow and automatically covers future Azure request-failure subtypes.

cpp/src/filesystem/azure/azurefs.cc
jiaqizho: we don't allow use IO thread pool to execute the callback from filesystem which same as arrow.

So it won happen

cpp/src/filesystem/azure/azurefs.cc
jiaqizho: ack

cpp/src/filesystem/azure/azurefs.cc
jiaqizho: We intentionally keep `AuthenticationException` mapped to non-retryable `AwsErrorAccessDenied` at the storage boundary. Retryable token-request failures are handled by Azure Identity's HTTP pipeline; we do not add another storage-level retry after authentication has failed. The original SDK error message is preserved.

cpp/src/filesystem/azure/azurefs.cc
jiaqizho: This PR retains the existing first-failure behavior in the Arrow-derived batch-delete implementation. The same selection logic predates this change. Introducing precedence across mixed failures would change the operation's error and retry semantics, so we are keeping it unchanged here.

cpp/src/filesystem/azure/azurefs.cc
jiaqizho: Keeping the explicit catch branches in this PR. Although the three `ReadAt` branches are equivalent with the current handlers, other catch ladders do not have a `RequestFailedException` arm. Replacing their derived catches with the base catch would broaden the handled set and change plain request failures from `UnknownError` to the Azure error mapping. We are preserving the current exception boundaries rather than applying that cleanup across the file.

cpp/src/filesystem/azure/azurefs.cc
```diff
@@ -993,6 +1104,32 @@ class ObjectInputFile final : public io::RandomAccessFile {
...
+      return Future<int64_t>::MakeFinished(status);
+    }
+    const auto content_length = GetCachedContentLength();
+    if (content_length != kNoSize) {
+      return Future<int64_t>::MakeFinished(content_length);
+    }
+
+    // Keep the file alive until the blocking Azure SDK call leaves the IO pool.
+    auto self = std::shared_ptr<ObjectInputFile>(shared_from_this(), this);
+    auto submitted = io_context_.executor()->Submit([self = std::move(self)]() -> Result<int64_t> {
```
buqian-zilliz, line 1119: **[P2] Preserve IOContext cancellation when scheduling Azure calls**

Both new async methods call `io_context_.executor()->Submit(...)` directly. This drops the `IOContext` stop token and `external_id` that Arrow's `io::internal::SubmitIO` forwards, so queued work can begin after the context is cancelled and custom executors lose task attribution. This is especially costly because each task can block on a remote Azure request. Submit through `io::internal::SubmitIO(io_context_, ...)` in both methods.

cpp/src/filesystem/azure/azurefs.cc
```diff
@@ -458,24 +496,36 @@ std::string AzureOptions::AccountDfsUrl(const std::string& account_name) const {
...
-Status AzureOptions::ConfigureAccountKeyCredential(const std::string& account_key) {
+Status AzureOptions::ConfigureAccountKeyCredential(const std::string& account_key) try {
   credential_kind_ = CredentialKind::kStorageSharedKey;
   if (account_name.empty()) {
     return Status::Invalid("AzureOptions doesn't contain a valid account name");
   }
   storage_shared_key_credential_ =
       std::make_shared<Storage::StorageSharedKeyCredential>(account_name, account_key);
   return Status::OK();
+} catch (const std::exception& exception) {
```
buqian-zilliz, line 527: **[P2] Preserve shared-key state when construction fails**

The new function-try-block turns credential-construction exceptions into a returned `Status`, but `credential_kind_` is changed to `kStorageSharedKey` before `StorageSharedKeyCredential` has been constructed. After the catch returns, a reused `AzureOptions` can therefore advertise shared-key authentication while `storage_shared_key_credential_` is null or still contains an older key; the client factories may use that stale credential, and `Equals` can dereference the null pointer. Construct the credential into a local and publish both members only after construction succeeds, matching the token-credential setters.

jiaqizho: ack

cpp/src/filesystem/azure/azurefs.cc
jiaqizho: We are keeping this unchanged for now. The current Reader/filesystem path does not provide end-to-end cancellation, and the standard Azure factory uses the default IOContext (`StopToken::Unstoppable()`, `external_id = -1`). Propagating cancellation and custom task metadata is outside this PR's scope; we can address it consistently across providers when that capability is needed.

PR #684, opened 2026-09-17
feat: add host-controlled OpenTelemetry tracing for storage reads

somfornot:
This PR continues fork-backed #668 as the bottom layer of native GitHub stack #685, with #683 as the dependent Talon adaptation. Prior reviews and discussion remain available in #668.

Storage reads cross reader, metadata-cache, filesystem and Rust async boundaries, making it difficult for a host request trace to account for the complete operation. This change adds host-controlled OpenTelemetry tracing and preserves parent context across those boundaries.

## Clang 14 compatibility

`OwnedAttributes::ForEachKeyValue` now binds ordinary references to each map entry before invoking its visitor. This avoids capturing a structured binding, which the CI clang-tidy 14 parser does not support, while retaining the same attribute ownership and avoiding copies. The fix is in commit `2d89676`; #683 is rebased onto this layer.

Development-container validation for this fix: clang-tidy 14 on the actual `runtime.cpp` reproduced the original compiler errors before the change and exited successfully after it; `milvus_test` rebuilt with `WITH_TALON=ON`; all 40 `StorageTracingTest` cases passed, including native attribute types and deferred array/link ownership. Clang-format 18, the error-handling ratchet (21 throw sites) and `git diff --check` passed. The existing ASAN build ran with leak detection disabled. The full repository tidy job is left to the newly triggered CI.

## Changes

- Add C++ provider configuration and scoped parent injection without installing a global provider. Trace reader, format, metadata-cache and file I/O operations, including asynchronous S3 requests, with span budgets and sanitized error classes. Expose generic RAII scopes with initial attributes, links and span kind, mutable attributes, explicit status completion, and parent construction from an OpenTelemetry SpanContext. C++ hosts must rebuild against the updated header.
- Observe returned errors, early returns and exception unwinding while preserving original business results and exceptions across synchronous, Folly and native asynchronous calls. Own deferred attributes and links until span creation, and provide initial attributes to the host sampler. Return configuration failures as Status without replacing the previous configuration; contain tracing attachment failures and keep metadata in-flight loads progressing.
- Propagate context through Folly, Arrow and Rust async work. Start lazy spans when work executes, preserve per-operation configuration, and keep shared readers and file handles independent of request parents.
- Reduce context checks, repeated span-start locking and Rust bridge handle allocations. Null-provider operations retain a lightweight disabled snapshot without span state, budgets, or I/O completion accounting. Remove duplicate Vortex context binding and task boxing while preserving Tokio scheduling, profiling labels and cancellation. Cover mixed traced/untraced cache callers, concurrent lazy starts, host sampler decisions and context restoration.
- Add tracing and complete-read benchmarks, a paired comparison runner, and design, integration and performance documentation.

## Rebase and stack validation (2026-09-17)

Rebased this layer onto `main` at `7f8e0b9`; the new head is `c5afcec`. Conflict resolution preserves the upstream Rust module relocation, Lance dataset version cache and current Talon fallback behavior. Talon tracing is the separate dependent layer #683, not part of this PR's incremental scope. The branch chain is `main <- feature/opentel <- feature/opentel-talon`. This is the bottom layer of the native GitHub stack, with #683 as the dependent Talon adaptation.

At the combined Talon adapter head `5fbbe7a`, development-container validation with `WITH_TALON=ON` passed: 394 selected Storage/Talon tests (including all 40 tracing tests), 77 conditional skips, 83 C FFI tests, and 7 new Talon tracing integration tests. Tracked-source clang-format 18 and the unchanged error ratchet passed. The Folly ThreadedExecutor ABI case remains excluded. The validation and performance history below predates this rebase; benchmarks were not rerun.

## Previous validation

All builds, formatting and tests ran in the development container with the required Conan cache mount. Current C++ validation uses the existing ASAN/coverage build with leak detection disabled:

- Tracing suite: 40 passed. Coverage includes real Reader early failures, original synchronous and asynchronous exceptions, scope completion and context restoration, sampler-visible initial attributes, deferred attribute/link ownership, configuration failure, and metadata in-flight progress under persistent tracing fault injection.
- Selected reader, format, metadata-cache, executor, local-filesystem and filesystem-metrics regression suite: 308 passed, 69 conditionally skipped (377 total), separately from the tracing suite.
- C FFI: 83 passed, 0 failed.
- Built `milvus_test`, `Test_FFI`, `tracing_benchmark` and `telemetry_read_benchmark`.
- Changed C++ files passed clang-format 18 checks; `git diff --check` and the error-handling ratchet passed (21 throw sites, with no baseline or gate changes). Unwind observers record tracing errors without catching and replacing business exceptions.
- Earlier validation of the unchanged Rust adapter used a standalone probe linked to the actual C++/Folly bridge: parent restoration across Pending/Ready/unwind, CPU versus blocking-pool placement, profiling labels, cancellation and subsequent work passed. That probe was not rerun for this C++ change.
- The known Folly `ThreadedExecutor` dependency/compiler ABI mismatch test remains excluded; this is not a full-suite pass. Remote storage and deployed end-to-end/exporter behavior have not been validated. Performance measurements below are historical and were not rerun for the current scope and exception changes.

## Sample: actual local trace output

Captured on commit `80cccd0`, before the disabled-path follow-up, in the development container, using the existing ASAN/coverage test build and a standalone local harness. The harness writes 100 rows in each of Parquet and Vortex, then calls `reader->take_async({0, 3, 7}, 2)` twice on the same Reader with metadata caching enabled. All four reads verified the returned IDs `[0, 3, 7]`; the test passed.

The host injects a parent context, an `AlwaysOnSampler`, and an `InMemorySpanExporter`, with `io_spans=true` and `max_spans_per_operation=256`. The tree below is reconstructed from actual exported span IDs and parent IDs; durations and attributes are measured values. This is a formatted SDK span dump, not OTLP wire-format output or a network-exporter test.

Actual Parquet cold-metadata read (head spans and some attributes omitted for readability):

```text
storage.read                           54.633 ms
| storage.operation=take_async
| storage.io.reads=3
| storage.io.requested_bytes=5103
| storage.io.returned_bytes=5103
| storage.spans.dropped=0
|
+-- storage.read_task                  53.153 ms
|   +-- storage.metadata.lookup         0.078 ms  storage.cache=miss
|   +-- storage.metadata.load           3.922 ms
|   |   +-- storage.fs.read             0.742 ms
|   |       storage.backend=local, storage.offset=2346
|   |       storage.requested_bytes=611, storage.returned_bytes=611
|   +-- storage.format.read            48.185 ms  storage.format=parquet
|       +-- storage.fs.read             0.834 ms
|           storage.offset=4, storage.requested_bytes=2246, storage.returned_bytes=2246
|
+-- storage.read_task                  53.314 ms
|   +-- storage.metadata.lookup         0.035 ms  storage.cache=in_flight
|   +-- storage.metadata.wait           3.268 ms
|   +-- storage.format.read            48.550 ms  storage.format=parquet
|       +-- storage.fs.read             1.376 ms
|           storage.offset=4, storage.requested_bytes=2246, storage.returned_bytes=2246
|
+-- storage.assemble                    0.075 ms  storage.operation=build_take_table
```

The two read tasks overlap: one loads metadata while the other waits for the in-flight load. Three logical file reads return 5,103 bytes in total. The subsequent read reports `storage.cache=hit` in both tasks.

| Local operation | Root span duration | Exported spans | Logical file reads | Returned bytes |
| --- | ---: | ---: | ---: | ---: |
| Parquet, cold metadata | 54.633 ms | 16 | 3 | 5,103 |
| Parquet, warm metadata | 4.663 ms | 12 | 2 | 4,492 |
| Vortex, cold metadata | 9.791 ms | 11 | 1 | 10,956 |
| Vortex, warm metadata | 3.290 ms | 8 | 0 | 0 |

These are individual trace examples from an ASAN/coverage build, not performance comparisons or acceptance results. Cold/warm refers to metadata caching, not disk caching; the Vortex warm read's zero file reads is specific to this small-file case. Format spans include I/O wait, and overlapping child durations must not be added to infer end-to-end latency. A parent/child tree alone does not depict overlap; span start times and durations provide the timeline. This sample has one file-read child per format-read span. Separately, the current `ReadManyAsync` wrapper emits one aggregate logical span with `storage.range_count`, rather than an individual span for each range.

## Performance status — pending acceptance

The optimization comparison uses commit `6966d49` as its baseline, with matching Release builds and jemalloc. Null-provider root-plus-four-child allocations fall from 27 to 23, and Rust capture-plus-attach allocations from 8 to 6. Holding 50,000 captured Rust context handles reduces jemalloc allocated bytes from 3.2 MB to 1.6 MB; this does not represent whole-service RSS savings.

Six representative local read comparisons detected no statistically significant regression against that baseline, but three confidence intervals remain inconclusive. The 2% measurement sensitivity is not an accepted regression allowance. No-regression acceptance against the version before Telemetry was introduced is still outstanding.

The recorded performance comparison predates the upstream merge; performance acceptance also remains pending for the merged head.

The disabled-path follow-up also compares the C++ runtime sources from `80cccd0` with `4b45ee7` using matching Release tooling and dependencies. Each operation includes parent attachment (empty tracestate), one root and four synchronous children; the asynchronous root returns a ready SemiFuture. After 100 warmup iterations and 1,000 counted iterations:

| Null-provider runtime probe | C++ new calls per operation | Requested bytes per operation |
| --- | ---: | ---: |
| Parent attached, synchronous root | 33 → 31 | 2,232 → 1,992 |
| Parent attached, ready asynchronous root | 48 → 32 | 5,408 → 2,152 |
| No parent, synchronous root | 0 → 0 | 0 → 0 |
| No parent, ready asynchronous root | 1 → 1 | 160 → 160 |

These counts include attachment and business-Future allocations and have a different denominator from the historical probe above. They measure C++ runtime allocations only, not complete-read latency, Rust allocation totals or RSS. These measurements describe `4b45ee7`. The current implementation preserves original business exceptions and returns disabled futures without adding an exception-conversion continuation; allocation and latency measurements have not been rerun for this change. Folly still propagates RequestContext, so these results do not establish zero overhead. Full no-regression acceptance against the pre-tracing version remains pending.

See [Storage OpenTelemetry](docs/opentelemetry-storage.md) for the API contract, propagation boundaries, validation limits, benchmark methodology and optimization measurements.


cpp/src/format/bridge/rust/src/runtime/storage_tracing.rs
```diff
@@ -0,0 +1,108 @@
+// Copyright 2026 Zilliz
```
jiaqizho, line 1: move this file into `cpp/src/format/bridge/rust/src/runtime/storage_tracing.rs` ?

cpp/src/reader.cpp
```diff
@@ -966,67 +1011,82 @@ class ReaderImpl : public Reader {
...
+          FOLLY_ARROW_ASSIGN_OR_RAISE(auto lazy_readers, create_lazy_readers(needed_column_groups, resolved_columns));
+
+          // Fan out file-aware tasks across all required column groups, then combine
+          // the reordered per-group tables into the requested logical column order.
+          return take_tables_async(row_indices, needed_column_groups, lazy_readers, parallelism)
+              .deferValue([storage_context = tracing::Capture(), row_indices,
+                           resolved_columns = std::move(resolved_columns), schema = schema_,
+                           lazy_readers](auto&& tables_result) -> arrow::Result<std::shared_ptr<arrow::Table>> {
+                tracing::ContextScope storage_scope(storage_context);
+                tracing::StartCurrent();
```
jiaqizho: The tracing helpers have quite different completion rules: `Run` handles synchronous results, `RunAsync` ends the span in a Folly continuation, `RunNativeAsync` relies on the native callback for operations that are still pending, and `Observe` ends the span in an Arrow Future callback. It takes a fair amount of implementation knowledge to choose the right one.

On top of that, continuations like this one manually repeat `Capture -> ContextScope -> StartCurrent`. Missing a task boundary can silently lose the context. Could we use the existing `Bind` helper consistently for these cases, and document which helper to use for each async path, who owns span completion, and which thread runs it? That would make future instrumentation much easier to get right.

cpp/src/tracing/runtime.h
```diff
@@ -0,0 +1,166 @@
...
+  return future;
+}
+
+// Observe the source Arrow future itself: completion does not depend on a
+// consumer running a continuation and remains observed after a future is dropped.
+template <typename T>
+arrow::Future<T> Observe(arrow::Future<T> future, OperationTrace trace) {
+  if (!trace.NeedsCompletion())
+    return future;
+  future.AddCallback([trace](const arrow::Result<T>& result) { trace.Finish(result.status()); });
```
jiaqizho, line 195: This callback can end up exporting spans on the CRT thread. `Finish()` calls `Span::End()`, so a host-provided `SimpleSpanProcessor` with a synchronous exporter runs the export inline. The S3 paths call `MarkFinished()` from a CRT callback, which means a slow exporter can block that thread and delay other requests.

In the local completion-thread probe, an exporter that slept for 50ms made `MarkFinished()` take about 55ms, and both ran on the same thread. This was a local probe rather than a remote S3 test. Please move potentially blocking span completion/export work to an execution boundary controlled by the caller so it cannot run on the CRT thread.

cpp/src/format/bridge/rust/src/runtime/rust_runtime.rs
```diff
@@ -71,7 +71,32 @@ fn default_runtime_thread_count() -> usize {
...
 /// not create separate Tokio worker and blocking thread pools per format.
-pub(crate) static TOKIO_RT: LazyLock<tokio::runtime::Runtime> = LazyLock::new(|| {
+pub(crate) struct StorageTokioRuntime(tokio::runtime::Runtime);
+
+impl StorageTokioRuntime {
+    pub(crate) fn handle(&self) -> &tokio::runtime::Handle {
+        self.0.handle()
+    }
+    pub(crate) fn block_on<F: std::future::Future>(&self, future: F) -> F::Output {
+        self.0.block_on(crate::storage_tracing::instrument(future))
```
jiaqizho, line 81: Wrapping the outer future here does not cover the actual I/O tasks created by Lance's default standard scheduler. Its `run_io_loop` calls `tokio::spawn(task.run())` inside the dependency, bypassing this runtime wrapper. By the time the task reaches OpenDAL and the C++ filesystem, the Storage parent context has been lost.

The local probe successfully read the first chunk under a valid parent but exported only `storage.format.read`, with no I/O spans. Running the same probe with the lite scheduler produced 9 I/O spans, which helps narrow this down to the scheduler's task boundary. We need to carry the context with the actual I/O requests; otherwise default Lance reads still miss both the I/O spans and their aggregate counters.

cpp/src/format/parquet/parquet_format_reader.cpp
```diff
@@ -928,85 +964,97 @@ static arrow::Future<std::shared_ptr<::parquet::arrow::FileReader>> create_file_
...
+            // Resolve a missing size without blocking the consuming thread. A
+            // native async file owns the request; the generic fallback submits
+            // GetSize() to the executor already owned by the file's IOContext.
+            arrow::Future<int64_t> file_size_future;
+            if (auto* async_file = dynamic_cast<milvus_storage::NonBlockingRandomAccessFile*>(file.get())) {
+              file_size_future = transfer_arrow_future(async_file->GetSizeAsync(), arrow_executor);
+            } else {
+              FOLLY_ARROW_ASSIGN_OR_RAISE(auto submitted_size_future,
+                                          file->io_context().executor()->Submit([file] { return file->GetSize(); }));
+              file_size_future = transfer_arrow_future(std::move(submitted_size_future), arrow_executor);
```
jiaqizho: This `GetSize()` task is submitted without capturing the tracing context. When `file_size_ == 0` and the file does not implement `NonBlockingRandomAccessFile`, the lambda runs on the file's Arrow executor. `TracedFile::GetSize()` then sees no Storage context and skips the `storage.fs.head` span. Restoring the context in the later continuation cannot recover that missing span.

The local `open_async()` probe with an unknown-size Parquet file reproduced this: opening succeeded, but the head span was missing. Please wrap this lambda with `tracing::Bind` so the context is captured at submission and restored when the task runs.

cpp/src/filesystem/s3/s3_filesystem.cpp
```diff
@@ -896,39 +906,48 @@ class ObjectCrtInputFile final : public arrow::io::RandomAccessFile, public NonB
...
+                  expected, *content_length, std::memory_order_acq_rel, std::memory_order_acquire);
+            }
+            if (bytes_read > 0) {
+              ctx->metrics->IncrementReadBytes(bytes_read);
+            }
+            ctx->future.MarkFinished(bytes_read);
+          });
+    } catch (...) {
+      ctx->metrics->IncrementFailedCount();
+      ctx->future.MarkFinished(arrow::Status::UnknownError("GetObject submission exception"));
```
jiaqizho: This catch replaces every submission exception with the fixed `UnknownError("GetObject submission exception")`, losing the original diagnostic details. Callers can no longer tell what actually failed, and `std::bad_alloc` also loses its out-of-memory classification. The `HeadObjectAsync` catch has the same issue, and both affect callers even when tracing is disabled.

Please handle `std::bad_alloc` and other identifiable exceptions separately, map them to an appropriate Arrow Status, and preserve details such as `what()`. The span can still record only a sanitized error classification while the error returned to the caller retains the information needed to diagnose the failure.

cpp/src/ffi/filesystem_c.cpp
```diff
@@ -444,28 +446,33 @@ LoonFFIResult loon_filesystem_reader_readat_async(FileSystemReaderHandle handle,
...
-        });
+        }
+
+        auto status = read_result.status();
+        auto result = CreateFFIResult(FFIErrorCodeFromExtendStatus(status, LOON_ARROW_ERROR),
+                                      "Fail to read object data asynchronously, details: ", status.ToString());
+        callback(user_data, result, 0);
+        return;
+      } catch (...) {
+        callback(user_data, LoonFFIResult{LOON_GOT_EXCEPTION, nullptr}, 0);
```
jiaqizho: If processing the completion throws, this catch reports only `LOON_GOT_EXCEPTION` with a null message. The caller still knows that something failed, but the exception details are gone. Normal Arrow read failures preserve `status.ToString()` just above, so this specifically affects exceptions raised while handling the completion.

Could we preserve the diagnostic details for ordinary `std::exception` failures where reporting them is safe, while keeping a no-allocation fallback for allocation failures? We still need to prevent exceptions from escaping the C boundary, but the fallback should not discard every identifiable error unconditionally.

cpp/src/tracing/runtime.cpp
```diff
@@ -0,0 +1,500 @@
...
+    FIU_RETURN_ON(FIUKEY_TRACING_CONFIGURATION_FAIL, arrow::Status::UnknownError("tracing configuration failure"));
+    auto tracer = provider ? provider->GetTracer("milvus-storage", MILVUS_STORAGE_VERSION) : nullptr;
+    std::lock_guard<std::mutex> lock(configuration_mutex);
+    auto next = std::make_shared<Configuration>(*configuration);
+    next->provider = std::move(provider);
+    next->tracer = std::move(tracer);
+    configuration = std::move(next);
+    return arrow::Status::OK();
+  } catch (...) {
+    return arrow::Status::UnknownError("Failed to configure Storage tracer provider");
```
jiaqizho: Both `SetTracerProvider()` and `SetTraceOptions()` replace every caught exception with a fixed `UnknownError`. The failure is returned to the host, but the original reason and error category are lost, so the host cannot diagnose why configuration failed.

Please preserve useful exception details and map identifiable failures to an appropriate Arrow Status, including an out-of-memory status for `std::bad_alloc`. Keeping the previous configuration on failure is useful; we should retain the diagnostic information as well.

cpp/src/tracing/runtime.cpp
```diff
@@ -0,0 +1,500 @@
...
+      if (link.IsValid()) {
+        span = config->tracer->StartSpan(name, initial, {{link, {}}}, options);
+      } else {
+        span = config->tracer->StartSpan(name, initial, options);
+      }
+    }
+    started.store(true, std::memory_order_release);
+  } catch (...) {
+    // No partially initialized span is published by this path.
+  }
```
jiaqizho, line 268: An exception caught here disappears without any diagnostic signal. Span creation can fail while the storage operation continues, and the host has no way to distinguish that failure from intentionally missing tracing.

Could we make these failures observable through a lightweight failure counter or a rate-limited diagnostic, while preserving the business result? The same concern applies to the silent fallbacks in `OperationTrace` construction, `WithAttributes()`, `MetadataLoadTrace()`, and attribute/link iteration. Any reporting needs to remain safe under allocation failure and avoid blocking native I/O callback threads.

cpp/src/tracing/runtime.cpp
```diff
@@ -0,0 +1,500 @@
...
+      failed_ = true;
+      return;
+    });
+    const auto* request = folly::RequestContext::try_get();
+    const auto* data = request ? static_cast<const Data*>(request->getContextData(storage_key)) : nullptr;
+    if (context || (data && data->context))
+      scope_.emplace(storage_key, std::make_unique<Data>(std::move(context)));
+  } catch (...) {
+    failed_scopes.fetch_add(1, std::memory_order_acq_rel);
+    failed_ = true;
```
jiaqizho, line 336: This records that an attachment failed, but discards the reason. `failed_scopes` is also process-wide: while a failed guard is alive, `Capture()` and `HasContext()` suppress tracing for other requests too. That makes a silent attachment failure particularly difficult to investigate when traces start disappearing.

Could we retain a lightweight diagnostic for the failure, without changing the storage operation's result or adding blocking work to the callback path? The two `TraceScope` constructors have the same catch-and-suppress behavior and should be covered as well.

cpp/src/tracing/runtime.cpp
```diff
@@ -0,0 +1,500 @@
...
+        op->span->SetAttribute("storage.io.reads", op->budget->reads.load());
+        op->span->SetAttribute("storage.io.requested_bytes", op->budget->requested_bytes.load());
+        op->span->SetAttribute("storage.io.returned_bytes", op->budget->returned_bytes.load());
+      }
+    } catch (...) {
+      // Annotation failure must not prevent span completion.
+    }
+    op->span->End();
+  } catch (...) {
+  }
```
jiaqizho, line 657: Both catch blocks discard the exception here. An annotation failure can leave the span without its error classification or I/O counters, and a failure caught by the outer block can leave completion unreported, with no diagnostic explaining what happened.

Keeping tracing failures from replacing the business result makes sense. Could we still record a lightweight indication of which tracing step failed, using the same non-blocking failure reporting as the creation path? This should remain safe during unwinding and allocation failure.

cpp/src/format/bridge/rust/src/storage_tracing.rs
```diff
@@ -0,0 +1,108 @@
...
+pub(crate) fn instrument<F: Future>(future: F) -> impl Future<Output = F::Output> {
+    let context = ffi::capture_trace_context().unwrap_or_else(|_| cxx::SharedPtr::null());
+    async move {
+        // Pin inside the wrapper's state rather than adding a heap allocation
+        // for every future, including futures with tracing disabled.
+        futures::pin_mut!(future);
+        futures::future::poll_fn(move |cx| {
+            // The guard is local to each poll, including Pending/unwinding.
+            // Empty snapshots still mask an unrelated worker-thread parent.
+            let _scope = ffi::attach_trace_context(&context).ok();
```
jiaqizho: The CXX bridge errors are discarded in both places: a capture failure becomes a null context, and an attachment failure becomes `None` through `.ok()`. The task keeps running, but we lose the reason its tracing context could not be propagated. `bind()` repeats the same pattern below.

Could we retain a lightweight diagnostic for these failures before falling back? The wrapped future or task should keep its original result, but an actual capture/attachment error needs to be distinguishable from the normal case where no parent context was supplied. Any reporting on this path should remain non-blocking.

PR #692, opened 2026-09-18
enhance: add native async paths for Lance

jiaqizho:
Lance's async reader APIs currently fall back to blocking calls, which can tie up caller threads during dataset opens and small reads. Add native async paths so callers can keep requests in flight without waiting for each operation to finish.

Run dataset and fragment opens, metadata I/O, take, and range reads on the shared Tokio runtime. Keep the C++ bridge as a thin callback wrapper, and let LanceTableReader own the async context and complete Folly promises. Collect the requested batches before reporting success so consuming the returned Arrow stream does not trigger more I/O.

Reuse the opened fragment reader through Arc ownership and keep a bounded cache of native async read handles to avoid reopening the same objects for repeated small reads. Hold reader resources until completion, and release replaced or evicted handles outside the cache lock and CRT callbacks.

Extend dataset and fragment metadata caching to async opens while keeping the existing snapshot and filesystem boundaries. Make abandoned loads release their in-flight state and notify waiting callers, and preserve the original error when setting up a continuation fails. Move the global dataset cache into lance_table_cache.cpp to keep caching logic separate from reader operations.


cpp/src/format/lance/lance_table_cache.cpp
```diff
@@ -0,0 +1,313 @@
...
+    uint64_t version;
+    std::string base_uri;
+    std::string filesystem_cache_key;
+
+    bool operator==(const Key&) const = default;
+    auto operator<=>(const Key&) const = default;
+  };
+
+  static LanceDatasetCache& Instance() {
+    static LanceDatasetCache cache;
```
liliu-z: `LanceDatasetCache::Instance()` returns a function-local static (lance_table_cache.cpp:87), and `GetOrOpenAsync` captures a raw `this` into three continuations that outlive the call: the deferValue body (:165), the `abandoned` scope guard (:189), and the deferred completion (:202). Those continuations are run or destroyed from Tokio worker threads, and `TOKIO_RT` is a `LazyLock<Runtime>` that is never dropped, so its threads are still live while `__cxa_atexit` destroys the static; a completion arriving after that point makes the `noexcept` guard call `CompleteOpen`, which locks an already-destroyed `std::mutex` and touches destroyed containers. The two sibling single-flights do the opposite and document why: `format_reader_cache.cpp:140-142` and `lance_table_reader.cpp:198` both pin the cache with `shared_from_this()`.

cpp/test/format/lance/lance_table_test.cpp
```diff
@@ -658,6 +677,193 @@ TEST_F(LanceBasicTest, MaterializedReadsTranslateDeferredErrors) {
 
 #ifdef WITH_CRT
```
liliu-z: `lance_table_test.cpp:678-910` gates `AsyncOpenReturnsBeforeIoCompletes`, `AsyncRangeMaterializesBeforeCompletionAndRetainsReader`, `AsyncTakeRetainsReaderUntilCompletion` and `TimedOutAsyncOpenDoesNotStrandDatasetCache` on `#ifdef WITH_CRT`, which `cpp/CMakeLists.txt:15` defaults to OFF and no CI job sets. The Rust path these tests exercise is gated on the `s3-crt-async` cargo feature, which `Cargo.toml:12` also enables through `talon`, so every CI build (`cpp-ci.yml:68`/`:148`, `cpp-mac-ci.yml:87` all pass `WITH_TALON=True`) compiles the code but not its tests. Nothing in the test harness needs the C++ `WITH_CRT` define: `CallbackThreadRandomAccessFile` only implements the unconditional `milvus_storage::NonBlockingRandomAccessFile` interface that `filesystem_c.cpp:414` dynamic_casts on. `TimedOutAsyncOpenDoesNotStrandDatasetCache` is the only coverage for the new process-wide `LanceDatasetCache` abandon path (`lance_table_cache.cpp:185-221`), whose regression mode is a permanently unfulfilled `async_result` that hangs every later async open of that dataset key until restart.

cpp/src/format/lance/lance_table_cache.cpp
jiaqizho: ACK

cpp/test/format/lance/lance_table_test.cpp
jiaqizho: ACK

cpp/src/format/bridge/rust/src/filesystem/filesystem_object_store.rs
```diff
@@ -682,6 +687,14 @@ impl FilesystemObjectStore {
                 source: Box::new(source),
             })?;
 
+        #[cfg(feature = "s3-crt-async")]
+        {
+            let cached = { self.read_handles.lock().unwrap().get(location).cloned() };
```
liliu-z, line 692: `read_range` short-circuits through `read_handles` on every call (line 692) and inserts every async-capable `ReaderHandle` (line 778), but no code path removes an entry on error — only LRU eviction or dropping the object store. Once a cached handle stops serving reads, every subsequent read of that object through this store fails, including Lance's own `DEFAULT_DOWNLOAD_RETRY_COUNT` retries, which re-enter the same cache entry; the store lives as long as the cached dataset. The field's only documentation, `// Diagnostic only: immutable benchmark objects`, contradicts this: the `talon` feature enables `s3-crt-async`, so the cache is active in every CI and production build, and `grep read_handles` finds no test coverage.

cpp/src/format/lance/lance_table_cache.cpp
```diff
@@ -0,0 +1,316 @@
...
+      std::lock_guard<std::mutex> lock(mutex_);
+      if (in_flight_open->done) {
+        return in_flight_open->status.ok() ? arrow::Result<DatasetPtr>(in_flight_open->dataset)
+                                           : in_flight_open->status;
+      }
+      if (status.ok()) {
+        auto cached = datasets_.get(key);
+        if (cached.has_value()) {
+          if (auto existing = cached->lock()) {
+            dataset = std::move(existing);
```
liliu-z, line 257: At line 257 `dataset = std::move(existing)` drops the last reference to the just-opened `BlockingDataset` inside the `std::lock_guard` taken at line 248, running the Rust `Dataset` teardown — including closing every cached native `ReaderHandle` — under the process-wide `LanceDatasetCache` mutex, which serializes all dataset lookups and opens. The branch is reachable whenever a synchronous `GetOrOpen` takes the async-leader bypass at lines 148-161 and publishes its own dataset before the async leader completes. The synchronous path at lines 152-157 already orders this correctly, and the PR description states replaced handles are released outside the cache lock.

jiaqizho: Keeping the current implementation. Destroying the duplicate dataset under the lock can increase contention, but we have not identified a correctness issue or a deadlock here. Moving the destruction outside the lock would be a performance improvement, so we are leaving it out of this change.

cpp/src/format/bridge/rust/src/filesystem/filesystem_object_store.rs
jiaqizho: Keeping the current implementation. ReaderHandle owns a C++ RandomAccessFile, not an individual request or connection. A failed read does not invalidate the stream in the current CRT/Talon implementations, so subsequent reads can reuse it, as they do elsewhere in the codebase. We have not identified a failure that requires reopening the stream to recover, so we will not add eviction on read errors.

